PR #7, opened 2020-08-01
ignore local folder

pisuke:
The `local` folder is meant to have configuration specific files, can it be added to .gitignore?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ workspace.xml
 /validator/out/
 /pubber/build/
 /sites/
+/local/


grafnu:
What configuration specific files are there with UDMI?  The local/ folder was like that with DAQ, but right now the tools don't use it at all. Was there something specific you were trying to keep there?   We can do it if it's legit, but just wanted to avoid the "because DAQ did it" line of reasoning!

pisuke:
There are some references in the docs to use local/system.conf to add the service account json file and system folder.
Maybe it's good to keep this behaviour and use this file with `registrar` and `pubber`?
Another reason is that the site files can be added inside the local folder and ignored by git.

pisuke:
Looking again at `.gitignore`, the `sites` folder seems to be the space where to put sites files, I had not picked that up from the docs. So OK to close this PR without merging if there is no intention to add config files in the future.

grafnu:
Ah, so the docs are out-of-date :-).  And let's pick up the discussion in the other thread (about credentials). I would like to do it properly for the right reasons... and so if we work by use-case we'll get there quicker.

Issue #5, opened 2020-08-01
realpath dependency


pisuke:
`realpath` is invoked by the wrapper scripts, but it is not present in all operating systems.
For instance on MacOS I had to install it with `brew install coreutils`.
It would be useful to declare the dependency in the documentation and to create an install script.

grafnu:
Alternatively as per [Stack Overflow](https://stackoverflow.com/questions/4175264/how-to-retrieve-absolute-path-given-relative), something like:

`echo "$(cd "$(dirname "$1")"; pwd)/$(basename "$1")"`

would work without the dependency. 

I'm not sure I'm going to have the bandwidth to keep making sure things work on MacOS, so if you find things like this that need to be tweaked, please feel free to make PR requests with the changes!

pisuke:
I've done PR #8 to address this.

PR #8, opened 2020-08-01
realpath substitution

pisuke:
To address issue #5 realpath dependency.

### DIFF
--- a/bin/pubber
+++ b/bin/pubber
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-ROOT_DIR="$(pwd)/$line"
+ROOT_DIR="$(pwd)"
 
 if [ $# != 3 ]; then
     echo $0 PROJECT_ID SITE_PATH DEVICE_ID
--- a/bin/pubber
+++ b/bin/pubber
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-ROOT_DIR=$(realpath $(dirname $0)/..)
+ROOT_DIR="$(pwd)/$line"
 
 if [ $# != 3 ]; then
     echo $0 PROJECT_ID SITE_PATH DEVICE_ID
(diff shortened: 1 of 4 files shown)


bin/pubber
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-ROOT_DIR=$(realpath $(dirname $0)/..)
+ROOT_DIR="$(pwd)/$line"
```
grafnu, line 3: what's $line?

bin/test_schema
```diff
@@ -31,7 +31,7 @@ if [ "$build" == y ]; then
     validator/bin/build
 fi
 
-jarfile=$(realpath validator/build/libs/validator-1.0-SNAPSHOT-all.jar)
+jarfile="$(cd "$(dirname "validator/build/libs/validator-1.0-SNAPSHOT-all.jar")"; pwd)/$(basename "validator/build/libs/validator-1.0-SNAPSHOT-all.jar")"
```
grafnu, line 34: should split out the root-path thing into a separate variable... too ugly to comprehend otherwise.  Also, 'basename' and 'dirname' not necessary with a static string -- you can just use the filename directly.

bin/pubber
pisuke: sorry, lazy copy/paste from another script without thinking, now removed

bin/test_schema
pisuke: I've put the JAR filename into a different variable and removed the static string, I think it's easier to read now.

bin/pubber
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-ROOT_DIR=$(realpath $(dirname $0)/..)
+ROOT_DIR="$(pwd)"
```
grafnu, line 3: this won't work since there shouldn't be a requirement to run it from a certain directory. It needs to be in reference to $0, since that specifies the target (udmi install directory), not the directory it's run from.

PR #36, opened 2020-11-16
update docs

pbatta:


.github/workflows/testing.yml
```diff
@@ -11,6 +11,7 @@ jobs:
         uses: actions/checkout@v2.3.4
       - name: Install dependencies
         run: |
+          sudo apt-get update --fix-missing
```
grafnu, line 14: What's this necessary for?   This shouldn't be necessary since everything is done from a clean install and any dependency should be explicitly listed.

pbatta: the --fix-missing is indeed unnecessary, took it out. The update seems to be needed: the apt-get install has failed in the last few test runs because it is unable to find packages, seems to be running off some older list?

.github/workflows/testing.yml
```diff
@@ -11,6 +11,7 @@ jobs:
         uses: actions/checkout@v2.3.4
       - name: Install dependencies
         run: |
+          sudo apt-get update
```
grafnu, line 14: Is this necessary as part of this PR?  I think you might be trying to address two things at the same time. I'd keep the documentation stuff separate from the build issues.

pbatta: good point, split that out to a separate PR

grafnu (approved):
I think you can submit this without the tests passing... that really has nothing to do with documentation updates!

PR #79, opened 2021-01-21
Basic writeback examples

grafnu:


schema/event_pointset.tests/writeback.json
```diff
@@ -0,0 +1,25 @@
+{
+  "version": 1,
+  "timestamp": "2018-08-26T21:39:29.364Z",
+  "etag": "2a8b718",
```
rniemo-g, line 4: I don't think etag is strictly necessary in the telemetry message. If the cloud wants to figure out what state the device was in when a given telemetry message was sent, the cloud can compare timestamps. I won't block on this, but WDYT?

rniemo-g (approved):

grafnu:
Seems to me the etag is strictly more useful in the pointset event than in
state. WIth queues there could be quite a number of "old" messages lying
about, so seems like you'd want a way to easily distinguish old from new.
The argument about timestamps is true for everything tho, so that would
argue for no etag anywhere (e.g. in state either). For me, I'd say it
should either be everywhere or nowhere :-).

Either way, we can have it for now and then take it out later if somehow it
causes a problem. Right now, the field is optional so won't fundamentally
break anything if it's not there.

PR #133, opened 2021-05-14
adding updated pyudmi to contrib

paulharter:
This is an updated version of pyudmi that correctly uses the latest json schemas.

There is one quirk: I am using a library called fastjsonschema which needs to have `$` always escaped in regex. See docs here.

https://horejsek.github.io/python-fastjsonschema/

So I have escaped a `$` in the `metadata_system.json` which should have no adverse effects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,10 +19,5 @@ firebase-debug.log
 /venv/
 /local/
 
-/contrib/pyudmi/build/
-/contrib/pyudmi/src/udmi/schema/
-/contrib/pyudmi/dist/
-/contrib/pyudmi/pyudmi.egg-info/
-
 .DS_Store
 .idea/
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,11 @@ firebase-debug.log
 /attributes.json
 /venv/
 /local/
+
+/contrib/pyudmi/build/
+/contrib/pyudmi/src/udmi/schema/
+/contrib/pyudmi/dist/
+/contrib/pyudmi/pyudmi.egg-info/
+
+.DS_Store
+.idea/
--- a/contrib/pyudmi/src/udmi/base.py
+++ b/contrib/pyudmi/src/udmi/base.py
@@ -1,0 +1,96 @@
+import json
+import os
+import pytz
+import datetime
+import fastjsonschema
+import udmi
+import dateutil.parser
+
+DEFAULT_UDMI_VERSION = 1
+
+VALIDATORS = {}
+PACKAGE_SCHEMATA_DIR = os.path.join(os.path.dirname(udmi.__file__), "schema")
+UDMI_SCHEMATA_DIR = os.path.join(os.path.dirname(udmi.__file__), "..", "..", "..", "..", "schema")
+
+SCHEMATA_DIR = PACKAGE_SCHEMATA_DIR if os.path.exists(PACKAGE_SCHEMATA_DIR) else UDMI_SCHEMATA_DIR
+
+
+def get_path(uri):
+    filename = os.path.basename(uri)
+    path = "%s/%s" % (SCHEMATA_DIR, filename)
+    with open(path, "r") as f:
+        return json.loads(f.read())
+
+
+def get_validator(name):
+    validator = VALIDATORS.get(name)
+    os.chdir(SCHEMATA_DIR)
+    if validator is None:
+        file_path = os.path.join(SCHEMATA_DIR, name)
+        with open(file_path, "r") as f:
+            schema = json.loads(f.read())
+        handlers = {"file": get_path}
+        validator = fastjsonschema.compile(schema, handlers=handlers)
+        VALIDATORS[name] = validator
+
+    return validator
+
+
+class UDMIBase:
+    schema = "none"
+
+    def __init__(self, version):
+        self.version = version
+        self.validate()
+
+    def __str__(self):
+        return json.dumps(self.as_dict(), indent=4, sort_keys=True)
+
+    def as_udmi(self):
+        return json.dumps(self.as_dict(), indent=4, sort_keys=True)
+
+    @classmethod
+    def from_string(cls, s):
+        return cls.from_dict(json.loads(s))
+
+    @classmethod
+    def from_dict(cls, d):
+        return cls(**d)
+
+    def as_dict(self):
+
+        d = {}
+
+        for name in self.__slots__:
+            value = getattr(self, name, None)
+            if value is not None:
+                if hasattr(value, "as_dict"):
+                    d[name] = value.as_dict()
+                elif type(value) in (str, int, float, list, dict, tuple, bool):
+                    d[name] = value
+                else:
+                    raise Exception("Can't serialise this value %s for json" % value)
+        return d
+
+    def validate(self):
+        validator = get_validator(self.schema)
+        validator(self.as_dict())
+
+
+    @staticmethod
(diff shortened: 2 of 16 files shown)


.gitignore
```diff
@@ -18,3 +18,11 @@ firebase-debug.log
 /attributes.json
 /venv/
 /local/
+
+/contrib/pyudmi/build/
```
grafnu, line 22: I think these can go in a .gitignore file in the contrib/pyudmi directory itself... makes it easier to compartmentalize

contrib/pyudmi/src/udmi/config.py
```diff
@@ -0,0 +1,57 @@
+"""
+{
```
grafnu, line 2: what are these?  They seem redundant with the schema definitions in the UDMI/schema directory -- redundancy is bad (much harder to maintain), so I'm trying to understand what they are... ?

paulharter: The schema json strings in these files are documentation only. The python models are oo way to create udmi objects that are validated on creation

contrib/pyudmi/src/udmi/envelope.py
```diff
@@ -0,0 +1,65 @@
...
+
+"""
+
+from .base import UDMIBase, DEFAULT_UDMI_VERSION
+
+
+class Envelope(UDMIBase):
+
+    schema = "envelope.json"
+    __slots__ = [
```
grafnu, line 50: what's the effective impact of the __slots__? From reading online, it seems to be a memory optimization, useful for when you're creating "thousands or millions" of objects... was there something else this was for?  It seems like a lot of extra overhead (the __slots__) for something that is likely to have much impact (when are there going to be so many objects?)

grafnu:
Just wondering if you investigated any of the python code generators for jsonschmea -- e.g. https://pypi.org/project/json-schema-codegen/ -- seems like a cleaner way that manually writing all the base class code?

contrib/pyudmi/src/udmi/config.py
paulharter: sure - i'll take take these strings out

contrib/pyudmi/src/udmi/envelope.py
paulharter: yes - slots are there mostly as an optimisation. I thought it might be useful if working with large collections of pointsets. But then i'm also reusing them for serialisation. This would be better done directly from the schemas.

paulharter: yes - the slots are mostly an optimisation, which i though might be useful if working with large collections of pointsets. But I have also reused them in the serialisation to json, this could/should be done by referring directly to the schemas instead, then the slots could be removed.

paulharter: I have  tried the json codegen libs but they might be useful (they've changed the way these comments work and I can't work out how to inline comment below)

grafnu:
Is there missing "not" in there?  By comments do you mean the GitHub
comments?

grafnu:
I also had some comments about the file structure. I'm not super-duper with all the python systems out there, so partly this is just based on my limited experience and expectations.

Right now, there's a basic structure that has
* `udmi/contrib/pyudmi/src/udmi` -- the basic source
* `udmi/contrib/pyudmi/src/udmi/tests` -- the tests

Normally, I would expect it to look like:
* `udmi/contrib/pyudmi/pyudmi` (source)
* `udmi/contrib/pyudmi/tests` (tests)

Was there a template you were following?  Or, do you think we could reconsider moving some of the files around?

Ideally, there'd also be an CI test added, but we can do that later. That would go in `udmi/.github/workflows/testing.yml` or similar.

contrib/pyudmi/src/udmi/base.py
```diff
@@ -0,0 +1,95 @@
...
+import pytz
+import datetime
+import fastjsonschema
+import udmi
+import dateutil.parser
+
+DEFAULT_UDMI_VERSION = 1
+
+VALIDATORS = {}
+PACKAGE_SCHEMATA_DIR = os.path.join(os.path.dirname(udmi.__file__), "schema")
```
grafnu, line 12: the use of schemata (rather than just schema) seems incongruous with other uses (e.g., we talk about "The UDMI schema") -- so I think it might be more coherent overall to use "schema"?

pisuke:
> Just wondering if you investigated any of the python code generators for jsonschmea -- e.g. https://pypi.org/project/json-schema-codegen/ -- seems like a cleaner way that manually writing all the base class code?

@grafnu thanks for the suggestion, I just tried this with the files in the schema folder and I got some errors like this one:
```
NotImplementedError: {'enum': 1}
```
so it looks like the JSON Schema support of this library is not complete (`enum` keyword missing in this case).

As we're already using `pyudmi` in a few projects and tools [like this one to generate site models from Google Sheets](https://github.com/arupiot/gsheet2UDMI), we can perhaps include it with your suggestions and then see if it's worthwhile contributing to the `json-schema-codegen` later.

grafnu:
Interesting -- I've been working at converting some of the Java code I have
to use a generator, and have been running into a few problems but for the
most part I've been able to resolve them.  Will likely have to go through a
similar exercise for a python version.

But -- either way -- given that you're already using pyudmi places then we
can keep it as-is and then sort it out later.

The problem with any hand-made classes is that they work fine now, but over
time they just get out of sync with the schema overall since it's near
impossible to keep all the different representations up-to-date when any
changes are made!  So, this is definitely a problem "of the future" and
we'll just deal with it when we're there.

Cheers,
 Trevor

pisuke:
> The problem with any hand-made classes is that they work fine now, but over time they just get out of sync with the schema overall since it's near impossible to keep all the different representations up-to-date when any changes are made! So, this is definitely a problem "of the future" and we'll just deal with it when we're there. Cheers, Trevor

Thank you Trevor, I'll reply to the other comments then so that we can proceed with the inclusion, and I agree that at some point we'll have to revisit the library generation strategy.

contrib/pyudmi/src/udmi/base.py
pisuke: I think schemata is plural of schema, but OK to change this term if felt more in line with other uses elsewhere.

grafnu:
Yes -- I'd assert there's actually only one schema ("The UDMI schema") --
so that directory is the (singular) schema directory -- I don't think it's
accurate/consistent to pluralize it. All the things in that directory are
sub-components of the overall schema.  It's tricky because it really is a
schema of schemas (of schemas)... but to be consistent :-).

pisuke:
> Was there a template you were following? Or, do you think we could reconsider moving some of the files around?

We could move things around, however I think Paul has been following some of [these recommendations](https://blog.ionelmc.ro/2014/05/25/python-packaging/) and [this article](https://hynek.me/articles/testing-packaging/) (@paulharter please confirm).

Changing the `udmi` import name to `pyudmi` will break some of our other work based on this library, but I guess if we feel we must do this, it's not too late.

grafnu:
Ok -- I think the only real problem might be the "tests" inside of the src
directory -- I think that's one of the alternatives mentioned. But, if it's
udmi, pydmi, src/udmi or whatever I don't think matters much.

grafnu:
Closing for now.  There is now a generated-python code capability (done in-house because none of the external ones were viable), but it's not clear of the ongoing utility of this library. Feel free to re-open if it's of active interest.

PR #138, opened 2021-06-23
Bugfix for x509 device registration

pisuke:
I've discovered a bug with the x509 device registration that prevents the registration of devices if RSA and ES public keys are contained together with X509 certificates in the device folder.
This bugfix corrects this behaviour and allows all devices to be registered without having to manually remove public keys from the device folder when the creation of a key with a self-signed X509 certificate is selected.

### DIFF
--- a/validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
+++ b/validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
@@ -104,6 +104,7 @@ class LocalDevice {
       ES_AUTH_TYPE, ES_PRIVATE_KEY_FILES,
       ES_CERT_TYPE, ES_PRIVATE_KEY_FILES
   );
+
   private static final Map<String, String> PUBLIC_KEY_FILE_MAP = ImmutableMap.of(
       RSA_AUTH_TYPE, RSA_PUBLIC_PEM,
       RSA_CERT_TYPE, RSA_CERT_PEM,
@@ -112,11 +113,16 @@ class LocalDevice {
   );
 
   private static final Set<String> OPTIONAL_FILES = ImmutableSet.of(
-      RSA2_PUBLIC_PEM, RSA3_PUBLIC_PEM, ES2_PUBLIC_PEM, ES3_PUBLIC_PEM,
+      RSA_PUBLIC_PEM, RSA2_PUBLIC_PEM, RSA3_PUBLIC_PEM, 
+      ES_PUBLIC_PEM, ES2_PUBLIC_PEM, ES3_PUBLIC_PEM,
       GENERATED_CONFIG_JSON, DEVICE_ERRORS_JSON, NORMALIZED_JSON, SAMPLES_DIR);
   private static final Set<String> ALL_KEY_FILES = ImmutableSet.of(
       RSA_CERT_PEM, RSA_PUBLIC_PEM, RSA2_PUBLIC_PEM, RSA3_PUBLIC_PEM,
       ES_CERT_PEM, ES_PUBLIC_PEM, ES2_PUBLIC_PEM, ES3_PUBLIC_PEM
+  );
+  private static final Set<String> ALL_CERT_FILES = ImmutableSet.of(
+      RSA_CERT_PEM, 
+      ES_CERT_PEM
   );
 
   private static final Map<String, String> AUTH_TYPE_MAP = ImmutableMap.of(
@@ -270,7 +276,15 @@ class LocalDevice {
       if (!hasAuthType()) {
         throw new RuntimeException("Credential cloud.auth_type definition missing");
       }
-      for (String keyFile : ALL_KEY_FILES) {
+      String authType = getAuthType();
+      Set<String> keyFiles;
+      if (authType.equals(ES_CERT_TYPE) || authType.equals(RSA_CERT_TYPE))
+      {
+        keyFiles = ALL_CERT_FILES;
+      } else {
+        keyFiles = ALL_KEY_FILES;
+      }
+      for (String keyFile : keyFiles) {
         DeviceCredential deviceCredential = getDeviceCredential(keyFile);
         if (deviceCredential != null) {
           deviceCredentials.add(deviceCredential);
--- a/validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
+++ b/validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
@@ -311,10 +311,9 @@ class LocalDevice {
     Set<String> certFile = getCertFiles();
     Set<String> publicKeyFile = Set.of(getPublicKeyFile());
     Set<String> privateKeyFiles = getPrivateKeyFiles();
-    Set<String> rtn = (authType.equals(ES_CERT_TYPE) || authType.equals(RSA_CERT_TYPE)) ? 
+    return (authType.equals(ES_CERT_TYPE) || authType.equals(RSA_CERT_TYPE)) ? 
         Sets.union(Sets.union(publicKeyFile, certFile), privateKeyFiles) : 
         Sets.union(publicKeyFile, privateKeyFiles);
-    return rtn;
   }
 
   private Set<String> getPrivateKeyFiles() {
--- a/validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
+++ b/validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
@@ -111,24 +111,21 @@ class LocalDevice {
       ES_CERT_TYPE, ES_PUBLIC_PEM
   );
   private static final Map<String, String> CERT_FILE_MAP = ImmutableMap.of(
-    RSA_CERT_TYPE, RSA_CERT_PEM,
-    ES_CERT_TYPE, ES_CERT_PEM
+      RSA_CERT_TYPE, RSA_CERT_PEM,
+      ES_CERT_TYPE, ES_CERT_PEM
   );
   private static final Set<String> OPTIONAL_FILES = ImmutableSet.of(
-    RSA2_PUBLIC_PEM, RSA3_PUBLIC_PEM, ES2_PUBLIC_PEM, ES3_PUBLIC_PEM,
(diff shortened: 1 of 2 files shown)


validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -270,7 +276,15 @@ public void loadCredentials() {
       if (!hasAuthType()) {
         throw new RuntimeException("Credential cloud.auth_type definition missing");
       }
-      for (String keyFile : ALL_KEY_FILES) {
+      String authType = getAuthType();
+      Set<String> keyFiles;
+      if (authType.equals(ES_CERT_TYPE) || authType.equals(RSA_CERT_TYPE))
+      {
```
grafnu, line 282: opening brace should be on previous line

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -270,7 +276,15 @@ public void loadCredentials() {
       if (!hasAuthType()) {
         throw new RuntimeException("Credential cloud.auth_type definition missing");
       }
-      for (String keyFile : ALL_KEY_FILES) {
+      String authType = getAuthType();
+      Set<String> keyFiles;
```
grafnu, line 280: declare final or use ternary if

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -112,12 +113,17 @@
   );
 
   private static final Set<String> OPTIONAL_FILES = ImmutableSet.of(
-      RSA2_PUBLIC_PEM, RSA3_PUBLIC_PEM, ES2_PUBLIC_PEM, ES3_PUBLIC_PEM,
+      RSA_PUBLIC_PEM, RSA2_PUBLIC_PEM, RSA3_PUBLIC_PEM,
```
grafnu, line 116: RSA_ and ES_ shouldn't be optional.  They are strictly required if that's the designated key type. If there's an error you're encountering then those cases should be handled more explicitly.   (E.g., RSA_PUBLIC_PEM is required iff the auth type is RSA.)

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -112,12 +113,17 @@
...
-      RSA2_PUBLIC_PEM, RSA3_PUBLIC_PEM, ES2_PUBLIC_PEM, ES3_PUBLIC_PEM,
+      RSA_PUBLIC_PEM, RSA2_PUBLIC_PEM, RSA3_PUBLIC_PEM, 
+      ES_PUBLIC_PEM, ES2_PUBLIC_PEM, ES3_PUBLIC_PEM,
       GENERATED_CONFIG_JSON, DEVICE_ERRORS_JSON, NORMALIZED_JSON, SAMPLES_DIR);
   private static final Set<String> ALL_KEY_FILES = ImmutableSet.of(
       RSA_CERT_PEM, RSA_PUBLIC_PEM, RSA2_PUBLIC_PEM, RSA3_PUBLIC_PEM,
       ES_CERT_PEM, ES_PUBLIC_PEM, ES2_PUBLIC_PEM, ES3_PUBLIC_PEM
   );
+  private static final Set<String> ALL_CERT_FILES = ImmutableSet.of(
+      RSA_CERT_PEM,
```
grafnu, line 124: Do these values need to still be in ALL_KEY_FILES, or can/should they be removed from there?

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
pisuke: Done.

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
pisuke: Thanks for the suggestion, I corrected this to ternary if.

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
pisuke: OK, understood. I've changed the code to be specific about the RSA and ES keys and certs.

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
pisuke: They need to be there.

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -104,24 +104,28 @@
...
+  private static final Map<String, String> CERT_FILE_MAP = ImmutableMap.of(
+    RSA_CERT_TYPE, RSA_CERT_PEM,
+    ES_CERT_TYPE, ES_CERT_PEM
   );
-
   private static final Set<String> OPTIONAL_FILES = ImmutableSet.of(
-      RSA_PUBLIC_PEM, RSA2_PUBLIC_PEM, RSA3_PUBLIC_PEM, 
-      ES_PUBLIC_PEM, ES2_PUBLIC_PEM, ES3_PUBLIC_PEM,
-      GENERATED_CONFIG_JSON, DEVICE_ERRORS_JSON, NORMALIZED_JSON, SAMPLES_DIR);
+    RSA2_PUBLIC_PEM, RSA3_PUBLIC_PEM, ES2_PUBLIC_PEM, ES3_PUBLIC_PEM,
```
grafnu, line 118: hanging intent should be 4

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -309,12 +307,17 @@ private DeviceCredential getDeviceCredential(String keyFile) throws IOException
...
+    if (!isDirectConnect() || !hasAuthType()) {
       return ImmutableSet.of();
     }
+    String authType = getAuthType();
+    Set<String> certFile = (authType.equals(ES_CERT_TYPE) || authType.equals(RSA_CERT_TYPE)) ? Set.of(certFile()) : null;
     Set<String> publicKeyFile = Set.of(publicKeyFile());
     Set<String> privateKeyFiles = privateKeyFiles();
-    return Sets.union(publicKeyFile, privateKeyFiles);
+    Set<String> rtn = (authType.equals(ES_CERT_TYPE) || authType.equals(RSA_CERT_TYPE)) ? 
+       Sets.union(Sets.union(publicKeyFile, certFile), privateKeyFiles) :
```
grafnu, line 318: hanging indent should be 4 (I think it's 3 here?)

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -309,12 +307,17 @@ private DeviceCredential getDeviceCredential(String keyFile) throws IOException
   }
 
   private Set<String> keyFiles() {
-    if (!isDirectConnect()) {
+    if (!isDirectConnect() || !hasAuthType()) {
       return ImmutableSet.of();
     }
+    String authType = getAuthType();
+    Set<String> certFile = (authType.equals(ES_CERT_TYPE) || authType.equals(RSA_CERT_TYPE)) ? Set.of(certFile()) : null;
```
grafnu, line 314: I would move all of this logic into the certFile() function (and rename it getCertSet()) -- no need to have this exposed here since otherwise that function is trivial.

Also, variable should be named certFiles or certSet -- it's not a singular file.

grafnu (commented):
Why do the certs need to be in ALL_KEY_FILES? Looking at the logic, if certs are used then it uses ALL_CERT_FILES... and the cert files shouldn't be there if it's in key (non-cert mode)... ?

pisuke:
> Why do the certs need to be in ALL_KEY_FILES? Looking at the logic, if certs are used then it uses ALL_CERT_FILES... and the cert files shouldn't be there if it's in key (non-cert mode)... ?

Sorry, I misunderstood your original question. Yes, I agree the certs don't need to be there. I'll remove them in my next commit.

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
pisuke: yup, my bad, sorry

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
pisuke: yup, my bad, sorry

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
pisuke: In the tests I've done with the RS256_X509 and ES256_X509 cases I've only experienced having a single cert file, so I've modelled this function similarly to the public key one.
Does it mean I need to change both the cert and the public key functions?

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -295,25 +304,41 @@ private DeviceCredential getDeviceCredential(String keyFile) throws IOException
...
       return ImmutableSet.of();
     }
-    Set<String> publicKeyFile = Set.of(publicKeyFile());
-    Set<String> privateKeyFiles = privateKeyFiles();
-    return Sets.union(publicKeyFile, privateKeyFiles);
+    String authType = getAuthType();
+    Set<String> certFile = getCertFiles();
+    Set<String> publicKeyFile = Set.of(getPublicKeyFile());
+    Set<String> privateKeyFiles = getPrivateKeyFiles();
+    Set<String> rtn = (authType.equals(ES_CERT_TYPE) || authType.equals(RSA_CERT_TYPE)) ?
```
grafnu, line 314: don't abbreviate -- either just return inline, or use semantically meaningful name (not "return" or "rtn")

pisuke: done, I've chosen to return inline

grafnu (approved):

PR #140, opened 2021-07-04
docs/gradle: Fix formatting and align Gradle distribution versions.

kendfinger:
Hello! I was reading through a lot of the repo and figured I would try to make some simple cleanup changes while I'm still digging into things.

1. Fix some trivial typos in the docs.
2. Docs formatting, mostly newline related.
3. Gradle now downloads the -all version of the Gradle distribution, instead of -bin and -all differing between pubber and validator.
4. settings.gradle is now setting the rootProject.name in both projects.
5. Gradle build files now use one quoting format.


grafnu (approved):

grafnu:
Thanks for cleaning all that up!

PR #190, opened 2021-10-05
add z, lat, long, to metadata, define x & y

noursaidi:
- added `z` axis to the the metadata schema
- added `lat` & `long` to the meta data schema, and defined `x` and `y` as the position in a project specific coordinate reference system. The reasons:
  - Lat and long are established terms and used for wider scale geolocation
  - lat and long are very not suited to smaller scales, e.g. approx 0.00001° represents a 5m displacement
  - metres is the default for smaller scales - understood, building drawings/ plans will have measurements in metres, etc
  - `x` and `y` can be inferred from `lat` and `long` and vice versa, but it's good to have both, and there are project requests to have both


grafnu:
Ok -- based off all of this, here's my proposal:

   - lat/long/alt are global, specified in the site_model.
   - x/y/z are per-device, as a relative offset to the global offset.

There is a degenerate case of lat/long/alt being empty, where then the
x/y/z would effectively be either global (if they're lat/long/alt), or
floating (just enough to display on a map of the building).

From UDMIs perspective I think both sets need to be optional -- but there
might be another layer of requirements (DBS?) that stipulate what is
compulsory.

Probably good as an agenda item for the UDMI meeting Thr.

noursaidi:
@grafnu @pisuke as per all the discussions, this PR has been updated so:
- z axis is added
- x,y,z are no longer required and are optional
- x,y,z defined as reference in a project specific reference system

grafnu (approved):
Might be some merge conflicts requiring a docs update?

noursaidi:
Merge conflicts resolved

PR #192, opened 2021-10-05
Add partial_update to base schema.

grafnu:


pisuke (approved):
LGTM, good point about adding some guidance on default values for telemetry / logging updates

noursaidi (approved):

PR #199, opened 2021-11-28
Smart Buildings Technical Primer

noursaidi:


docs/tech_primer.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Devices Classification
+
+Various adjectives qualify the different flavors of a "device" found in a building. More than one can apply, and each has a specific technical meaning/definition (i.e. they are not subjective).
+
+### Connectivity
+
+How data is moved around on-prem
+
+-   **Passive**: The device has only passive components (thermistor, dry contact)
```
grafnu, line 33: I don't understand what "passive" means in this context -- can you give a specific example for this?  It feels like it might be getting out-of-scope for what needs to be differentiated in this context... The comparison would of course be "active" -- but I'm still not sure why that would matter here.

docs/tech_primer.md
```diff
@@ -0,0 +1,157 @@
...
+-   **Analog**: The device uses an analog connection (1-10V, 4-20mA etc)
+-   **Serial**: The device does not have an IP address, rather uses a wired serial connection.
+-   **Networked**: The device has its own IP address and presence on the wired or wireless IP network.
+-   **Bridge connected**: The device is on an isolated wired or wireless network behind a gateway device that is networked.
```
grafnu, line 37: Need one word. Just "Bridged"

docs/tech_primer.md
```diff
@@ -0,0 +1,157 @@
...
+-   **Direct**: Maintains an authenticated connection directly to Cloud IoT Core.
+    -   Must be networked
+    -   Has a unique private auth key
+-   **Gateway**: A direct device that manages data for other proxied devices.
+    -   Bind & attach of proxied devices
+    -   Encapsulation anti-pattern
+-   **Proxied**: A device that is not direct and is managed through a gateway.
+    -   Proxied networked devices, e.g. BACnet/IP
+    -   Proxied non-networked devices, e.g. RS485, DALI
+-  **Cloud-Cloud**: A device is managed by a third party who provides a cloud connector
```
grafnu, line 52: I don't get what this means. What's an example of what you're trying to talk about here?

docs/tech_primer.md
```diff
@@ -0,0 +1,157 @@
...
+    -   Approved third party managed service
```
grafnu, line 53: "third party" and "managed" are not _technical_ concepts so shouldn't be here. What is _technically_ different about this that needs to be covered?

docs/tech_primer.md
```diff
@@ -0,0 +1,157 @@
...
+    -   UDMI virtual device server connects to third party cloud API
+
+### Representation
+
+How data is semantically represented
```
grafnu, line 58: mmm -- not quite sure "semantically" fits here, there still is no _meaning_ behind it. It's really something about how the data is incorporated into the data processing pipeline.

docs/tech_primer.md
```diff
@@ -0,0 +1,157 @@
...
+Qualification can be automated using the [DAQ tool](https://github.com/faucetsdn/daq/). 
+The output of DAQ is a report. 
+[example report](https://github.com/faucetsdn/daq/blob/master/docs/device_report.md))
+
+### Required Actions
+
+* Devices must be officially qualified for use on the project.
+
+### Verification
+
```
grafnu, line 105: "...device _are_ available."

docs/tech_primer.md
```diff
@@ -0,0 +1,157 @@
...
+* Device qualification configuration and reports for the device available.
+
+## System Testing
```
grafnu, line 108: Let's leave this ("System Testing") out for now since it's not very well defined and will just add confusion.

docs/tech_primer.md
```diff
@@ -0,0 +1,157 @@
...
+
+### Required Actions
+* Data flow diagrams and Interoperability Matrix provided
+
+### Validation
+* Interoperability validated against qualified device configurations.
+
+## Managed Network 
+
+### Required Actions
```
grafnu, line 118: I think "Prerequisite" is likely a better word. "Required Actions" is getting into "process" (who should do what), rather than "the technical thing that must exist"

docs/tech_primer.md
```diff
@@ -0,0 +1,157 @@
...
+
+* [site model](site_model.md)  provided and all devices registered into GCP
+* devices are setup in a qualified configuration
+
+### Verification
+
+* [registrar](registrar.md) and [device telemetry validations](validator.md) are clean
+
+## Digital Buildings
```
grafnu, line 145: Add "Ontology"

docs/tech_primer.md
```diff
@@ -0,0 +1,157 @@
...
+### Required Actions
+* Data flow diagrams and Interoperability Matrix provided
+
+### Validation
+* Interoperability validated against qualified device configurations.
+
+## Managed Network 
+
+### Required Actions
+* All devices appropriately registered for network, and asset information provided
```
grafnu, line 119: The "registered for network" is going to be problematic. We need to think through this completely. You can't assume any of the Google-internal systems.  "Validated with ATA" is fine -- so it might be that the Digital Building Registry is provided?

docs/tech_primer.md
```diff
@@ -0,0 +1,157 @@
...
+## Device Qualification 
+
+[Device qualification](https://github.com/faucetsdn/daq/blob/master/docs/qualification.md)
+qualifies device types as meeting a baseline connectivity and security
+requirement by performing a series of predefined tests against a device.
+Qualification can be automated using the [DAQ tool](https://github.com/faucetsdn/daq/). 
+The output of DAQ is a report. 
+[example report](https://github.com/faucetsdn/daq/blob/master/docs/device_report.md))
+
+### Required Actions
```
grafnu, line 100: The wording of this is problematic since it's getting into "process" rather than technical considerations. This needs to be stated in the form of "what needs to exist for this to proceed" and then "what needs to happen to know that it is completed"

docs/tech_primer.md
noursaidi: I think the intent was that electrically speaking some building "devices" **are** just passive components (e.g. a thermistor), but in this context I think we can remove it as it's _analogues_ with analogue (they're both considered analogue inputs/outputs and utilise 0-10V/4-20mA)

docs/tech_primer.md
noursaidi: I've removed the breakdown of 'proxied' as reading it again it's confusing and I don't think its needed here

docs/tech_primer.md
noursaidi: The idea here is a device installed in a building which communicates over some proprietary protocol to a SaaS platform. This platform has an API to get data from the device. A connector may be written which reads/interacts from this API and publishes data over MQTT/UDMI. The connector may be running as a virtual machine somewhere. This means some of the physical aspects may not be applicable. I think the desire is UDMI is supported from devices, so this use case may be an exception and can be removed from this document

noursaidi: Additional note - this may apply to VT/lift suppliers in future projects, or some IoT sensor products.

docs/tech_primer.md
noursaidi: "How devices are represented within different information sources"? 

Reading it again it's more about the device (logical, reporting, virtual) rather than data?

docs/tech_primer.md
grafnu: So separate out the technical aspects (should be included) of what you said, with the administrative ones (should not be included). "Third party" is administrative, not technical.  I think what you're describing is something like "external" -- which is indicating that the data source is _external_ to the "confined" system. Or something like that that characterizes the architecture that you outline.

docs/tech_primer.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Devices Classification
+
+Various adjectives qualify the different flavors of a "device" found in a building. More than one can apply, and each has a specific technical meaning/definition (i.e. they are not subjective).
+
+### Connectivity
+
+How data is moved around on-prem
+
+-   **Analog**: The device uses an analog connection (1-10V, 4-20mA etc)
```
grafnu, line 33: I think we should take out "Analog" for now.  It's not reading right, and I think you're blending the line between what is considered a "device" here.  Practically speaking, in the analog case the "device" is the thing with the ADC that's making it digital. It's not the same use of a "data moving around" as the other uses.  I'd rather keep this clean for now and add it in later if it's necessary. Unless you have a clear articulation of why it's _necessary_ to have this defined here -- but our goal is not to detail every kind of connection possible, rather the ones with salient meaning (e.g. people talk about "serial" devices all the time).

noursaidi: I think it should be listed here because analog devices are a key part of building control and include things like actuators, temperature & flow sensors, etc. Although they are always connected to an ADC, which will either implement UDMI or connect over some serial interface to a gateway which does; I expect it to be common there are reporting devices which comprise only of points from multiple analog field devices, or represent a single analog field device. Therefore having the line in the tech primer makes it immediately clear that there is 'support' in UDMI for these sorts of devices.

In the industry the analog devices would be known as field devices, so the line could be  `A field device connected using an analog connection`?

docs/tech_primer.md
grafnu: I think it's about "device relationships and how they are modeled"... ?

docs/tech_primer.md
grafnu: What you're describing, tho, is something fundamentally different than the other things.  You're talking about how to access a single point, while everything else is at a higher level (e.g. there's commonly the concept of an address or ID you have with serial and networked connections, but not so with analog). Practically speaking, I have never seen anybody discuss "analog" connections while doing a BOS design. Overall, UDMI doesn't expose things just because they're "industry standard" -- since that's rife with pain... rather, trying to focus on semantically what adds value to the concept.  Can you give an example of a discussion where the concept of "analog" is relevant in the same context as talking about "UDMI"?

noursaidi: All points true and I cannot give an example of a discussion where it's relevant in the same context as UDMI. I've removed it

docs/tech_primer.md
noursaidi: I used the words `**External** A device (on-prem or off-prem) consumes data from an external source and publishes it to Cloud IoT core`

docs/tech_primer.md
```diff
@@ -46,10 +53,12 @@ How data gets up into the cloud
     -   Bind & attach of proxied devices
     -   Encapsulation anti-pattern
 -   **Proxied**: A device that is not direct and is managed through a gateway.
+-   **External**: A device (on-prem or off-prem) consumes data from an external
+    source and publishes it to Cloud IoT core
```
grafnu, line 57: take out the "and publishes it to Cloud IoT core" part -- once you're doing external stuff there are other paths of ingress. But, I thikn the "External" moniker still still apples.

docs/tech_primer.md
```diff
@@ -20,20 +20,27 @@ For the most part, this document addresses the scope of a Smart Ready building:
 
 ## Device
 
-Hardware that exists in the building that can change state in a way that can be machine communicated. A potentiometer or a flow switch is a device if it is monitored by a digital system. A commissioning valve or damper with no communication is not a device. 
+Hardware that exists in the building that can change state in a way that can be
```
grafnu, line 23: This isn't quite accurate.  A "device" is a logical entity that encapsulates a coordinated set of functionality.  The problem is it's a very over-loaded term and so you can't say much in the way of definitions.  Not all devices are hardware. Not all devices an be machine communicated (although it would be kinda pointless otherwise). 

I'm not sure trying to define "device" will actually get you very far... which is why I jumped strait to the descriptive adjectives.

docs/tech_primer.md
```diff
@@ -20,20 +20,27 @@ For the most part, this document addresses the scope of a Smart Ready building:
...
+machine communicated. A potentiometer or a flow switch is a device if it is
+monitored by a digital system. A commissioning valve or damper with no
+communication is not a device. 
 
 ## Devices Classification
 
-Various adjectives qualify the different flavors of a "device" found in a building. More than one can apply, and each has a specific technical meaning/definition (i.e. they are not subjective).
+Various adjectives qualify the different flavors of a "device" found in a
+building. More than one can apply, and each has a specific technical
```
grafnu, line 31: "found in a building" ==> "present in the system"

PR #209, opened 2021-12-06
Make gendocs git friendly

noursaidi:
Issue #208

Try to make gendocs more git friendly and avoid merge conflicts by removing times from footer and fixing the version of the json-schema-for-humans library so that the HTML is not minified.

Alternatively can we setup gendocs and the other code generators be setup to run as GitHub Actions when there is a merge which changes the schema JSON files?


bin/gendocs
```diff
@@ -22,6 +22,12 @@ if [ -d venv ]; then
     source venv/bin/activate
 fi
 
+# Ensure git friendly files are generated by enforcing version
+if ! pip3 freeze | grep -q "json-schema-for-humans==0.39.1"; then
```
grafnu, line 26: this version should be explicitly specified in the etc/requirements.txt file -- I'm not sure a runtime check here is appropriate.  They do slightly different things, but generally speaking the requirements.txt way is the standard way so an explicit check isn't normal/expected.

bin/gendocs
```diff
@@ -33,7 +39,7 @@ mkdir -p $TMP_DIR
 cp -r schema/. $TMP_DIR
 
 # Remove file: prefix from references so document generator works
-find $TMP_DIR -type f -exec sed -i 's/file://' {} \;
+find $TMP_DIR -type f -exec sed -i  's/file://' {} \;
```
grafnu, line 42: extra space

grafnu (commented):
The gitHub actions thing is a neat idea -- but I'd leave that as a separate activity (form this PR)...

bin/gendocs
grafnu: what's the utility of keeping this check here?   Usually we just rely on the requirements.txt enforcement b/c it keeps it uniform and simple. If there's some reason to have this explicit check here, then it should at least be clarified as to _why_ its there in the comment.  ("why" as in why the explicit check, not the explicit version.)

noursaidi: @grafnu I removed this check from - you might have missed it when you did the re-review.

grafnu: sigh -- yes -- I don't find GitHub's interface particularly easy to use!

grafnu (approved):

PR #196, opened 2021-11-19
Docs update

pisuke:
Work in progress proposal to make the documentation easier to browse.

PR #210, opened 2021-12-07
Rearrange Documentation Folder

noursaidi:
Builds on PR #196 to rearrange the documentation folders.

readme's for the new directories to come in a future update, as will documentation for the new items in the list

Comments on the PR were incorporated. Additional changes were also:
- Renamed messages to messaging like status, pointset aren't messages but are part of the messages; and we also have the schema)
- Moved metadata out of the messages directory
- Renamed dashboard to cloud
- Renamed `config writeback` to `writeback`
- Renamed `implementation guide` to `guides`


grafnu (commented):
Overall looks good, thanks. I think a few top-level names are inconsistent and should be munged just a bit:

specification --> specs
messaging--> messages
transactions --> sequences (nothing inherently wrong with transactions, but the testing code currently calls them sequences so trying to be consistent)
cloud --> clouds/gcp

Also, I wouldn't put "messages" under "specification" -- I think messages is significant enough to be its own top level entry.

I'm not sure the "applications" hierarchy is useful... there are tools (e.g. pubber.md) and then cloud stuff (clouds/gcp) -- but the cloud stuff isn't really tools, it's a complete ecosystem.

PR #216, opened 2022-01-06
Predefined status/log entries

noursaidi:
Published as a draft for now for discussion and to build on.

I think there is a need to have a list of predefined system status logs/log entries. Already in UDMI, there are some are defined (for write back), but there's benefit in consolidating these into one file as a reference, and to add additional information such as log levels.

Additionally, we're now asking device manufacturers to send out some logs (e.g. a system start up log entry) but we are not defining the details of this. 

Without doing so, different devices may represent the same issues/logs/status differently

This attempts to consolidate the log entries and define the minimum log entries needed


noursaidi:
Started to be addressed in #248

PR #230, opened 2022-01-28
Add connection model documentation

grafnu:


docs/specs/connecting.md
```diff
@@ -0,0 +1,115 @@
+# Connecting Devices to Cloud
```
noursaidi, line 1: Breadcrumbs - [**UDMI**](../../) / [**Docs**](../) / [**Specs**](./) / [Connecting](#)

grafnu: Added, also updated the tool to check for breadcrumbs (limited form)

PR #236, opened 2022-02-04
Fix codegen disconnect and add auto-check

grafnu:


grafnu:
Note part of the PR is reverting a schema change to make the "testing targets" more explicit. The problem was that that schema change broke some of the underlying testing code... thus then adding in the check to make sure tests were being run with a new version of generated code when the schema changed. The schema change itself is possibly OK, but would require going through and fixing the code as well, so was more work than I wanted to get into at this time (since there's other changes that were blocked by this)!

johnrandolph (approved):

PR #240, opened 2022-02-07
cloud infrastructure automation

pisuke:
@grafnu, @noursaidi  here is a PR to start adding automation to create UDMI cloud infrastructure.
I've tried to make the code as generic as possible.
Things missing: cloud functions and bootstrapping of the cloud project, that I will address in another, later PR.
Let me know if you have any comments, happy to demonstrate how it works with a live session too.


cloud/readme.md
```diff
@@ -0,0 +1,63 @@
+# UDMI terraform configuration files
```
grafnu, line 1: should this go in docs/cloud/gcp?

cloud/readme.md
```diff
@@ -0,0 +1,63 @@
...
+The next step is to initialize the required terraform backend, provider and modules:
+
+```
+terraform init
+```
+
+It is then necessary to import the project and the bucket previously created to the terraform state:
+
+```
+terraform import google_project.udmi-project GCP-PROJECT-NAME
```
grafnu, line 46: ${GCP_PROJECT_NAME} so it's very clear that it's a variable (also makes for easy cut-and-paste of commands)

cloud/readme.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Getting started
+
+To get started, copy the `terraform.tfvars.template` file to `terraform.tfvars`
+
+```
+cp terraform.tfvars.template terraform.tfvars
+```
+
+and edit it according to the project settings.
```
grafnu, line 27: personally I find this a bit too disjointed in terms of individual steps, I'd propose bigger blocks that have multiple commands per block. Often when I'm following a guide like this I just like to cut-and-paste, and if there's more in one block (e.g. all the cp commands) then it can be one step.

cloud/readme.md
```diff
@@ -0,0 +1,63 @@
...
+```
+cp udmi-site.tf.template udmi-site.tf
+```
+
+and edit it according to the UDMI site you need to create and group permissions you want to attribute to it.
+
+The next step is to initialize the required terraform backend, provider and modules:
+
+```
+terraform init
```
grafnu, line 40: similar here, I'd combine the init & import steps into one block for easy cut-and-paste

grafnu (commented):
How GCP-specific is all this?  If it's "mostly" GCP at this point, I'd just go ahead and put it in cloud/gcp to keep the structure with (e.g.) docs/cloud/gcp -- not a big deal, we can always move it later if desired.

I'm not sure cloud/auth/put-credentials-json-file-here is useful -- since it doesn't actually tell me what to name the file... So you could have a credential file named appropriately that contains obvious junk in it ("Replace me with credentials downloaded from GCP") -- or just mention it in the docs?

cloud/readme.md
pisuke: > 



> How GCP-specific is all this? If it's "mostly" GCP at this point, I'd just go ahead and put it in cloud/gcp to keep the structure with (e.g.) docs/cloud/gcp -- not a big deal, we can always move it later if desired.

Yup, good idea about moving all to cloud/gcp, I've done it

pisuke:
> I'm not sure cloud/auth/put-credentials-json-file-here is useful -- since it doesn't actually tell me what to name the file... So you could have a credential file named appropriately that contains obvious junk in it ("Replace me with credentials downloaded from GCP") -- or just mention it in the docs?

I've created a `credentials.json.template` file. I prefer not to put the `credential.json` named file as it is included in .gitignore and I can keep a proper `credentials.json` file that I use for testing that is ignored by github in that folder.

cloud/readme.md
pisuke: > should this go in docs/cloud/gcp?

I'm a bit ambivalent about this. I know we have all the other documentation inside docs, so I guess the answer is yes, but it's also nice to get into that folder and read how to use it. What do you think?

cloud/readme.md
pisuke: sure, done.

cloud/readme.md
pisuke: done

cloud/readme.md
pisuke: done

cloud/readme.md
pisuke: I checked with @noursaidi and we've added a link [here](https://github.com/faucetsdn/udmi/blob/b3f9d8c545253d23f2c1cde6f569b8dc4489fef9/docs/cloud/gcp/readme.md?plain=1#L8)

noursaidi: There is also the [`dashboard`](https://github.com/faucetsdn/udmi/tree/master/dashboard) which is also GCP only, so a `cloud/gcp` to evetnually house the `dashboard` directory makes sense

grafnu: I'm not sold on the location of the readme.md file. Part of the reason to have all the docs in one place is for discoverability and automatic tool processing. As soon as they're spread all over then things get a lot more complicated!  It's generally a Bad Idea (TM) to have links that go up-and-over except in some categorical cases (i.e. from documentation to examples, or documentation to code).

grafnu (commented):
I'm not sold on the location of the readme.md file. Part of the reason to have all the docs in one place is for discoverability and automatic tool processing. As soon as they're spread all over then things get a lot more complicated! It's generally a Bad Idea (TM) to have links that go up-and-over except in some categorical cases (i.e. from documentation to examples, or documentation to code).

What do you think about linking the other way around? Have the main body of documentation be in docs/ and then a non-docs/ readme.md pointing back at the other one in docs/?

cloud/readme.md
pisuke: I've just moved the docs to `docs/cloud/gcp` where @grafnu likes them to be. My rationale is that it's easier to find a helpful note (especially for use and installation) directly in the folder of the code, rather than have to search for it in the docs folder. More formal documentation can still be in docs, but it has a different purpose. Anyway, moved to `docs/cloud/gcp` now.

grafnu (approved):

PR #243, opened 2022-02-14
Terraform - global storage bucket name

pisuke:
A small PR to ensure that the storage bucket name for the terraform config is always unique

### DIFF
--- a/cloud/gcp/.gitignore
+++ b/cloud/gcp/.gitignore
@@ -1,0 +1,2 @@
+auth/credentials.json
+.terraform*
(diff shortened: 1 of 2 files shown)


grafnu (approved):

grafnu (commented):
There's also a place in main.tf where it's referenced!

pisuke:
> There's also a place in main.tf where it's referenced!

Good catch! This is annoying though as I cannot put a variable in there (not allowed) so it might mean editing this file too, which I wanted to avoid.
I'll do a bit more research to see if it's possible to edit the storage bucket name into one single place.

grafnu:
'm creating a script to do all this file editing...

PR #246, opened 2022-02-15
Add basic secrets processing for GitHub integration

grafnu:


cloud/gcp/bin/clean
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -e
+
+echo Cleaning existing state...
+
+terraform state rm `terraform state list`
```
pisuke, line 5: The terraform state is kept in the remote backend bucket, it must not be erased

grafnu:
This is just a utility function (clean) -- if you don't call it then it
won't be used/removed!   Maybe it's not useful... I still can't get my
system to work so it's not clear to me what *does* need to be cleaned out
to get to stable state.

cloud/gcp/bin/init
```diff
@@ -0,0 +1,18 @@
...
+for file in main.tf udmi-sites.tf terraform.tfvars; do
+    if [[ ! -f $file ]]; then
+        echo Configuration file $file does not exist, did you run bin/setup?
+        false
+    fi
+done
+
+terraform init $*
+
+terraform import google_project.udmi-project ${GCP_PROJECT_NAME:?}
```
pisuke, line 16: This must be removed because the state already includes the project

cloud/gcp/bin/init
```diff
@@ -0,0 +1,18 @@
...
+
+terraform import google_storage_bucket.tf-bucket ${GCP_PROJECT_NAME}-terraform
```
pisuke, line 18: This must be removed because the state already includes the storage bucket

grafnu:
OK -- this was what you had in the readme!

pisuke:
> This is just a utility function (clean) -- if you don't call it then it won't be used/removed! Maybe it's not useful... I still can't get my system to work so it's not clear to me what *does* need to be cleaned out to get to stable state.

It looks dangerous to me. If we lose the state, we must start from scratch, which is not what we want. I would remove it to avoid misuse.

grafnu:
Ok -- I took it out for now, but at some state we'll need something that
cleans out the local state on a machine -- not sure what that would be,
exactly.  Maybe it's removing the ".terraform" directory?  Basically, it
needs to undo whatever might have been done with the local setup... (but
that can be added later)

PR #248, opened 2022-02-17
Bad config sequence checks

grafnu:


docs/tools/troubleshooting.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Cloud IoT Core Dashboard
+
+### Site Registry
+
+* udmi_target
+* udmi_state
+* device-under-test
+
+### UDMI-Reflect Registry
```
pisuke, line 13: Shouldn't this be UDMI-REFLECT (all capitals)?

dashboard/functions/index.js
```diff
@@ -222,7 +243,7 @@ function update_device_config(message, attributes) {
     versionToUpdate: version,
     binaryData: binaryData
   };
-  const commandFolder = `devices/${deviceId}/update/config`;
+  const commandFolder = `devices/${deviceId}/update/configs`;
```
pisuke, line 246: Maybe I'm missing something, but why has this folder name changed to configs (plural)?

johnrandolph (approved):

grafnu:
As I was working through things I realized some of the messages were
mislabelled and it was causing much confusion. Right now, then, there are a
few command folders that get used, e.g.:

   - devices/${deviceId}/update
      - config -- config for *just one* subblock (e.g. pointset)
      - configs -- *all* device configs (essentially what is written to a
      device)
      - state -- state for *just one* subblock (e.g. pointset)
      - states -- *all* device states
      - event -- events, which are always segmented by subblock
      - query -- ability to query something in the system (e.g. query
      device state)

My goal with the PR was to clean things up as much as possible to make them
more coherent in the long-run, but it does require some tweaking along the
way. All this stuff is effectively in alpha so breaking changes are
somewhat expected/allowed if they get us ot a better state in the long run.
It's a bit confusing because there's effectively two worlds out there --
one where "state" is a monolithic block, as enforced by IoT core that
extends down to devices, and the other in backend systems where things are
neatly segregated by subblocks. There is no easy way around it, and
somewhere something has to do an impedance adjustment!

grafnu:
Yes, updated.

PR #254, opened 2022-03-01
Updates to hardware/firmware version reporting

grafnu:


grafnu:
Again, my apologies for a large CL (number of files) -- but it ended up touching a lot of auto-generated code and test cases... not sure how to avoid that. But, it should be pretty easy to "ignore" a bunch of the files since they're derivative changes (not manually written code).

grafnu:
Ok, I've extracted some of the auto-generated changes into a separate branch, so this PR now only contains the "manual" changes that might need a review

validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
```diff
@@ -632,16 +632,44 @@ private JsonSchema getSchema(File schemaFile) {
...
+  private File getTargetPath(String baseFile, String addedFile) {
+    File prefix = new File(baseFile);
+    File outBase = new File(prefix.getParentFile(), "out/tests");
+    File full = new File(outBase, addedFile);
+    full.getParentFile().mkdirs();
+    return full;
+  }
+
+  private void upgradeMessage(String schemaName, JsonNode jsonNode) {
+    new MessageUpgrader(schemaName, jsonNode).upgrade();
```
johnrandolph, line 670: Do we want more robust handling of messages that don't upgrade?

etc/validator.out
```diff
@@ -18,6 +18,7 @@ While converting to json node: 2 schema violations found
   2 schema violations found
     instance value ("states") not found in enum (possible values: ["update","discovery","system","gateway","localnet","metadata","pointset","blobset"])
     instance value ("update") not found in enum (possible values: ["event","command","state","config"])
-While converting to json node: 1 schema violations found
-  1 schema violations found
+While converting to json node: 2 schema violations found
```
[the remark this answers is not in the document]
grafnu: yes it should be -- it's really the "expected out" file that the tests use to make sure the output is as expected.

validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
grafnu: After the upgrade it's run through the current schema validator to make sure the upgraded message is proper, so if "didn't upgrade" means that there are still lingering fields or something it will be caught. I'm not sure what additional handling would be useful... might depend on the specifics of the "didn't upgrade" (like, missing parameter, wrong parameters, etc...).

johnrandolph: I think that "its run through the current schema validator" is at least consistent enough behavior for the user so I'm fine with that.

johnrandolph (approved):

PR #262, opened 2022-03-08
If device names are supplied on argv, process only these devices

johnrandolph:
In the case that device names e.g. AHU-1 are supplied on argv to registrar, all devices are still processed.
This change filters devices to only match the ones requested.

### DIFF
--- a/validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
+++ b/validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
@@ -516,14 +516,14 @@ public class Registrar {
       throw new RuntimeException("Not a valid directory: " + devicesDir.getAbsolutePath());
     }
 
-    String[] devices;
+    final String[] devices;
     if (specifiedDevices == null) {
       devices = devicesDir.list();
     } else {
       devices = devicesDir.list(new FilenameFilter() {
-      public boolean accept(File dir, String name) {
-        return specifiedDevices.contains(name);
-      }
+        public boolean accept(File dir, String name) {
+          return specifiedDevices.contains(name);
+        }
       });
     }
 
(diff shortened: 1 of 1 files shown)


validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
```diff
@@ -508,13 +509,24 @@ private void shutdown() {
...
-  private Map<String, LocalDevice> loadLocalDevices() {
+  private Map<String, LocalDevice> loadLocalDevices(Set<String> specifiedDevices) {
     Preconditions.checkNotNull(siteDir, "site directory");
     File devicesDir = new File(siteDir, DEVICES_DIR);
     if (!devicesDir.isDirectory()) {
       throw new RuntimeException("Not a valid directory: " + devicesDir.getAbsolutePath());
     }
-    String[] devices = devicesDir.list();
+
+    String[] devices;
```
grafnu, line 519: declare final (just add the "final" keywoard before String[]) -- this just means that it should be assigned once, and exactly once, and then is just enforced by the compiler.

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
```diff
@@ -508,13 +509,24 @@ private void shutdown() {
...
+    if (specifiedDevices == null) {
+      devices = devicesDir.list();
+    } else {
+      devices = devicesDir.list(new FilenameFilter() {
+      public boolean accept(File dir, String name) {
```
grafnu, line 524: something seems off about the indent... the "public" should not line up with "devices"  -- the lint checks are complaining because of "  Checkstyle violations by severity: [warning:151]" -- and I think this is because they are capped at 150 (previous high-water mark).  You should be able to look through the logs there to see exactly what the errors are (search for the filename). Ideally you'd also see these in the IDE or running "validator/bin/build check" locally on the command line.

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
johnrandolph: Done

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
johnrandolph: Fixed this which should -1 lint problem also improve -1 more lint by moving null overload adjacent

grafnu (approved):

PR #264, opened 2022-03-10
Add min_loglevel, sample_{limit,rate}_sec to pointset schema and config

johnrandolph:


grafnu (approved):
Code looks fine, just need the tests to pass!

PR #267, opened 2022-03-15
Replace system.statuses with system.status

grafnu:


bin/test_schema
```diff
@@ -10,21 +10,28 @@ rm -f $errorfile
...
+            subsets=${OPTARG}
+            ;;
         d)
             testdir=${OPTARG}
             ;;
+        p)
+            parallel=y
+            ;;
         \?)
             echo "Usage: $0 [-f] [-n] [-d TEST_DATA_DIR]"
```
johnrandolph, line 36: Update the usage help?

johnrandolph (commented):
[ test reviewing one commit in a PR ]  Did not work

validator/src/main/java/com/google/daq/mqtt/validator/PointValidator.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.daq.mqtt.validator.validations.SkipTest;
+import com.google.daq.mqtt.validator.validations.WritebackValidator;
+import java.util.HashMap;
+import java.util.Optional;
+import org.junit.Before;
+import udmi.schema.PointPointsetConfig;
+import udmi.schema.PointsetConfig;
+import udmi.schema.TargetTestingMetadata;
+
+public abstract class PointValidator extends SequenceValidator {
```
johnrandolph, line 12: Is this the writeback test?

noursaidi:
@grafnu so the change is that instead of the incumbent system `statuses` which comprises an array of `status` objects, there is only ever a single status for a device?

grafnu:
There's no good answer to the question "is there only a single status" -- I
can slice and dice systems any arbitrary number of ways to get whatever
answer we want. In the degenerate form, I can always come up with an
algorithm to distill multiple down to one, so "yes" there can be a single
status. All the cases I've been tracking I figured it was better to have
multiple "status" blocks around in their context-relevant location (e.g.
system, pointset, point-of-pointset, etc...) -- and so that's one way to
handle multiple statuses. E.g. for discovery I'm adding a discovery status.
The problem with the array/map of statuses is that then somebody needs to
define the categories and figure out how to distill things down, because
generally people just want to know "what's the status of the device?" (as
an end-user).  It's also easier to add in more later of various forms if
necessary, but the existing map was throwing a lot of complexity into the
system as I was trying to build out some parts (e.g. config tests).

grafnu:
It's a base-clase for "tests that deal with points" -- but the Writeback
tests themselves are in a different subclass in WritebackValidator.java

johnrandolph (approved):

noursaidi:
What do we do about the devices (few - maybe 1?) which implement the `statuses` block presently - if it's removed from the schema it will bring up a validation error (which we can clear with an exceptions) - but then do we still care that it's that it's correct? Maybe for schema-_breaking_ changes we could use the [deprecated field in JSON Schema](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.3) so changes can still be validated but marked as so?

grafnu:
I'm not sure what value the "deprecated" field has, other than an indicator
somewhere of our intent -- but I doubt people would actually pay attention
to it... just not sure what the effective workflow there would be.
Basically, if the json validates ok people aren't going to do anything,
essentially rendering the "deprecated" flag moot. I'd rather have the more
obvious error that has to be explicitly called out as an exception so it's
more obvious what "fixes" are required. Also, it would help keep the schema
more clean (less random cruft in the documentation).

My thought was exactly what you cited... The validator will flag it as a
validation error, but the *interpretation* of that is somewhat site/project
dependent. So, then some site-specific mechanism can be used to allow
certain exceptions.  Mostly I'm expecting some engineer somewhere to be
annoyed by this, because it means some more work for them, but I think it's
easy enough to manage if we get the exception mechanism flow in place
(which I believe we have to do anyway).

At some point, we'll likely need/want to be more careful with potentially
breaking changes... but... since the installed base is so small and a
proper definition of a "Smart Ready" device was never really ratified in
the first place, I'd like to avoid keeping cruft in the spec "just because
maybe"...

PR #268, opened 2022-03-16
Adding discovery schema only (not sequencer)

grafnu:


tests/event_discovery.tests/implicit.json
```diff
@@ -0,0 +1,37 @@
+//
+// Discovery result for a network scan (e.g. BACnet discovery scan), including
+// an implicit object point enumeration.
+//
+// Sent on MQTT topic: /devices/{device_id}/events/discovery
+// where {device_id} is that of the discovery node (e.g. gateway). The {device_id}
```
noursaidi, line 6: The last sentence is confusing because it uses the same {device_id} - I assume it's to say that if a discovered device has a IoT ID, then this is provided in the families.iot.id field but the message is always published to the discovery node (/devices/{discovery node id}/events/discovery)?

schema/event_discovery.json
```diff
@@ -13,23 +13,53 @@
...
+      "$ref": "file:common.json#/definitions/entry"
     },
     "families": {
-      "description": "Collection of address families identifying the scan target",
+      "description": "Address family results for a scan. Not included for device enumeration messages.",
+      "type": "object",
       "additionalProperties": false,
       "patternProperties": {
-        "^[a-z][a-z0-9]*(_[a-z0-9]+)*$": {
+        "^iot|bacnet|ipv4|ipv6|ethmac$": {
```
noursaidi, line 34: Should we explicitly define these as properties? e.g. in https://gist.github.com/noursaidi/f8431051eed682de0fd967cd8745f68b ? 
Allows for validation and avoids ambiguity, e.g. the BACnet ID you gave was hexadecimal, generally it's number but I've seen it as a hexadecimal in the documentation for an older system. Similarly to be pedantic about Mac addresses, I've seen both upper and lower case, delimited by either : or - . 

Even though it's easy to still interpret the results, where I think it's important is building protocols and gateways. A gateway configuration which uses a hexadecimal ID will stop working if the physical gateway is replaced with a gateway which uses a numeric ID, unless we specify _a BACnet gateway should support both a numeric and hexadecimal ID, but then it's just easier to pick one and stick to it?

tests/event_discovery.tests/implicit.json
grafnu: Yes -- it's tricky because it's some nuanced information that isn't going to be explained in a single comment. I've tried to make it a bit better, but I think here it just needs to be technically accurate, but it's not trying to explain everything.

schema/event_discovery.json
grafnu: We want to be as specific/pedantic as possible about the various formats, otherwise it's crazyland. E.g. a bacnet address is really just a 24-bit number, and we should stipulate that it's always an UPPER-CASE HEX Address.  That said, I'm not sure the schema is the right place to enforce all of that (rather than in code somewhere), since it can get really ugly really quickly. So, "explicitly define them as properties" -- sure, but I'd still want them to be homogenous properties, not different (too hard to maintain).

It wasn't clear from your comment exactly what you were proposing... (define as properties vs. heterogeneous definitions).

docs/specs/discovery.md
```diff
@@ -1,23 +1,61 @@
 [**UDMI**](../../) / [**Docs**](../) / [**Specs**](./) / [Discovery](#)
 
-# Discovery and Enumeration
+# Discovery
 
-There's two related processes that are used to extract the
-'as built' state of a system, detailing the physical actualities
-of a physical system:
+Discovery consists of two related processesd to describe the 'as built'
```
johnrandolph, line 5: low_nit: typo

docs/specs/sequences/discovery.md
```diff
@@ -0,0 +1,113 @@
...
+  for scanned devices (keyed by matchhing `generation` field): one _event_ for each unique device scanned.
+* [_stop state_](../../../tests/state.tests/scan_stop.json): Once complete, the _active_ field is `false`
+  (or removed). Ideally the `generation` field would remain to indicate the last scan performed.
+
+At this point, the _config_ `generation` entry can be removed with no effect, or updated to initiate a new scan.
+
+## Periodic Scan
+
+A _periodic scan_ is like a _sporadic scan_ except that the scan automatically occurs due to a predefined
+interval (rather than individual tirgger _config_s). This allows for repeated scans without any _config_ changes.
```
johnrandolph, line 43: nit: trigger

docs/specs/sequences/discovery.md
```diff
@@ -0,0 +1,113 @@
...
+  * Unlike the _sporadic_ case, the `generation` field will be greater than or equal to the _config_ specification.
+  * Loop terminates when either the `generation` or `scan_interval_sec` parameter is removed from _config_.
+
+Note that the _scanning_ should occur at intervals directly determined by the _config_ `generation` timestamp plus
+integral increments of the scan _interval_, i.e. _Ts = Tc + N*Ti_, so that there is no clock drift.  E.g., it
+should be possible to setup a schema to scan every day exactly at midnight.
+
+## Continuous Scan
+
+A _continuous scan_ is the mode for a system that can passivly monitor traffic and deduce scan results, so
```
johnrandolph, line 58: nit: typo passively

schema/event_discovery.json
johnrandolph: Would it be possible for one device to participate in multiple address families?

schema/common.json
```diff
@@ -45,6 +45,9 @@
         "timestamp",
         "level"
       ]
+    },
+    "families": {
+      "enum": [ "bacnet", "ipv4", "ipv6" ]
```
johnrandolph, line 50: Why isn't this   +=  [ "iot", "ethmac" ]   ?

schema/event_discovery.json
grafnu: Yes, but that is supported by the "patternProperties" block. What this says is that "all properties in this block must match this given pattern" -- so, as long as the pattern can match different key names, multiple properties can exist. In this case, e.g., { iot, bacnet } would both match, so it would allow having two properties "iot" and "bacnet"...

bin/loop_sequences
```diff
@@ -47,6 +48,12 @@ for test_class in ConfigValidator WritebackValidator; do
         echo Sequence test $test_class exited with error code $result
         exit_code=$result
```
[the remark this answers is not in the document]
grafnu: I think the rest of the code loop is still value -- there might be intermediate files generated by the testing that are useful for diagnostic/debugging, which the remainder of the loop essentially processes.

schema/common.json
grafnu: dead code -- wasn't used at all -- removed!

schema/event_discovery.json
johnrandolph: Ok, that's great.

I think the schema is a better place to express any formatting constraints because it will require less digging to see the format, and it can't drift if code and docs are out of sync.

johnrandolph (approved):

PR #271, opened 2022-03-16
Functionality-neutral refactor (mostly lint fixes)

grafnu:


pubber/src/main/java/daq/pubber/PubSubClient.java
```diff
@@ -13,13 +13,22 @@
 import org.slf4j.LoggerFactory;
 import org.threeten.bp.Duration;
 
+/**
+ * Wrapper class for a PubSub client.
+ */
 public class PubSubClient {
 
-  private final static Logger LOG = LoggerFactory.getLogger(PubSubClient.class);
+  private static final Logger LOG = LoggerFactory.getLogger(PubSubClient.class);
```
johnrandolph, line 21: NOP:  oof

validator/build.gradle
```diff
@@ -31,7 +31,7 @@ sourceSets {
 
 checkstyle {
     ignoreFailures = false
-    maxWarnings = 150
+    maxWarnings = 66
```
johnrandolph, line 34: What is this, threshold of current warnings?

johnrandolph (approved):

grafnu:
What does "NOP: oof" mean?

Yes, the maxWarnings is the threshold... idea is that it should only go
down as we either improve things or not go up as we add things!

PR #272, opened 2022-03-16
Fix subType streaming validator errors

grafnu:


bin/validator
```diff
@@ -22,4 +25,9 @@ $ROOT_DIR/validator/bin/build > /dev/null
 
 echo Running tools version `(cd $ROOT_DIR; git describe)`
 
-$ROOT_DIR/validator/bin/validate $project_id schema pubsub $subscription $site_path
+if [[ -z $ubscription ]]; then
```
noursaidi, line 28: type - $subscription

.github/workflows/testing.yml
```diff
@@ -57,18 +57,28 @@ jobs:
...
           GCP_TARGET_PROJECT: ${{ secrets.GCP_TARGET_PROJECT }}
         if: "${{ env.GCP_TARGET_PROJECT != '' }}"
         run: |
           bin/test_sequencer $GCP_TARGET_PROJECT
-          fgrep 'RESULT ' /tmp/test_log.txt | sed -e 's/.*sequencer RESULT/RESULT/' > /tmp/results.txt
-          diff /tmp/results.txt etc/sequence.out && echo No output diff detected.
-      - name: validation logs
+          fgrep 'RESULT ' /tmp/test_log.txt | sed -e 's/.*sequencer RESULT/RESULT/' > /tmp/sequencer.out
+          diff -u /tmp/sequencer.out etc/sequencer.out && echo No output diff detected.
+      - name: device ouput logs
```
johnrandolph, line 81: nit: output

bin/test_validator
```diff
@@ -0,0 +1,68 @@
...
+
+site_path=udmi_site_model
+device_id=AHU-1
+
+serial_no=validator-$RANDOM
+echo Using pubber with serial $serial_no
+
+PUBBER_OUT=pubber.out
+
+pids=`ps ax | fgrep pubber | fgrep java | awk '{print $1}'`
```
johnrandolph, line 21: This feels heavy handed? Is this to strengthen repeated runs of this test, or to stop interference with other running tests?

noursaidi (approved):

bin/test_validator
```diff
@@ -0,0 +1,68 @@
...
+    kill $pids
+fi
+
+rm -rf out/devices
+
+# Prepare auth key used by reflector
+cp udmi_site_model/devices/AHU-1/rsa_private.pkcs8 \
+   udmi_site_model/gcp_reflect_key.pkcs8
+
+bin/validator $site_path $project_id &
```
johnrandolph, line 33: Should we trap on exit of validator so we know if it exits while running pubber?

bin/test_validator
```diff
@@ -0,0 +1,68 @@
...
+
+if [[ $i == $WAITING ]]; then
+    echo pubber startup failed:
+    cat $PUBBER_OUT
+    false
+fi
+
+echo Waiting for system to run for a bit...
+sleep 30
+
```
johnrandolph, line 61: If this becomes more cumbersome, or we get unpredictable test results, I suggest using the shell's jobs command to focus our kill usage

For example:

jobs -x kill %pubber
jobs -x kill %validator

Although I'm not sure whether the shell would see either java or pubber/validator

bin/test_validator
```diff
@@ -0,0 +1,68 @@
...
+pids=`ps ax | fgrep pubber | fgrep java | awk '{print $1}'`
+echo Killing pids $vpid $pids
+kill $vpid $pids
+
+outfiles=`find out/devices -name \*.out` || true
+echo Found .out files $outfiles, copying to /tmp/validator.out
+more $outfiles > /tmp/validator.out || true
```
johnrandolph, line 68: should this be cat?

bin/test_validator
```diff
@@ -0,0 +1,68 @@
...
+if [[ $i == $WAITING ]]; then
+    echo pubber startup failed:
+    cat $PUBBER_OUT
+    false
+fi
+
+echo Waiting for system to run for a bit...
+sleep 30
+
+pids=`ps ax | fgrep pubber | fgrep java | awk '{print $1}'`
```
johnrandolph, line 62: Or at least pids=\`pgrep pubber\` ?

.github/workflows/testing.yml
grafnu: I'll fold this into another PR

bin/test_validator
grafnu: Maybe-- the code as it stands checks the output of what validator does, so it doesn't matter if it exits prematurely or not... Either the generated output is correct or not.

bin/test_validator
grafnu: more includes the filename in the output, so it's convenient for logging since you get more context, rather than just a bunch of stuff combined together

PR #277, opened 2022-03-21
Make gateway error structure more flexible and single test runner

grafnu:


grafnu:
Just making the error structure for gateway reporting more general/flexible...

grafnu:
Also added capability for the -v flag to control the verbosity of the output.  Default is "notification" which is meant to be just high-level useful to somebody running the suite. Using -v means "info" which is just more stuff, and then another -v or -vv puts it into "debug" mode which is all sorts of junk (like stack traces).

pisuke:
I've checked functionalities of the cloud functions. It is now possible to deploy and run them.
LGTM.

pisuke (approved):
LGTM.

noursaidi (approved):
Looks like a good addition

validator/src/main/java/com/google/daq/mqtt/validator/SequenceValidator.java
```diff
@@ -234,6 +234,18 @@ private static Metadata readDeviceMetadata() {
     }
   }
 
+  protected static String getTimestamp(Date date) {
+    try {
+      if (date == null) {
+        return "null";
+      }
+      String dateString = OBJECT_MAPPER.writeValueAsString(date);
+      return dateString.substring(1, dateString.length() - 1);
```
johnrandolph, line 243: Please add a comment that describes what this output will look like.  Is the first char 'T' ?

Issue #193, opened 2021-10-26
Validator not validating incomplete state pointset


noursaidi:
Using the validator, the state_pointset does not raise an error if there are missing points. Conversely the validator correctly identifies missing points from the event_pointset

Example below, can be tested by adding points to the metadata for the pubber device
```
Processing device #1/12: AHU-9/state_pointset
Warning: Nashorn engine is planned to be removed from a future JDK release
Validation complete AHU-9/state_pointset
Processing device #1/12: AHU-9/state_system
Validation complete AHU-9/state_system
Processng device #2/12: AHU-9/event_pointset
java.lang.RuntimeException: Metadata validation failed: Missing points: missing_point
	at com.google.daq.mqtt.validator.ReportingDevice.validateMetadata(ReportingDevice.java:67)
	at com.google.daq.mqtt.validator.Validator.validateUpdate(Validator.java:381)
	at com.google.daq.mqtt.validator.Validator.validateMessage(Validator.java:286)
	at com.google.daq.mqtt.validator.Validator.lambda$messageValidator$0(Validator.java:233)
	at com.google.daq.mqtt.util.PubSubClient.processMessage(PubSubClient.java:112)
	at com.google.daq.mqtt.validator.Validator.messageLoop(Validator.java:261)
	at com.google.daq.mqtt.validator.Validator.validatePubSub(Validator.java:239)
	at com.google.daq.mqtt.validator.Validator.main(Validator.java:124)
Processing device #2/12: AHU-9/event_system

```

grafnu:
Should be fixed now: https://github.com/faucetsdn/udmi/pull/356

PR #286, opened 2022-04-01
Add MessageUpgrader to validator

grafnu:


noursaidi:
The validator doesn't appear to be process state messages with this change - when a state message is published, nothing appears in the command line output or output directory. 
 
I have compared against the current master  (checksum `29530a7c8e265230d5127501feeee7b49811b1d5`) and this does process state messages, so my message flow seems to be working. In both cases I was manually triggering state messages from an older instance of pubber so I'm certain there were state messages being sent

grafnu:
Did you update your cloud functions?  There's been a bunch of changes there
that need to go along with this.  Do I have access to the project? What is
it?  Would just make it easier for me to poke at it!

noursaidi:
I did [deploy the cloud functions](https://console.cloud.google.com/functions/list?project=daq1-273309) - I had to manually add the GCP project ID as an environmental variable in the `dashboard_deploy_gcloud` script as it was coming up undefined (same issue as Francesco)

You have access to my GCP project. You run the validator with the following command
`bin/validator <any site model> daq1-273309 udmi_target`

I have configured a script to publish a static state payload every minute under device [AHU-9](https://console.cloud.google.com/iot/locations/us-central1/registries/registrar_test/devices/details/AHU-9/history?project=daq1-273309) (you do have access to my [site model](https://github.com/noursaidi/nes-site-model) but the payload I'm publishing doesn't match up with the site model)

grafnu:
Ok, I can reproduce the problem... I think I was just so focused on the
config and event messages that status kinda for lost :-).

noursaidi:
I did add `const PROJECT_ID = process.env.UDMI_GCP_PROJECT` in the cloud function, I was testing something with Firebase deploy which is also not working for me. 

Though for this, `FIREBASE_CONFIG` is set and within this there is a field `projectId` with the project ID which can be used in place of `process.env.GCLOUD_PROJECT` for this deploy method

It is now processing state messages and old payloads aren't triggering an error - but behaviour has changed. There used to be separate `state_pointset.out` and `state_system.out` etc, but now there's a single `state.out` - was this expected? It's also still not validating (against metadata) the points in pointset (issue #193)

grafnu:
Expected... kinda?   I wasn't going into the changes with a specific target
in mind, but more of a "trying to make it logically consistent."  Right now
it's validating the entire "state.json" message, which is why there is one
"state.out" file -- it's not individually validating the sub-blocks
(state_pointset.json) -- the sequence tester does that (well, and records
both), but it's operating in a different way.

This PR was primarily about adding the MessageUpgrader to the validator, so
it didn't try to validate against metadata -- that would be something else.

grafnu:
Note that one change that IS for sure expected is the location of the
output files.  They used to be rooted in the udmi/out/ directory, but now
they should appear in the {site_model}/out/ directory -- moving towards the
general expectation that the udmi directory would be "system level" so all
site-specific things would be kept differently.  Validator and sequencer
should work this way now, but I think pubber is till in the old model.

noursaidi (approved):

validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
```diff
@@ -473,7 +484,7 @@ private File makeDeviceDir(String deviceId) {
     return deviceDir;
   }
 
-  private String makeSchemaName(Map<String, String> attributes) {
+  private String mungeAndExtractSchema(Map<String, String> attributes) {
```
johnrandolph, line 487: It isn't immediately obvious why we are crossing subFolder and subType values. 

Can we either make it more clear with comments, or, reference the doc? envelope.md doesn't explain it ... ?

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -379,7 +379,6 @@ private synchronized void cancelExecutor() {
     if (scheduledFuture != null) {
       try {
         scheduledFuture.cancel(false);
-        scheduledFuture.get();
```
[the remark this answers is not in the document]
grafnu: It actually causes problems -- once the Future is cancelled (previous line) the get() fails... Just never noticed it until I started to trace down errors...

validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
grafnu: Ok, you shamed me into fixing the underlying confusion, so the crossing the values is no longer required. It did involve some additional changes, but it's now a lot cleaner overall, I think.

grafnu:
I'm submitting this now even though there might be further tweaks necessary, since it at least cleans up a bunch of underlying issues that might be impacting other threads. If there's any additional comments/changes, just post them and I can address in another PR.

PR #288, opened 2022-04-06
Add site_defaults.json in site model directory, which can supply default values to all Metadata.

johnrandolph:


validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
```diff
@@ -655,6 +655,28 @@ private void loadSchema(String key) {
...
+  private void loadSiteDefaults() {
+    this.siteDefaults = null;
+
+    if (!schemas.containsKey(METADATA_JSON))
+      return;
+
+    File siteDefaultsFile = new File(siteDir, SITE_DEFAULTS_JSON);
+    try (InputStream targetStream = new FileInputStream(siteDefaultsFile)) {
+      schemas.get(METADATA_JSON).validate(OBJECT_MAPPER.readTree(targetStream));
+    } catch (ProcessingException | ValidationException e) {
```
grafnu, line 667: I'm typically lazy with my catch -- you can just catch (Exception e) and then have one line since there's no functional difference between them (just adds context and then walks it up the stack)

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
```diff
@@ -73,6 +69,7 @@
   private static final String SCHEMA_SUFFIX = ".json";
   private static final String REGISTRATION_SUMMARY_JSON = "registration_summary.json";
   private static final String SCHEMA_NAME = "UDMI";
+  private static final String SITE_DEFAULTS_JSON = "site_defaults.json";
```
grafnu, line 72: I was pondering if this might be better named "metadata_defaults.json" -- just because there might be other defaults that apply to the site that aren't metadata... I don't have a specific alternative, but trying to future-proof the name. (The sites part is also implied by it being at the top-level.) OTOH, site_defaults.json makes the name itself unique in the case when there are other kinds of default metadata (e.g. device type).

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -244,19 +263,41 @@ public void validateExpected() {
     exceptionMap.throwIfNotEmpty();
   }
 
+  private void deepMerge(Map<String, Object> map1, Map<String, Object> map2) {
```
grafnu, line 266: nit -- rename parameters to indicate which is target vs. src -- the IDE can be good about hinting based on names, so it does provide useful context for somebody using the function

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
johnrandolph: Ok but I got this from LocalDevice line 294. Should we clean that up too?

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
johnrandolph: I would agree I'm not a fan of the name now that we use the Metadata schema instead of a different, related one. It doesn't show the relationship. As long as the naming allows room for other defaults to named by a similar method I'm ok with it. metadata_defaults, then?

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
grafnu: There's a slight difference -- the code in LocalDevice.java is different in the two cases, where here it's the same in all cases. Doesn't really matter here (it's functionally correct) -- just something that jumped out at me.

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
grafnu: Another name to consider would be "site_metadata.json" which would (eventually then) be unique compared with a "hardware_metadata.json" (e.g. for all Delta O3 devices) or "entity_metadata.json" (that would be valid for, e.g. all "FCU"s in the system regardless of the hardware).  I'm thinking that's better than "metadata_defaults.json" which is still ambiguous (is it the default for the { site, entity, hardware }?)...  Personally, after this thinking I'd be leaning towards "site_metadata.json"... ?

johnrandolph: Ok so site_metadata?

grafnu:
SGTM

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
grafnu: SGTM

johnrandolph (commented):
Ok ptal. Note that calling it "site_metadata" also looks good in code w/r/t variable names.

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -180,19 +196,25 @@
   private final ExceptionMap exceptionMap;
   private final String generation;
   private final List<DeviceCredential> deviceCredentials = new ArrayList<>();
+  private final TreeMap<String, Object> siteMetadata;
```
grafnu, line 199: Seems like it would be cleaner overall to keep the class variable as Metadata (rather than TreeMap), and then convert to TreeMap only at the point of use?

johnrandolph: Then we convert it once per metadata loaded?

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -180,19 +196,25 @@
...
 
   private String deviceNumId;
 
   private CloudDeviceSettings settings;
 
   LocalDevice(
       File siteDir, File devicesDir, String deviceId, Map<String, JsonSchema> schemas,
-      String generation) {
+      String generation, Metadata siteMetadata) {
```
johnrandolph, line 207: Note the public interface loads it well structured, then it's converted for internal use

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
grafnu: True -- although I've never seen performance be an actual issue with any of this stuff :-). I also think it's only likely to be called relatively infrequently per device (once, maybe twice?)  Your call, I'm good either way.

grafnu (approved):

grafnu (commented):
Oh, "unapprove" -- was hoping to have the update for the udmi_site_model reference in the PR as well (for testing)...

PR #289, opened 2022-04-07
Mapping Examples

grafnu:


tests/config.tests/mapping.json
```diff
@@ -0,0 +1,29 @@
...
+      "FCU-123": {
+        "applied": "2018-08-27T22:39:12.364Z", // Last time a promotion was successfully applied externally
+        "requested": "2018-08-27T22:39:12.364Z", // Timestamp of requested model export
+        "status": {
+          "message": "Update in progress",
+          "category": "mapping.apply.start",
+          "timestamp": "2018-08-28T21:39:30.364Z",
+          "level": 500
+        }
+      }
```
charbull, line 26: The mapping must contain the following:

Which building is it for?
the cloud device id for the device.
Any dependencies such as links to another entity or connection to another entity must be present as well.


```
6798-guid:
  code: CHWS_WDT-17
  type: HVAC/CHWS_WDT
  cloud_device_id: 123456
  guid: CHWS_WDT-17-guid
  translation:
    return_water_temperature_sensor:
      present_value: "points.return_water_temperature_sensor.present_value"
      units:
        key: "pointset.return_water_temperature_sensor.units"
        values:
          degrees_celsius: "degC"
    supply_water_temperature_sensor:
      present_value: "points.supply_water_temperature_sensor.present_value"
      units:
        key: "pointset.points.supply_water_temperature_sensor.units"
        values:
          degrees_celsius: "degC"

1234-guid:
  type: FACILITIES/BUILDING
  code: UK-LON-S2
  ```

**Review by charbull — CHANGES_REQUESTED** [review]

File: tests/config.tests/mapping.json (hunk 1 above)
Comment (grafnu) [reply tests/config.tests/mapping.json:26]: The actually mapping info is in the other file/message -- that's in tests/event_mapping.tests/building_config.json  -- I'm unopinionated about the semantics contents of that message, so I could just take the snippet you had here and put it there. I don't understand the structure of the message, but that doesn't really matter.

File: tests/config.tests/mapping.json [hunk 2]
```diff
@@ -0,0 +1,30 @@
...
+//    applied < promoted: This means a entry has been promoted but has not yet been applied by the external system
+//    requested > exported: This means that an export has been requested but not yet triggered (mapper should export data)
+//
+{
+  "version": "1.3.14",
+  "timestamp": "2018-08-28T21:39:29.364Z",
+  "mapping_format": "building_config",
+  "mapping": {
+    "devices": {
+      "FCU-123": {
```
charbull, line 18: uuid4 guid here as well (after reading reply by email :)

tests/config.tests/mapping.json
charbull: yes my bad, it is the mapping info

charbull (changes requested):

bsimmons-onboard (commented):
@grafnu here is an example of the prediction contents.
[staged_equip_response.txt](https://github.com/faucetsdn/udmi/files/8466561/staged_equip_response.txt)

grafnu:
Thanks -- are there any significance to some of the property names like
"model.equip_type_predictions" -- specifically the "." in there?  This
typically causes lots of problems with various bits of code that like to
treat property names as variable names, etc...  Also, you have others that
are like "model.equip_name" -- and so there's an implicit hierarchy there
(rather than explicit in JSON). This would be better served as:

"model": {
  "equp_type_predictions": {
  },
  "eqip_name": {
  }
}

Which is semantically the same, but actually structured in JSON.  There's
some other restructuring I'd like to propose, but I wanted to understand
the semantics first.  The actual values you have in place all seem good --
thanks!

bsimmons-onboard (commented):
@grafnu good point about the implicit vs explicit hierarchy for the "model." fields. These were to distinguish between the individual predictions and the final one that gets autofilled. Here's a new version with an explicit model hierarchy. 
[staged_equip_response.txt](https://github.com/faucetsdn/udmi/files/8474287/staged_equip_response.txt)

grafnu:
I don't think this is valid JSON -- I assume you mean it's just a
string property?

                    "equip_name":
                    {
                      "VAV-1" // predicted equipment instance ID
                    },

bsimmons-onboard:
actually these fields are also probabilistic, added more examples with probabilities. 
[staged_equip_response.txt](https://github.com/faucetsdn/udmi/files/8474474/staged_equip_response.txt)

PR #293, opened 2022-04-12
Send out model and config messages through registrar

grafnu:
Initial use is for testing front-end work -- to make sure they can pull messages to properly set-up their internal DB.


validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
```diff
@@ -434,31 +434,47 @@ private Device fetchDevice(String localName) {
...
   }
 
-  private void sendConfigMessage(LocalDevice localDevice, SubFolder subFolder, Object subConfig) {
+  private void sendUpdateMessage(LocalDevice localDevice, String subType, SubFolder subfolder, Object target) {
+    String fieldName = subfolder.toString().toLowerCase();
+    try {
+      Field declaredField = target.getClass().getDeclaredField(fieldName);
+      sendSubMessage(localDevice, subType, subfolder, declaredField.get(target));
+    } catch (Exception e) {
+      throw new RuntimeException(String.format("Getting field %s from target %s%n", fieldName, target.getClass().getSimpleName()));
```
johnrandolph, line 458: Is this %s%n correct / is it some magic with the getSimpleName()?

grafnu:
%n is the java String.format equivalent of \n --
https://www.geeksforgeeks.org/format-specifiers-in-java/#:~:text=It%20is%20simply%20an%20escape,directly%20into%20the%20format%20string.&text=%2F%2F%20the%20newline%20%25n%20format%20specifiers
.

Although in this context it's not actually needed since it's a message for
an exception, which would not normally be terminated (so I removed it and
just pushed an update).

johnrandolph (approved):

PR #296, opened 2022-04-14
Merge UDMIF into faucetsdn UDMI

slevertbiot:
This is an initial merge of the UDMIF projects which includes an Angular front end and a GraphQL API service.  It also includes a preliminary IaC implementation.  All of the work is in progress.


slevertbiot:
I spoke to the IaC team.  They're reviewing their work and should get to updating this next week.

grafnu:
There's still a cloud/gcp/index.js and cloud/gcp/src/index.js file... are they both right? Also cloud/gcp/package.json is an exact copy of cloud/gcp/src/package.json... ?

grafnu (commented):
The index.js and package.json files under cloud/gcp/src and cloud/gcp seem redundant... are both sets supposed to be there?

slevertbiot:
For now, I'm acting as the middle man of this PR. I've let the team know about your latest comments.

github-biot:
> I think there's one stray .zip file cloud/gcp/function.zip? Or, if it's intended, what is that?
> 
> cloud/gcp/index.js is empty -- seems like it only half got moved to cloud/gcp/src/ --?

Those were put there temporary to be able to deploy the cloud function. This will be dynamic in the next PR. We will delete that src folder.

grafnu (approved):

PR #300, opened 2022-04-19
Fix schema for min_loglevel, etc

johnrandolph:


validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -571,9 +571,6 @@ PointPointsetConfig configFromMetadata(PointPointsetMetadata metadata) {
     if (Boolean.TRUE.equals(metadata.writable)) {
       pointConfig.set_value = metadata.baseline_value;
     }
-    pointConfig.min_loglevel = metadata.min_loglevel;
```
grafnu, line 574: Shouldn't there be new stuff that copies the values into the pointset config?   Somewhere in getDevicePointsetConfig()?

johnrandolph: My understanding is that this is an unused artifact of my earlier attempt to provide defaults by setting them directly.

With the schema changes

PointsetMetadata (not PointsetPointMetadata) holds sample_{limit,rate}_sec
SystemConfig and SystemMetadata holds min_loglevel

grafnu (approved):

PR #301, opened 2022-04-19
Shell change to be more explicit about shell numeric comparisons

johnrandolph:
Related to running these scripts on OSX and wc -w had different stdout

### DIFF
--- a/bin/loop_sequences
+++ b/bin/loop_sequences
@@ -39,7 +39,7 @@ if [[ ! -d $site_model ]]; then
 fi
 
 awc=$(echo $project_id $site_model $device_id $serial_no | wc -w)
-if [[ $awc != 4 ]]; then
+if [[ $awc -ne 4 ]]; then
     echo Missing configuration paramaters.
     exit -1
 fi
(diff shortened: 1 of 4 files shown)


grafnu (commented):
What's the reason for the change?  I don't have a problem with it... but I would have assumed that == was better than -eq in this case?  Also, seems to be a lot of stray files from the other PR in here...

johnrandolph:
-eq -ne are numeric comparison operators. This fixes the OSX    wc -w  output is "         4 "  we saw the other day.

grafnu:
Huh -- tried to read about and the linux man page (for bash) implies that
== is also arithmetic. But, it doesn't matter -- as long as it starts
working properly.  Huh, I think the difference (between linux and mac)
actually comes down to how the $( ) parts are handled... i can't quite
figure it out (because I don't have a mac), but some preliminary
experiments did not work as expected on my Linux setup. (Still doesn't
change the fact that using -eq fixes the problem in all cases.)

grafnu (approved):

PR #323, opened 2022-05-09
Change examples according to latest schema version 

noursaidi:
- Adds tests to cover validator message upgrader (new version but old schema and vice versa)
- Upgrade existing examples to match latest schema (string `version`, `hardware`/`software` instead of `make_model`/`firmware`


grafnu (approved):
Remember that we don't need to change all the files (or likely shouldn't change them all) since if we want to test backwards compatibility we'll need those files around. Not really a concern with this PR, just FYI in case you were on the warpath to change all the things.

PR #338, opened 2022-05-24
Add additional pubber opts

noursaidi:
- Adds option for an extra point 
- Adds ability to run additional opts in pubber cli - `extra_field` and `extra_point`, e.g.
`bin/pubber sites/udmi_site_model project_id AHU-1 123 extra_point=blahblah`
`bin/pubber sites/udmi_site_model project_id AHU-1 123 extra_point extra_field`


bin/pubber
```diff
@@ -2,8 +2,8 @@
 
 ROOT_DIR=$(realpath $(dirname $0)/..)
 
-if [ $# != 4 ]; then
-    echo $0 SITE_PATH PROJECT_ID DEVICE_ID SERIAL_NO
+if [[ "${#}" -lt "4" ]]; then
```
grafnu, line 5: Why this?  $# is a lot easier to read!  (Also, quotes are unnecessary with [[, since it's an explicit different parser to avoid the extra clutter!)

bin/pubber
```diff
@@ -23,4 +23,35 @@ $ROOT_DIR/pubber/bin/build > /dev/null
 
 echo Running tools version `(cd $ROOT_DIR; git describe)`
 
-$ROOT_DIR/pubber/bin/run $project_id $site_path $device_id $serial_no
+if [[ "${#}" -gt "0" ]]; then
```
grafnu, line 26: I would flip the if condition so the short branch is first, rather than after the else (just makes it easier to read)

bin/pubber
```diff
@@ -23,4 +23,35 @@ $ROOT_DIR/pubber/bin/build > /dev/null
...
+        if [[ $option == *"="* ]]; then
+            k=$(echo $option | cut -d'=' -f1)
+            v=$(echo $option | cut -d'=' -f2)
+            options[$k]=$v
+        else
+            options[$option]=$option
+        fi
+    done
+
+    extra_field="${options[extra_field]}"
```
grafnu, line 38: where is this used? I would expect somewhere in this PR to see how the added functionality is used...

noursaidi: `extraField` is an existing pubber config, `extra_field` provides access to it. I'll update the documentation to describe the additional functionality

bin/pubber
noursaidi: Changed to use `((` instead as per [style guide](https://google.github.io/styleguide/shellguide.html)

noursaidi:
@grafnu I had to make a change because if extraField is in the pubber config then it's included by pubber, so I made a change to only include extraField if it's set.

grafnu (approved):

PR #340, opened 2022-05-25
add no_hardware to pubber options

noursaidi:
Add pubber runtime option to omit `hardware` field to trigger validation error (missing required field)


noursaidi:
@grafnu PTAL -  the options are now an object in the configuration file

bin/pubber
```diff
@@ -31,29 +31,24 @@ else
         if [[ $option == *"="* ]]; then
             k=$(echo $option | cut -d'=' -f1)
             v=$(echo $option | cut -d'=' -f2)
-            options[$k]=$v
         else
-            options[$option]=$option
+            k=$option
+            v=$option
```
grafnu, line 36: I'm wondering if this should be a boolean.  Ideally we can be more semantic about this, and having it as a boolean would allow us to make sure people are using options correctly...

pubber/src/main/java/daq/pubber/Configuration.java
```diff
@@ -15,8 +17,7 @@ public class Configuration {
   public String keyFile = "local/rsa_private.pkcs8";
   public byte[] keyBytes;
   public String algorithm = "RS256";
-  public Object extraField;
   public String serialNo;
   public String macAddr;
-  public String extraPoint;
+  public HashMap<String, String> options = new HashMap<>();
```
grafnu, line 22: I would make this an explicit Options object. The reason is once you go to a general map here you lose the ability to have any checking on the contents and types.  Ideally, this structure would be specified as a protobuf or json schema itself, so we get the auto-documenting process with that.  It also makes the code simpler since it's not a lot of key-lookups-in-maps and makes the code itself more stable through any changes.

bin/pubber
noursaidi: Agree, turned it into a boolean with correct type assignment for checking

pubber/src/main/java/daq/pubber/Configuration.java
noursaidi: Turned options into an options object

noursaidi:
@grafnu PTAL if the changes now address the comments

bin/pubber
```diff
@@ -31,29 +31,26 @@ else
         if [[ $option == *"="* ]]; then
             k=$(echo $option | cut -d'=' -f1)
             v=$(echo $option | cut -d'=' -f2)
-            options[$k]=$v
+            if ! [[ $v =~ ^[0-9]+([.][0-9]+)?$ ]]; then
```
grafnu, line 34: I wouldn't go about it this way, since although it seems "simple" it's actually somewhat of a minefield because, say, what if you want a string that looks like a number?  Better to be just simpler (all options are strings) and then deal with the conversion later, if/when necessary.

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -357,24 +362,32 @@ private void initializeDevice() {
...
-    deviceState.system.software.put("firmware", "v1");
+    deviceState.system.software.put("firmware", "v1");    
     deviceState.system.last_config = new Date(0);
-    devicePoints.extraField = configuration.extraField;
 
-    if (configuration.extraPoint != null && !configuration.extraPoint.isEmpty()) {
-      addPoint(makePoint(configuration.extraPoint,
+    // Pubber runtime options
+    if (configuration.options.extraField) {
+      devicePoints.extraField = "extra_field";
```
grafnu, line 379: see, now if this were still a string then you have your extra field value!

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -357,24 +362,32 @@ private void initializeDevice() {
...
+    }
+
+    if (configuration.options.extraPoint) {
+      addPoint(makePoint("extra_point",
```
grafnu, line 383: ditto -- extra_point here would be useful

pubber/src/main/java/daq/pubber/Pubber.java
grafnu, line 379: see, now if this were still a string then you have your extra field value!

bin/pubber
```diff
@@ -31,29 +31,26 @@ else
...
+                v="\"$v\""
+            fi
         else
-            options[$option]=$option
+            k=$option
+            v=true
```
grafnu, line 39: strings or boolean -- this case is fine, I think -- no magic here

bin/pubber
```diff
@@ -31,29 +31,26 @@ else
...
+        printf -v options_json '%s"%s":%s,' "$options_json" "$k" "$v"
     done
-
-    extra_point="${options[extra_point]}"    
-    
-    if [ -n "${options[extra_field]}" ]; then
-        printf -v extra_field '"extraField": "%s",' "${options[extra_field]}"
-    fi
-    
+    options_json="{${options_json%?}}"
```
grafnu, line 43: what's the ? for?  If it's a wildcard, can you be explicit about what it's removing (trailing comma, right?)

noursaidi: last character, changed to remove trailing comma

pubber/src/main/java/daq/pubber/Pubber.java
noursaidi: extra field value now taken from string

pubber/src/main/java/daq/pubber/Pubber.java
noursaidi: extra point value now taken from string

bin/pubber
noursaidi: Removed so options are either boolean true or string value

proxy/pubber.json
```diff
@@ -2,11 +2,13 @@
...
-  "extraField": 1234,
   "serialNo": "test_aux-16132",
   "macAddr": "3c5ab41e8f0a",
   "keyFile": "./udmi_site_model/devices/GAT-123/ec_private.pkcs8",
   "algorithm": "ES256",
   "gatewayId": "GAT-123",
-  "deviceId": "SNS-4"
+  "deviceId": "SNS-4",
+  "options": {
+    "extraField": true
```
grafnu, line 12: shouldn't extraField be a string, not boolean?

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -369,12 +371,21 @@ private void initializeDevice() {
     deviceState.system.software = new HashMap<>();
     deviceState.system.software.put("firmware", "v1");
     deviceState.system.last_config = new Date(0);
-    devicePoints.extraField = configuration.extraField;
 
-    if (configuration.extraPoint != null && !configuration.extraPoint.isEmpty()) {
-      addPoint(makePoint(configuration.extraPoint,
+    // Pubber runtime options
+    if (!configuration.options.extraField.isEmpty()) {
```
grafnu, line 376: should be "extraField != null" -- if somebody really wants to specify an empty extra field then they should be able to.

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -369,12 +371,21 @@ private void initializeDevice() {
...
+      devicePoints.extraField = configuration.options.extraField;
+    }
+
+    if (!configuration.options.extraPoint.isEmpty()) {
```
grafnu, line 380: semantically this should be != null, and then something else would throw an error saying "can't define an empty point name" (or just ignore that case because it's all for error checking anyway)

pubber/src/main/java/daq/pubber/ConfigurationOptions.java
```diff
@@ -0,0 +1,11 @@
+package daq.pubber;
+
+/**
+ * Pubber configuration options which change default behavior.
+ */
+public class ConfigurationOptions {
+  public Boolean noHardware = false;
+  public String extraPoint = "";
```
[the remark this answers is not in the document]
noursaidi: @grafnu thanks! should that extend to noHardware too which I've assigned a default value? Then the check is `!= null && !configuration.optionsnoHardware`?

grafnu:
Correct... generally avoid a default unless it's really really
useful/necessary.

grafnu (approved):
Brilliant. Thanks for working through all the iterations with me!

PR #347, opened 2022-06-01
Script to delete bad tags from the repo

grafnu:


bin/clean_tags
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -e
+
+repo=$1
+shift || true
+
+if [[ -n $repo ]]; then
+    git fetch $repo --tags
+fi
+
+for tag in `git tag`; do
```
johnrandolph, line 10: New, better style is $()  not ``   (backtics)

johnrandolph (approved):

PR #357, opened 2022-06-07
pass arguments to gcloud deploy

noursaidi:
I needed this so I can deploy the gcloud functions with the non-default service account (the default service account has editor role over a project by default).

Just passes the arguments onto the `gcloud deploy` call, so [gcloud deploy parameters](https://cloud.google.com/sdk/gcloud/reference/functions/deploy) can be used. I only needed `service-account`, but figured power users may want to modify some of the other parameters too

### DIFF
--- a/dashboard/deploy_dashboard_gcloud
+++ b/dashboard/deploy_dashboard_gcloud
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 
-if [ $# != 1 ]; then
-    echo Usage: $0 [project_id]
+if (( $# < 1 )); then
+    echo Usage: $0 PROJECT_ID [options]
     false
 fi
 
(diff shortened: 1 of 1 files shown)


dashboard/deploy_dashboard_gcloud
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 
-if [ $# != 1 ]; then
-    echo Usage: $0 [project_id]
+if (( $# < 1 )); then
+    echo Usage: $0 PROJECT_ID [options]
```
grafnu, line 4: What options are potentially interesting here?  I've never had to add anything else, so just wondering... I'm a bit worried about making this too flexible since ideally the cloud-deploy part would be very regular/consistent...

noursaidi: I looked into it, I don't think any. I made some assumptions based on the names of some things but I don't think they're relevant anymore, so it's only the service account then.

grafnu:
I missed something I think -- what about the service account?  Really
trying to deprecate those so I'm not sure in this case why they'd be
necessary?

noursaidi:
It's the runtime service account (https://cloud.google.com/functions/docs/securing/function-identity#runtime_service_account) for the cloud function. An instance of a cloud function inherits its permissions from either the defined service account or the app engine default account if unspecified.

So by specifying the service account the cloud functions are associated with, their access to the GCP project can be controlled, and reduced from the everything access they have by default at the moment.

grafnu (approved):

grafnu:
Sure, but when does this come up?  I'm basically wondering if this should
be a required thing or is optional really ok. Practically speaking this
script was more meant for a quickstart, but I was expecting the "real"
deployments to go through terraform.

Anyway -- nothing specifically wrong with this change, more just
questioning the overall workflow!

PR #358, opened 2022-06-08
Add sequence test to make sure config messages are properly acked by a client.

grafnu:


noursaidi:
I tested with pubber option from #359 and it seems works, only comment is I think the config ack test should be separate from the last_update test

johnrandolph (approved):
Overall fine just comment on the variable renames--I'm an APPROVE if you are good with my suggestion.

grafnu:
Fair enough -- I was doing that just to avoid the line-length warning, but
not really an issue restoring and keeping it on two lines. PTAL.

Issue #334, opened 2022-05-13
Registrar localnet.subsystem no longer accepted 


noursaidi:
the `subsystem` field has been replaced with `families`, however existing devices still use `subsystem`.

The registrar cannot be used with these devices because it disregards the `subsystem` block

PR #362, opened 2022-06-15
Downgrade config generation

grafnu:
Addresses https://github.com/faucetsdn/udmi/issues/334 and b/234435528

Main contribution is a new MessageDowngrader class that is used to downgrade a configuration message when it's applied for a device:  validator/src/main/java/com/google/daq/mqtt/registrar/MessageDowngrader.java

tests/downgrade.site is a small test site_model (rather than using the external udmi_site_model) that is used for an integration test on this (runs bin/registrar and compares the output against redacted expected files).

There's also a unit tests for the MessageDowngrader class (and supporting .json files for consumption)

Everything else is all the other bits to make it fit together, or testing infrastructure.


johnrandolph (approved):

grafnu:
There's not one (junit convention) that I'm aware of... Usually with this
sort of thing I try to go with "more interesting stuff at the top" and so
tend to shove helper methods to the bottom.  Staticy stuff (so class or
initialization) things tend to go up top. If you think of "helper methods"
as more-privaty, then I would say the rough order is *static -> public ->
private,* but that's really just something I made up in my head right now
(which is essentially how this method is organized).  The other
(conflicting) bit is that helper methods should go "closest" to where they
are most commonly used.... I'm sure that applies sometimes too!  (Either
way, open to suggestions.)

PR #365, opened 2022-06-17
Onboarding discovery mapping flow

grafnu:


grafnu:
Main thing to look at would be the docs -- https://grafnu.github.io/udmi/docs/specs/onboarding.html -- everything else is either a massive refactoring or other noise...

PR #370, opened 2022-06-21
add a discovery event example from a bacnet scan

gnmerritt:
We'd like to ensure that the discovery events our scanner is emitting our valid UDMI, so we're checking an example into the tests directory. This example is constructed from a live building controller and includes some of its quirks and field naming conventions


noursaidi:
Hi Nathan,

You can run these tests locally on your machine by using `bin/test_schema`, and then looking at the `.out` file which in this example would be `tests/event_discovery.tests/from_bacnet.out`. If the file is blank then there are no errors

gnmerritt:
Yep, ran locally before putting the PR up. 

Also maybe I'm missing something but it looks to me like the test runner actually compares the `.out` file with the actual output to make sure they match (rather than writing to it). So that's why e.g. https://github.com/faucetsdn/udmi/blob/master/tests/event_discovery.tests/errors.out is checked in, to make sure that the validator produces those same messages. Is that right?

grafnu:
Nathan, yes -- that is correct.  Some .out files are _not_ empty because they are specifically testing that errors are handled correctly -- but in your case it's good/appropriate that it's empty!

Note that given some of the other changes I'm proposing this will likely need to be adjusted, but that'll come out when I put that PR together.

grafnu (approved):

PR #374, opened 2022-06-22
Update discovery system fields

grafnu:


gnmerritt:
This makes sense and LGTM. Thank you Trevor

PR #376, opened 2022-06-27
Redirect

grafnu:
Initial version of Pubber that accepts a registryRedirect option


pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -547,6 +549,16 @@ private void initialize() {
       throw new RuntimeException("While creating out dir " + outDir.getPath(), e);
     }
 
+    initializeMqtt();
+  }
+
+  private void disconnectMqtt() {
+    Preconditions.checkState(mqttPublisher != null, "mqttPublisher not defined");
+    mqttPublisher.close();
```
johnrandolph, line 557: Does this handle any kind of completion of sending messages (whether that's desired or not), or just drop the connection?

johnrandolph (approved):

grafnu:
I'm not sure... I would assume it waits for everything in the outgoing
queue to be sent. There's also things we'll want to do with sending a UDMI
status update that says "we're about to redirect" and we'll need to make
sure that gets through, etc...   There's a lot of error handling in general
I didn't deal with here -- so I'll need to go through and clean that stuff
up later for sure!

PR #379, opened 2022-06-28
Blob config - add "applied" phase

johnrandolph:


tests/config.tests/prep.json
```diff
@@ -0,0 +1,25 @@
...
+    "blobs": {
+      "fw_payload": {
+        "phase": "prep",
+        "type": "base64",
+        "content_type": "application/zip",
+        "payload": "UEsDBAoAAAAAAAZZ3VSU7ZiuBQAAAAUAAAAIABwAYm9vdC5pbWdVVAkAA9xqvGLcarxidXgLAAEEFjQAAARTXwEAYm9vdApQSwMECgAAAAAAB1ndVK9osDMDAAAAAwAAAAYAHABmdy5pbWdVVAkAA95qvGLearxidXgLAAEEFjQAAARTXwEAZncKUEsBAh4DCgAAAAAABlndVJTtmK4FAAAABQAAAAgAGAAAAAAAAQAAAKSBAAAAAGJvb3QuaW1nVVQFAAPcarxidXgLAAEEFjQAAARTXwEAUEsBAh4DCgAAAAAAB1ndVK9osDMDAAAAAwAAAAYAGAAAAAAAAQAAAKSBRwAAAGZ3LmltZ1VUBQAD3mq8YnV4CwABBBY0AAAEU18BAFBLBQYAAAAAAgACAJoAAACKAAAAAAA="
+      },
+      "update_payload": {
+        "phase": "prep",
+        "type": "url",
```
grafnu, line 13: Looking at this, are the type and href/payload field redundant?  That is, could it be specified as either:
1) _type_ and _src_ fields -- where the _type_ says what to do with the _src_?  So if _type_ is "base64" then it's inline, if it's "url" then URL, etc...
2) One of { _base64_, _href_ } with no _type_ -- so if _base64_ is defined then it's that, if it's _href_ then URL, etc... It would be an error to have multiple fields defined.

tests/config.tests/apply.json
```diff
@@ -2,11 +2,10 @@
   "version": 1,
   "blobset": {
     "blobs": {
-      "codebase": {
-        "stage": "apply",
-        "target": "e21637e422d3153b866279af92ec1b1fd17c8846"
+      "hello_payload": {
+        "phase": "apply"
```
grafnu, line 6: I think the **apply** (and others) _phase_ might also have the _type/src/href/whatever_ fields defined as well, and if so, then it means "download and apply" -- sometimes you don't really care and there's no need to then implicitly walk through all the states.

tests/config.tests/prep.json
johnrandolph: Unless if it were expensive to have multiple fields, I think using explicitly named ones is better. Also It's much more straightforward to validate a field named url in one pass, instead of branching based on type etc. 

Alternatively I think you are saying it needs to be an error for type=url, href="...defined" and also payload!=null, but similarly it needs to be an error if type=url and payload!=Url_format, if we are doing error checking.

grafnu:
Multiple fields is relatively easy -- the only real expense is for
developer's time in terms of writing the JSON schema part :-).

If we go for the "no type" model and have the explicitly named ones then
I'm saying the error would be violation of a "one of" constraint (which is
possible to express in JSON schema), which is very easy to check/enforce.
Semantic error checking (URL formatting) is still there either way so don't
change much of that.

The only thing I'm twitchy about is the redundancy of having *type* and
*multiple-fields*, since then there's more things to go wrong (the type
doesn't match the field used) that doesn't add any useful information into
the system. Everything else I think its lost in the noise and/or has to be
done anyway.

grafnu (commented):
prep.json looks good to me!  No problems there. Technically, it's not necessary to have the other test .json files -- if the difference is just the value of "phase" it doesn't really matter. It would only really be useful if we needed them to link for examples. One way to approach this is to just remove the other ones, and then move on to updating the schema .json to support prep.json... And/or to write a documentation.md file somewhere to describe the overall flow state/machine...

johnrandolph:
> prep.json looks good to me! No problems there. Technically, it's not necessary to have the other test .json files -- if the difference is just the value of "phase" it doesn't really matter. It would only really be useful if we needed them to link for examples. One way to approach this is to just remove the other ones, and then move on to updating the schema .json to support prep.json... And/or to write a documentation.md file somewhere to describe the overall flow state/machine...

I'm a fan of examples so I added ones for apply and commit. Next I will update the schema json

tests/config.tests/apply.json
johnrandolph: I think that we need a specific phase that allows one-shot mode, otherwise two problems can result:

1. In the prep (download) phase the device pulls content and saves it. If the urls etc change between prep and apply, either the device has to keep track of the former prep blob to see the difference, or it has to needlessly do the download again.

johnrandolph: 2. It is not clear whether all devices will want to do the 3 step or 1 step if the lines are blurred between how they work

johnrandolph (commented):
Request comment

tests/state.tests/blobset_receive.json
```diff
@@ -0,0 +1,44 @@
...
+    "serial_no": "000000",
+    "last_config": "2022-07-11T00:00:10.000Z",
+    "operational": true
+  },
+  "blobset": {
+    "blobs": {
+      "arbitrary_manufacturer_id": {
+        "status": {
+          "message": "Received arbitrary_manufacturer_id",
+          "category": "blobset.blobs.receive",
```
grafnu, line 21: for pointset, we ended up with "pointset.point.receive" (singular 'point') -- which is slightly different than "blobset.blobs.receive" -- (plural 'blobs') which is more aligned around the JSON object tree. They're both valid perspectives, but in the face of a strong preference I'd go with the singular version because 1) consistency with pointset, and 2) it makes a bit more sense semantically on the receiving end (as it is scoped to just one blob).

grafnu (approved):

PR #387, opened 2022-07-07
change cloud function targets to constants

noursaidi:
Allows cloud functions and trigger topics to be renamed

Deploy instructions:

`dashboard/deploy_dashboard_gcloud daq1-273309 --udmi_state=state4 --udmi_target=target4 --udmi_config=config4 --udmi_reflect=reflect4`

### DIFF
--- a/dashboard/deploy_dashboard_gcloud
+++ b/dashboard/deploy_dashboard_gcloud
@@ -11,6 +11,30 @@ cd $ROOT/dashboard
 PROJECT=$1
 shift
 
+udmi_state=udmi_state
+udmi_reflect=udmi_reflect
+udmi_target=udmi_target
+udmi_config=udmi_config
+
+append_env_var () {
+   extra_args="$extra_args --set-env-vars $1=$2"
+} 
+
+while (( $# > 0 )) ; do
+    k=$(echo $1 | cut -d'=' -f1)
+    v=$(echo $1 | cut -d'=' -f2)
+
+  case $k in
+    --udmi_state) udmi_state=$v && append_env_var UDMI_STATE $v;;
+    --udmi_reflect) udmi_reflect=$v && append_env_var UDMI_REFLECT $v;;
+    --udmi_target) udmi_target=$v && append_env_var UDMI_TARGET $v;;
+    --udmi_config) udmi_config=$v && append_env_var UDMI_CONFIG $v;;
+    --reflect_registry) append_env_var REFLECT_REGISTRY $v;;
+    --service_account) extra_args="$extra_args --service_account=$v"
+  esac
+  shift
+done
+
 RUNTIME=nodejs12
 SOURCE=functions/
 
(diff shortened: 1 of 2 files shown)


dashboard/deploy_dashboard_gcloud
```diff
@@ -11,6 +11,30 @@ cd $ROOT/dashboard
 PROJECT=$1
 shift
 
+udmi_state=udmi_state
```
grafnu, line 14: what are these variables used for?

dashboard/deploy_dashboard_gcloud
```diff
@@ -11,6 +11,30 @@ cd $ROOT/dashboard
...
+append_env_var () {
+   extra_args="$extra_args --set-env-vars $1=$2"
+} 
+
+while (( $# > 0 )) ; do
+    k=$(echo $1 | cut -d'=' -f1)
+    v=$(echo $1 | cut -d'=' -f2)
+
+  case $k in
+    --udmi_state) udmi_state=$v && append_env_var UDMI_STATE $v;;
```
grafnu, line 28: I would keep these clauses as simple as possible, e.g., either have "append_env_var" or "udmi_state =" but not both. If the later, then have code elsewhere that converts them to the necessary ENV variable (but as per the other comment I don't understand what the udmi_state variable is for anyway).

dashboard/deploy_dashboard_gcloud
```diff
@@ -26,8 +50,16 @@ EOF
 
 PUBSUB_FUNCTIONS="udmi_target udmi_state udmi_config udmi_reflect"
 for func in $PUBSUB_FUNCTIONS; do
-    echo Deploying pubsub-trigger function $func...
-    gcloud functions deploy $func --set-env-vars GCP_PROJECT=$PROJECT --trigger-topic=$func --runtime=$RUNTIME --project=$PROJECT --source=$SOURCE "$@" &
+    echo Deploying pubsub-trigger function $func as ${!func}...
```
grafnu, line 53: whyfor the ${!}?

dashboard/functions/index.js
```diff
@@ -10,14 +10,19 @@ if (!process.env.GCLOUD_PROJECT) {
   process.env.GCLOUD_PROJECT = PROJECT_ID;
 }
 
+const UDMI_TARGET = process.env.UDMI_TARGET || 'udmi_target';
```
grafnu, line 13: I'm a bit concerned that we're ending up with two different "configuration" mechanisms here... one in the deploy_versions.js file, and the other through ENV variables. I'd be much happier if we could consolidate on one or the other. (I'm less concerned about the GCLOUD_PROJECT one since that's really outside UDMI itself).

dashboard/functions/index.js
```diff
@@ -10,14 +10,19 @@ if (!process.env.GCLOUD_PROJECT) {
...
+const REFLECT_REGISTRY =  process.env.REFLECT_REGISTRY ||'UDMS-REFLECT';
+
 const version = require('./version');
 const functions = require('firebase-functions');
 const admin = require('firebase-admin');
 const { PubSub } = require(`@google-cloud/pubsub`);
 const pubsub = new PubSub();
 const iot = require('@google-cloud/iot');
 
-const REFLECT_REGISTRY = 'UDMS-REFLECT';
```
grafnu, line 20: As per the other issue, should the region for the registry be configurable as well?  It looks like it might be inappropriately assuming the wrong one (that is, the one of the site registry itself), so we might need to fix that... so maybe that is worth a separate PR...

dashboard/deploy_dashboard_gcloud
noursaidi: This sets defaults for the different functions

dashboard/deploy_dashboard_gcloud
noursaidi: The functions (entry points) in `index.js` dont change name, but the triggers and name of the instance of deployed cloud function do. So, for udmi_target, we need the function name `$func` and the trigger/instance name `$udmi_target`, so using `${!func}` to access `$udmi_target`, `$udmi_state` etc programmatically within the loop

PR #390, opened 2022-07-17
Windows filesystem fix

pisuke:
Avoiding the metadata.system.aux key to be named Aux in the Java class name and file name, because in Windows filesystems it is forbidden to name a file as AUX or any of its different letter case variations.

### DIFF
--- a/schema/model_system.json
+++ b/schema/model_system.json
@@ -76,6 +76,7 @@
       ]
     },
     "aux": {
+      "title": "auxiliary",
       "type": "object",
       "additionalProperties": false,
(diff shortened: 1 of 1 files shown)


schema/model_system.json
```diff
@@ -76,6 +76,7 @@
       ]
     },
     "aux": {
+      "title": "auxiliary",
```
grafnu, line 79: "Auxiliary" in caps (output is the same, but just makes it look more like a title)

grafnu (approved):
And obviously need to run gencode!

PR #392, opened 2022-07-19
Add message trace/playback capability (for testing)

grafnu:
Files in traces/ are captured messages that are used for test playback (so don't need to really be reviewed). Everything else is to support trace/playback/test capabilities.


grafnu:
Likely yes, but I was trying to keep things focused rather than do
everything... Right now, the attributes are implicitly defined by the
message filename (which includes the device ID and subfolder), and the time
comes from the message itself. The other parameters (projectId,
deviceNumId, etc...) -- are largely irrelevant.

johnrandolph (approved):

PR #397, opened 2022-07-26
Strengthen the schema w/r/t format, required fields

johnrandolph:


bin/check_version
```diff
@@ -15,10 +15,14 @@ if [[ -z $origin ]]; then
     origin=faucetsdn
 fi
 
-git fetch $origin --tags 2>/dev/null
+git fetch $origin --tags # 2>/dev/null
```
grafnu, line 18: unexpected #

schema/config_blobset_blob.json
```diff
@@ -11,13 +11,20 @@
...
     },
     "url": {
-      "type": "string"
+      "type": "string",
+      "format": "uri"
     },
     "sha256": {
-      "type": "string"
+      "type": "string",
+      "pattern": "^[0-9A-Za-z]{64}$"
```
grafnu, line 23: what do you think about just making it all-lower or all-upper? Reason being, then it's easier to do a naive "string compare" to see if "hash matched" -- otherwise it takes extra code that somebody is likely to get wrong...  I'd rather force the generation to be all-lower than rely on the equality check.

schema/config_blobset_blob.json
```diff
@@ -11,13 +11,20 @@
...
     }
-  }
+  },
+  "oneOf": [
+    { "required": ["content_type", "base64"] },
```
grafnu, line 27: I would have expected oneOf { base64, url } (transport mechanism), and then sha256 is always required, and then content_type is optional?  I suppose sha256 isn't required with base64... But doesn't content_type potentially apply to all transport mechanisms? Oh wait, is it because content_type is included in the HTTP response?

schema/config_blobset_blob.json
johnrandolph: I'm fine with all lowercase because that's customary. I was wondering if device makers would have other opinions, especially from non-Unix platforms.

johnrandolph: Also this is a mistake and it should be a-f not a-z.

schema/config_blobset_blob.json
johnrandolph: Yes, content_type comes from http so we don't want to allow overriding it--feels dangerous to me

grafnu (approved):

PR #400, opened 2022-07-29
Fix output generation for device validation reports

grafnu:
Things in validator/trace/simple/ can be ignored (trace input files)
Things in validator/trace/simple.out/ are examples of what the output produced is, so should be checked for completeness.
Other code bits deserve code-review as per normal.


validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
```diff
@@ -538,6 +534,11 @@ private void sendValidationResult(Map<String, String> origAttributes, Object mes
       ReportingDevice reportingDevice) {
     try {
       ValidationEvent validationEvent = makeValidationEvent();
+      validationEvent.status = reportingDevice.getErrorStatus();
```
johnrandolph, line 537: Should makeValidationEvent() help with some of the building below?

grafnu:
The function creates the basic common bits, but then it's filled in quite
differently for the different uses -- I'm all for generalization, but not
quite sure how to do it in this case!

validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
grafnu: The function creates the basic common bits, but then it's filled in quite
differently for the different uses -- I'm all for generalization, but not
quite sure how to do it in this case!

validator/src/main/java/com/google/daq/mqtt/validator/ReportingDevice.java
```diff
@@ -200,35 +208,57 @@ public Entry getErrorStatus() {
     if (errors.isEmpty()) {
       return null;
     }
+    return errors.size() == 1 ? makeEntry(errors.get(0)) : makeCompoundEntry(errors);
+  }
+
+  private Entry makeCompoundEntry(List<Exception> erx) {
```
johnrandolph, line 214: Could we have docstrings that give example input/output here? github doesn't lend itself to jumping around very quickly to piece it together.

validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
johnrandolph: ok

johnrandolph (approved):
I think at some point we need a util class that has all these general purpose functions like exception translators etc? But for now lgtm

PR #408, opened 2022-08-04
Add reflector utility

grafnu:
This adds a basic utility that can be used to inject messages into the system through the reflector. This is used by testing scripts to, for example, reset or modify a device config.

### DIFF
--- a/bin/reset_config
+++ b/bin/reset_config
@@ -1,4 +1,4 @@
-#!/bin/bash -ex
+#!/bin/bash -e
 
 ROOT=$(realpath $(dirname $0)/..)
 
--- a/bin/reset_config
+++ b/bin/reset_config
@@ -15,7 +15,7 @@ cd $ROOT
 
 device_config=/tmp/${device_id}_config.json
 cp $site_dir/devices/${device_id}/out/generated_config.json $device_config
-now_date=$(date -Ins -u | sed -E 's/.{6}\+.*/Z/' | tr , .)
+now_date=$(python3 -c 'import datetime; print(datetime.datetime.utcnow().isoformat() + "Z")')
 echo Setting config timestamp $now_date
 jq .timestamp=\"$now_date\" < $device_config | sponge $device_config
 
--- a/bin/reset_config
+++ b/bin/reset_config
@@ -17,6 +17,7 @@ device_config=/tmp/${device_id}_config.json
 cp $site_dir/devices/${device_id}/out/generated_config.json $device_config
 now_date=$(python3 -c 'import datetime; print(datetime.datetime.utcnow().isoformat() + "Z")')
 echo Setting config timestamp $now_date
+sponge < /dev/null || echo sponge command not found, might need workaround for osx or other
 jq .timestamp=\"$now_date\" < $device_config | sponge $device_config
 
 echo Resetting device $device_id config...
--- a/bin/reset_config
+++ b/bin/reset_config
@@ -11,14 +11,14 @@ site_dir=$(realpath $1)
 project_id=$2
 device_id=$3
 shift 3
-cd $ROOT
+cd ${ROOT}
 
 device_config=/tmp/${device_id}_config.json
-cp $site_dir/devices/${device_id}/out/generated_config.json $device_config
+cp ${site_dir}/devices/${device_id}/out/generated_config.json ${device_config}
 now_date=$(python3 -c 'import datetime; print(datetime.datetime.utcnow().isoformat() + "Z")')
-echo Setting config timestamp $now_date
+echo Setting config timestamp ${now_date}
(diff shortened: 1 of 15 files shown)


validator/bin/reflector
```diff
@@ -0,0 +1,23 @@
...
+device_id=$3
+shift 3
+cd $ROOT
+
+validator/bin/build
+
+jarfile=validator/build/libs/validator-1.0-SNAPSHOT-all.jar
+mainclass=com.google.daq.mqtt.validator.Reflector
+
+cmd="java -cp $jarfile $mainclass -p $project_id -s $site_dir -d $device_id $*"
```
johnrandolph, line 22: Why not just invoke directly and "$*" on end?

validator/src/main/java/com/google/daq/mqtt/validator/Reflector.java
```diff
@@ -0,0 +1,140 @@
...
+    try (FileInputStream fis = new FileInputStream(dataFile)) {
+      String data = new String(fis.readAllBytes());
+      reflect(topic, data);
+    } catch (Exception e) {
+      throw new RuntimeException("While processing input file " + dataFile.getAbsolutePath(), e);
+    }
+  }
+
+  private void reflect(String topic, String data) {
+    client.publish(deviceId, topic, data);
```
johnrandolph, line 88: Is this going to verify that the json input is well-formed?

bin/reset_config
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -e
+
+ROOT=$(realpath $(dirname $0)/..)
+
+if [[ $# != 3 ]]; then
+    echo Usage: $0 site_dir project_id device_id
+    false
```
johnrandolph, line 7: Do you want exec false? or just exit 1?

bin/reset_config
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+site_dir=$(realpath $1)
+project_id=$2
+device_id=$3
+shift 3
+cd $ROOT
+
+device_config=/tmp/${device_id}_config.json
+cp $site_dir/devices/${device_id}/out/generated_config.json $device_config
```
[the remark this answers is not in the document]
grafnu: done

bin/reset_config
```diff
@@ -0,0 +1,23 @@
...
+now_date=$(date -Ins -u | sed -E 's/.{6}\+.*/Z/' | tr , .)
```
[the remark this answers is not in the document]
grafnu: done

bin/reset_config
```diff
@@ -0,0 +1,23 @@
...
+echo Setting config timestamp $now_date
+jq .timestamp=\"$now_date\" < $device_config | sponge $device_config
```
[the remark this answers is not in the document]
johnrandolph: Update: Just comment sponge is a Linuxism or whatever

validator/bin/reflector
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -e
+
+ROOT=$(realpath $(dirname $0)/../..)
+
+if [[ $# < 4 ]]; then
+    echo Usage: $0 site_dir project_id device_id directive [directives...]
+    echo "  Directive is something like update/config:sites/udmi_site_model/devices/AHU-1/out/generated_config.json"
+    false
```
johnrandolph, line 8: exit 1?

bin/reset_config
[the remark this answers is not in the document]
grafnu: added runtime check and warning message, would have to fix actual OSX compat in another round (with QA)

validator/bin/reflector
grafnu: there was supposed to be another diagnostic "echo" in there... added!

validator/src/main/java/com/google/daq/mqtt/validator/Reflector.java
grafnu: No -- it's not actually even strictly JSON. This was meant as a reasonably low-level utility that wasn't really opinionated about formats. Maybe future iterations could add a "validate schema" flag or something...  E.g. this tool can be used to validate that a borked garbage config does not tank the system.

bin/reset_config
grafnu: I avoid "exit 1" because sometimes I'll end up doing "source" on a script and then "exit 1" will totally bork the shell session... the behavior of "false" is only that if the -e flag is specified.  So, in this particular case they are functionally equivalent, but I'm not sure if there's really a compelling reason for one over the other in the big picture.

validator/bin/reflector
grafnu: same as other comment... not sure if there's a clear reason for one over the other because they're just "different" in how they handle different edge use cases...

johnrandolph (approved):

PR #409, opened 2022-08-04
Redirect

grafnu:


pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -73,6 +74,8 @@ public class Pubber {
 
   public static final int SCAN_DURATION_SEC = 10;
   public static final String DISCOVERY_ID = "RANDOM_ID";
+  public static final String IOT_ENDPOINT_BLOB_NAME = "_iot_config";
+  public static final String BLOB_COMMIT_PHASE = "commit";
```
johnrandolph, line 78: "applied"?  We didn't define a commit phase, after some discussion about 3 phases or 1

Also we need these as const somewhere...

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -73,6 +74,8 @@ public class Pubber {
 
   public static final int SCAN_DURATION_SEC = 10;
   public static final String DISCOVERY_ID = "RANDOM_ID";
+  public static final String IOT_ENDPOINT_BLOB_NAME = "_iot_config";
```
johnrandolph, line 77: Move to a constant somewhere - where's a good spot that won't be overridden by gencode?

johnrandolph (approved):
LGTM

PR #411, opened 2022-08-08
Restart schema and capability

grafnu:


tests/config.tests/applied.json
```diff
@@ -3,17 +3,17 @@
   "blobset": {
     "blobs": {
       "arbitrary_manufacturer_id": {
```
johnrandolph, line 5: Believe I deleted this file and moved the test to more specific filename blobset...

tests/config.tests/blobset_applied.json
```diff
@@ -3,17 +3,17 @@
   "blobset": {
     "blobs": {
       "arbitrary_manufacturer_id": {
-        "phase": "applied",
```
johnrandolph, line 6: Yes but now rename the file too

tests/config.tests/blobset_applied_incomplete_payload.json
```diff
@@ -3,7 +3,7 @@
   "blobset": {
     "blobs": {
       "arbitrary_manufacturer_id": {
-        "phase": "applied",
```
johnrandolph, line 6: Rename file to match

tests/config.tests/blobset_applied_incomplete_url.json
```diff
@@ -3,7 +3,7 @@
   "blobset": {
     "blobs": {
       "_firmware_update": {
-        "phase": "applied",
```
johnrandolph, line 6: same

schema/common.json
```diff
@@ -2,6 +2,22 @@
...
+        "initial",
+        "active",
+        "restart"
+      ]
+    },
+    "blobsets": {
+      "title": "System Blobsets",
+      "description": "Predefined system blobsets",
+      "enum": [
+        "_iot_config"
```
johnrandolph, line 18: So we can define _firmware_update etc in this enum, reference it with constants, but we aren't using it as a key filter (instead we have open regex on config_blobset.json:15 ) ?

pubber/bin/run
```diff
@@ -2,10 +2,20 @@
 
 ROOT_DIR=$(realpath $(dirname $0)/../..)
 
-java -Dorg.slf4j.simpleLogger.showThreadName=false \
-     -jar $ROOT_DIR/pubber/build/libs/pubber-1.0-SNAPSHOT-all.jar \
-     $@ &
+result=5  # Exit code used to indicate intentional system restart
 
-echo $! > $ROOT_DIR/.pubber.pid
+while [[ $result == 5 ]]; do
```
johnrandolph, line 7: -eq

pubber/bin/run
```diff
@@ -2,10 +2,20 @@
...
+    java -Dorg.slf4j.simpleLogger.showThreadName=false \
+         -jar $ROOT_DIR/pubber/build/libs/pubber-1.0-SNAPSHOT-all.jar \
+         $@ &
 
-wait
+    PUB_PID=$!
+
+    echo $PUB_PID > $ROOT_DIR/.pubber.pid
```
johnrandolph, line 14: Where are we reading this?

tests/config.tests/applied.json
grafnu: They are still in the master branch of faucetsdn -- https://github.com/faucetsdn/udmi/blob/master/tests/config.tests/applied.json

I can delete them, but they haven't been deleted as of yet.

pubber/bin/run
grafnu: removed

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -491,14 +496,39 @@ private void sendMessages() {
...
+      deviceState.system.mode = SystemMode.ACTIVE;
+    }
+  }
+
+  private void restartSystem() {
+    deviceState.system.mode = SystemMode.RESTART;
+    publishStateMessage();
+    System.err.println("Restarting system with extreme prejudice.");
+    System.err.flush();
+    System.exit(RESTART_EXIT_CODE);
```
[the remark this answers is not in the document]
grafnu: Unclear -- the docs out there are pretty sketchy. Seems like people suggest that it follows the standard BASH rules, which means <128 are internal, 128 + 32 are for Unix signals.... I'll map to 192 and call it a day!

pubber/bin/run
grafnu: Done

schema/common.json
grafnu: The enums are things that end up in constants in the code -- but I don't know how to make the schema match "this should be an enum or not start with _" -- There is theoretically a "propertyNames" field in the schema spec, but I can't get it to actually do anything so it's either not really implemented, or implemented in a way I don't expect!

pubber/bin/run
```diff
@@ -2,10 +2,19 @@
...
 
-wait
+    PUB_PID=$!
+
+    result=0
+    echo Waiting for pubber pid $PUB_PID to complete...
+    wait $PUB_PID || result=$?
+done
+
+echo Done with pubber run, exit code $result
```
johnrandolph, line 19: If wait fails we won't know it versus the process exiting with status 0. Do you care? 
If so you could assign wait $PUB_PID; result=$?   instead of ||   and get either wait failure 127 or the exit status.

tests/config.tests/blobset_applied.json
johnrandolph: The phase is now final not applied so

mv blobset_applied.json blobset_final.json

johnrandolph (approved):

johnrandolph (approved):

PR #414, opened 2022-08-09
Creating formal schema definitions for misc config files

grafnu:


schema/pubber_configuration.json
```diff
@@ -0,0 +1,39 @@
...
+      "default": "RS256"
+    },
+    "serialNo": {
+      "type": "string"
+    },
+    "macAddr": {
+      "type": "string"
+    },
+    "keyBytes": {
+      "description": "Should be type byte[], but that can't be represented in json schema"
```
johnrandolph, line 33: What type is it then?

grafnu:
The java code generator creates it as a generic Object, and then explicit
type-casts are used to make it byte[] as needed. I updated the comment to
add that bit of detail.

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -261,6 +265,13 @@ private static void startFeedListener(String projectId, String siteName, String
     }
   }
 
+  private static PubberConfiguration sanitizeConfiguration(PubberConfiguration configuration) {
```
johnrandolph, line 268: this is more like build, right?

johnrandolph (approved):

grafnu:
It's actually taking an existing object in cleaning it up (thus sanizing)
-- not just building one from scratch!

PR #416, opened 2022-08-10
Automatically generate category constants for Java and schema validation

grafnu:
Mostly gencode and test changes which more or less can be ignored. There's a bit of code to generate things in bin/  and then a change to one test to use the new contants.


bin/gencode_categories
```diff
@@ -0,0 +1,96 @@
...
+        out.write(JSON_FORMAT % (prefix, indent, target))
+        prefix = ',\n'
+    out.write('\n')
+
+def write_java_out(categories):
+    os.makedirs(os.path.dirname(java_out), exist_ok=True)
+    with open(java_in) as inp:
+        with open(java_out, 'w') as out:
+            while line := inp.readline():
+                if line.strip().startswith(GENCODE_MARKER):
```
johnrandolph, line 76: Here to? re: find() change

johnrandolph (approved):

noursaidi (approved):

PR #428, opened 2022-08-24
Sequence doc generator

grafnu:
Primarily as FYI and looking for sanity check. It's known "not complete" but the next steps will be to go through test-by-test and clean up as needed, and I didn't want to do this as part of this PR. Right now, it just has the overall framework that is functional and stable, but not complete.


docs/tools/gcloud.md
```diff
@@ -0,0 +1,16 @@
+[**UDMI**](../../) / [**Docs**](../) / [**Tools**](./) / [Registrar](#)
```
noursaidi, line 1: `Registrar` -> `Gcloud`

noursaidi (approved):

PR #431, opened 2022-08-25
Gencode to include inline examples from test directory 

noursaidi:
Draft PR of some code I've done previously to which does a live in-place update of documentation with examples from the examples folder for review.

It includes a `check` function for CI testing, but I just need to have a think about how it integrates (because `gencode check` still  runs `gendocs`, which means the documentation is updated before it's checked)

Usage:
Include anywhere in an MD document `<!--example:state/example.json-->` where `state` is the bit preceding the `.tests` (e.g. event_discovery) and `example.json` is the name of the file to include.
Run `bin/gencode_docs_examples` to update examples in documentation, or `bin/gencode_docs_examples check` to check if the documentation includes latest examples


grafnu:
Do you have an example file where this has been done (.md) file or
something?  I tried it out and it's not working, but there's a lot of
places I was guessing...   Maybe take it to the next level where it's setup
to work with one example somewhere?  (Don't need to worry about the check
part intiaily, just the generation part)

noursaidi:
Yes, in [docs/specs/sequences/endpoint_reconfiguration.md](https://github.com/faucetsdn/udmi/blob/master/docs/specs/sequences/endpoint_reconfiguration.md)

grafnu:
This PR was closed... is it still live?  Seems like it's a good idea still?

noursaidi:
It was merged into #447 and then eventually into master

grafnu:
Brilliant, thanks!

PR #439, opened 2022-09-02
Validator message window and report contents

grafnu:
Addresses multiple issues pertaining to validator. Includes a refactor that processes "recent messages" rather than just the "ast message" -- so should be more generalizable and scalable.


noursaidi (approved):

noursaidi (approved):
Thanks, that's fixed it!

validator/src/main/java/com/google/daq/mqtt/util/Common.java
```diff
@@ -30,4 +33,42 @@ static String capitalize(String str) {
...
   }
+
+  /**
+   * Get the code line where the exception occurred (based on package name).
+   *
+   * @param e         Throwable to trace
+   * @param container containing class to use as package prefix
+   * @return string indicating the offending line
+   */
+  public static String getExceptionLine(Throwable e, Class<?> container) {
```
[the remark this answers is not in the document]
grafnu: Actually the opposite -- if it's a ValidationException then it's something that much of the code would expect and know how to report intelligently. This is more for the non-ValidationException cases when it would otherwise just be a way-too-detailed expansion of the stack trace. It will work with a ValidationException, just much less needed.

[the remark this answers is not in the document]
johnrandolph: Ok great. That is what I was concerned about, i.e. making complicated parsers instead of smart exceptions.

validator/src/main/java/com/google/daq/mqtt/registrar/DeviceExceptionManager.java
```diff
@@ -9,15 +9,15 @@
 import java.util.Optional;
 import java.util.regex.Pattern;
 
-public class DeviceExceptionManager {
+class DeviceExceptionManager {
```
johnrandolph, line 12: Is this a lint-based change?

validator/src/main/java/com/google/daq/mqtt/util/FileDataSink.java
```diff
@@ -68,7 +68,9 @@ private File getOutputFile(String deviceId, String topic) {
   private File getOutputFile(String deviceId, String subType, String subFolder, String suffix) {
     File deviceDir = getDeviceDir(deviceId);
     deviceDir.mkdirs();
-    return new File(deviceDir, String.format("%s_%s.%s", subType, subFolder, suffix));
+    boolean invalid = subFolder == null || "update".equals(subFolder);
+    return new File(deviceDir,
+        String.format("%s%s.%s", subType, invalid ? "" : "_" + subFolder, suffix));
```
johnrandolph, line 73: Suggestion, use the string format to without additional tiny + string concat in the arguments?

validator/src/main/java/com/google/daq/mqtt/registrar/DeviceExceptionManager.java
grafnu: yes -- public classes require javadoc, and there's no need for this to be public since it's only used internally.

validator/src/main/java/com/google/daq/mqtt/util/FileDataSink.java
grafnu: Updated to make it, er, less ugly.

johnrandolph (approved):

PR #445, opened 2022-09-07
update external_refs with learning material links

noursaidi:


noursaidi:
Notice I'd included some `*.md.md` files which obviously don't exist but there was also some errors, so will fix and check the CI test

noursaidi:
List updated and CI test fixed to fail if a reference is missing

grafnu (approved):

PR #447, opened 2022-09-08
Endpoint redirection documentation

noursaidi:
https://noursaidi.github.io/udmi/docs/specs/sequences/endpoint_reconfiguration


noursaidi:
@johnrandolph PTAL only at the sequence diagrams - are they correct with what the sequencer tests? Quality of other information can be sorted later ..

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -0,0 +1,47 @@
...
+
+The [endpoint configuration blob](https://github.com/faucetsdn/udmi/blob/master/tests/configuration_endpoint.tests/simple.json) is a JSON object defined by
+[configuration_endpoint.json](https://faucetsdn.github.io/udmi/gencode/docs/configuration_endpoint.html), which is base64 encoded in the config message.
+
+
+## Tests
+
+### Valid Endpoint (Successful) Reconfiguration
+
+```mermaid
```
grafnu, line 16: Is there a place we can go to see a rendered version of this?

noursaidi:
@grafnu - https://noursaidi.github.io/udmi/docs/specs/sequences/endpoint_reconfiguration

But the states `final` and `applied` don't appear to be in the schema so either I might be wrong about those?

grafnu:
Don't worry about the schema -- getting the sequence diagrams correct will
be easier (since we can see all the parts and how they relate), then we go
back to update the schema.

Just a few notes -- I wouldn't put the device in the center of the two
endpoints -- generally the device is on the left. Yah, some arrows may be
longer but it's really confusing to have the main thing in the middle!

The final state should be the same as the intended config. Right now the
config is "final" and state is "applied" -- but it should be a static
description of the end state. So "final" in this case.

The intermediate state should be something like "apply" to keep the same
word tense as "final" -- and it's not really preparing anything, it's
actually applying the changes...

We should also have a CONFIG message from the new endpoint in there.  I
think we'll want some cloud-side locking mechanism (how do you know you go
to the right endpoint?) -- but at the very least just acknowledge that
there is another CONFIG.

But overall the sequences look great!

noursaidi:
@grafnu comments incorporated.

Re the locking mechanism, is that an additional test represented as another sequence diagram - "If the config which initiates the redirect includes a nonce, only commit the new endpoint if the new endpoint if the device receives a config from the new endpoint with the same nonce"? Or is it core functionality?

docs/specs/sequences/endpoint_reconfiguration.md
grafnu: https://noursaidi.github.io/udmi/docs/specs/sequences/endpoint_reconfiguration.html

tests/config.tests/endpoint_reconfiguration.json
```diff
@@ -0,0 +1,13 @@
+{
+  "version": 1,
+  "blobset": {
+    "blobs": {
+      "_iot_endpoint_config": {
```
johnrandolph, line 5: This is correct

tests/state.tests/endpoint_reconfiguration.json
```diff
@@ -0,0 +1,23 @@
...
+      "firmware": "1.2"
+    },
+    "serial_no": "000000",
+    "last_config": "2022-07-13T12:00:00.000Z",
+    "operational": true
+  },
+  "blobset": {
+    "blobs": {
+      "_iot_endpoint_config": {
+        "phase": "final"
```
johnrandolph, line 19: This is correct

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -0,0 +1,112 @@
...
+%%{wrap}%%
+sequenceDiagram
+    autonumber
+    participant D as Device
+    participant E as Original Endpoint
+    participant E' as New Endpoint
+    E->>D:CONFIG MESSAGE<br>blobset.blobs._iot_endpoint_config.blob = <ENDPOINT><br>blobset.blobs._iot_endpoint.blob.phase = "final"
+    D->>E:STATE MESSAGE<BR>blobset.blobs._iot_endpoint_config.blob.phase = "preparing"
```
johnrandolph, line 45: I don't understate this state. Would we not also see this in the success state?

noursaidi: Trevor noted it should be apply rather than preparing, but I didn't realise I only changed one of the diagrams. So yes, both would enter the `apply` phase after receiving an endpoint configuration - if that's answered your question

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -22,11 +22,11 @@ sequenceDiagram
     participant D as Device
     participant E as Original Endpoint
     participant E' as New Endpoint
-    E->>D:CONFIG MESSAGE<br>blobset.blobs._iot_endpoint_config.base64 = <ENDPOINT><br>blobset.blobs._iot_endpoint.blob.phase = "final"
-    D->>E:STATE MESSAGE<BR>blobset.blobs._iot_endpoint_config.blob.phase = "apply"
+    E->>D:CONFIG MESSAGE<br>blobset.blobs._iot_endpoint_config.base64 = <ENDPOINT><br>blobset.blobs._iot_endpoint.phase = "final"
```
johnrandolph, line 25: _iot_endpoint_config ?

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -22,11 +22,11 @@ sequenceDiagram
...
+    D->>E:STATE MESSAGE<BR>blobset.blobs._iot_endpoint_config.phase = "apply"
     D-->>E':CONNECTION ATTEMPT
     E'->>D:CONFIG MESSAGE
-    D->>E':STATE MESSAGE<BR>blobset.blobs._iot_endpoint_config.blob.phase = "final"
+    D->>E':STATE MESSAGE<BR>blobset.blobs._iot_endpoint_config.phase = "final"
```
grafnu, line 29: I think we should add an additional config/state sequence at the end that starts with `config.blobset.blobs._iot_endpoint_config = null` and then we verify that the `state.blobset.blobs._iot_endpoint_config == null`  The fundamental reason is that we want to verify that the state block only shows up when expected.

docs/specs/sequences/endpoint_reconfiguration.md
grafnu, line 25: pedantically there should be a check at the start that verifies that `state.blobset.blobs._iot_endpoint_config == null` (as the config block is empty so there should be no corresponding state).

docs/specs/sequences/endpoint_reconfiguration.md
grafnu, line 29: there should be a check in here that the system.last_update time of the state message matches the config timestamp from the new endpoint (so might need to indicate that too). Goal is to check that they don't just connect and then use the wrong reply.

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -0,0 +1,142 @@
...
+%%{wrap}%%
+sequenceDiagram
+    autonumber
+    participant D as Device
+    participant E as Original Endpoint
+    participant E' as New Endpoint
+    E->>D:CONFIG MESSAGE<br>blobset.blobs._iot_endpoint_config.blob = <ENDPOINT><br>blobset.blobs._iot_endpoint.blob.phase = "final"
+    D->>E:STATE MESSAGE<BR>blobset.blobs._iot_endpoint_config.phase = "apply"
+    D-->>E':CONNECTION ATTEMPT
+    note over D: Failure, e.g. endpoint doesn't exist, incorrect credentials, ...
```
grafnu, line 44: what about a time limit here?  I'd be worried that somebody implements a "try for 3h before reporting failure" scenario... so we should just have a fixed time-bound (which will be needed for the test anyway)... 30s?  We can always increase it later, and devices can always report failure faster.

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -0,0 +1,142 @@
...
+    D-->>E:CONNECTION ATTEMPT
+    E->>D:CONFIG MESSAGE
+    D->>E:STATE MESSAGE<BR>blobset.blobs._iot_endpoint_config.phase = "final"<BR/>blobset.blobs._iot_endpoint_config.status.level=500 (ERROR)
```
grafnu, line 47: somewhere we need to introduce a category(s) for this too...

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -0,0 +1,142 @@
...
+    participant E as Original Endpoint
+    participant E' as New Endpoint
+    E->>D:CONFIG MESSAGE<br>blobset.blobs._iot_endpoint_config.base64 = <ENDPOINT><br>blobset.blobs._iot_endpoint.phase = "final"
+    D->>E:STATE MESSAGE<BR>blobset.blobs._iot_endpoint_config.phase = "apply"
+    D-->>E':CONNECTION ATTEMPT
+    E'->>D:CONFIG MESSAGE
+    D->>E':STATE MESSAGE<BR>blobset.blobs._iot_endpoint_config.phase = "final"
+```
+
+### Invalid Endpoint (Unsuccessful Reconfiguration)
```
grafnu, line 32: probably not something to do now, but something to think about, is that there's another category of "invalid" here which is an invalid request -- i.e. the base64 is borked or the resulting JSON is not parsable.  In a perfect world that would be just another test and then another error category (which is semantically different than "we understood the request but that failed")...  Similar to writeback. There are clear stages where this is different so I think we can clearly separate out the tests.  (E.g. to make sure the JSON parser doesn't barf and bring the system down!)

grafnu (commented):
How come I never just have one simple comment for anything??!?!

docs/specs/sequences/endpoint_reconfiguration.md
noursaidi: Do we differentiate for different types of failure? Or is it a single `endpoint.reconfig.failed` or similar? 
A pretty common error I've encountered before is an endpoint "unreachable" because of firewall or network restrictions. Would that be its own category, or is that level of detail left to the message?

docs/specs/sequences/endpoint_reconfiguration.md
noursaidi: How long does `phase: final` stay in the state message for? Atleast 1? Noting also that the new endpoint probably won't have a `blobset.blob._iot_endpoint_config` blob

grafnu:
I try to think of it as the categories delineate the various phases
something would go through -- e.g. for config there's { receive, parse,
apply } -- but, the actual "why did parsing fail?" it left up to the
message. E.g. it could be a base64 parse error, or JSON parse error, or...
--- no need to categorize all that. So for this, I think there's the
obvious categories of { receive, parse, apply } -- like, I received the
update, I could understand the update, and then I could apply the update --
but the actual underlying reasons (firewall bad) wouldn't be categories.

Also, it's highly unlikely that a device would know that a firewall is
causing a problem. So... not something that would be a useful
categorization from that perspective.

grafnu:
The final should be there as long as the blob is in the config. If the blob
data is what is applied, then it's final. If the data is different, then
it's initial or some other phase.

If the new endpoint has the blob config or not is, well, up to the backend.
So, if the new endpoint wants the blob state to be defined then it needs to
have it in the config. If it doesn't care then it doesn't need it there.

Basically, we should be able to clearly articulate the device logic:
1) blob should be in state if and only if it is in state
2) If blob matches the finalized state (fully applied or errored) then it's
final, otherwise it's not final.

docs/specs/sequences/endpoint_reconfiguration.md
noursaidi: Changed

docs/specs/sequences/endpoint_reconfiguration.md
noursaidi: Added note for 30 seconds

noursaidi:
> So for this, I think there's the
> obvious categories of { receive, parse, apply } -- like, I received the
> update, I could understand the update, and then I could apply the update --
> but the actual underlying reasons (firewall bad) wouldn't be categories.
Okay, using `blobset.blob` as the prefix e.g. `blobset.blob.apply`

> Also, it's highly unlikely that a device would know that a firewall is
> causing a problem. So... not something that would be a useful
> categorization from that perspective.
Yes, butthe device can differentiate "could not reach host", "could connect to host but no MQTT broker" and "MQTT broker denied connection (usually incorrect credentials)". But I think you're notes are fine, the message is sufficient and it's not anything that can't be deduced from other means too.

docs/specs/sequences/endpoint_reconfiguration.md
noursaidi: Category added

docs/specs/sequences/endpoint_reconfiguration.md
noursaidi: Added

noursaidi:
@johnrandolph @grafnu  I think I've incorporated comments, please take a look.

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -0,0 +1,150 @@
...
+    participant E as Original Endpoint
+    participant E' as New Endpoint
+    E->>D:CONFIG MESSAGE<br>timestamp = t0<br/>blobset.blobs._iot_endpoint_config = null (unset)
+    D->>E:STATE MESSAGE<br/>system.last_update = t0<br/>blobset.blobs._iot_endpoint_config = null (unset)
+    E->>D:CONFIG MESSAGE<br>timestamp = t1<br/>blobset.blobs._iot_endpoint_config.base64 = <ENDPOINT><br>blobset.blobs._iot_endpoint_config.phase = "final"
+    D->>E:STATE MESSAGE<br/>system.last_update = t1<br/>blobset.blobs._iot_endpoint_config.phase = "apply"
+    D-->>E':CONNECTION ATTEMPT
+    E'->>D:CONFIG MESSAGE<br>timestamp = t2<br/>blobset.blobs._iot_endpoint_config.base64 = <ENDPOINT><br>blobset.blobs._iot_endpoint_config.phase = "final"
+    D->>E':STATE MESSAGE<br/>system.last_update = t2<br/>blobset.blobs._iot_endpoint_config.phase = "final"
+    E->>D:CONFIG MESSAGE<br/>timestamp = t3<br/>blobset.blobs._iot_endpoint_config = null (unset)
```
grafnu, line 34: Should be from E' not E for the closing null config.

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -0,0 +1,150 @@
...
+[configuration_endpoint.json](https://faucetsdn.github.io/udmi/gencode/docs/configuration_endpoint.html), which is base64 encoded in the config message.
+
+
+## Tests
+
+### Valid Endpoint (Successful Reconfiguration)
+
+**Notes**
+- `<ENDPOINT>` is a **base64** encoded endpoint object
+- `blobset.blobs_iot_endpoint_config` block is only present in a device's state messages when it is present in the last received config message
```
grafnu, line 18: "is present if and only if a" --  abit more strict definition than you have it.

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -0,0 +1,150 @@
...
+sequenceDiagram
+    autonumber
+    participant D as Device
+    participant E as Original Endpoint
+    participant E' as New Endpoint
+    E->>D:CONFIG MESSAGE<br>timestamp = t0<br/>blobset.blobs._iot_endpoint_config = null (unset)
+    D->>E:STATE MESSAGE<br/>system.last_update = t0<br/>blobset.blobs._iot_endpoint_config = null (unset)
+    E->>D:CONFIG MESSAGE<br>timestamp = t1<br/>blobset.blobs._iot_endpoint_config.base64 = <ENDPOINT><br>blobset.blobs._iot_endpoint_config.phase = "final"
+    D->>E:STATE MESSAGE<br/>system.last_update = t1<br/>blobset.blobs._iot_endpoint_config.phase = "apply"
+    D-->>E':CONNECTION ATTEMPT
```
grafnu, line 31: I'd put this in a "<30s" loop box as well, for consistency with the other diagram.  It doesn't need anything else inside it, but it clarifies the overall structure, I think.

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -0,0 +1,150 @@
...
+
+```mermaid
+%%{wrap}%%
+sequenceDiagram
+    autonumber
+    participant D as Device
+    participant E as Original Endpoint
+    participant E' as New Endpoint
+    E->>D:CONFIG MESSAGE<br>timestamp = t0<br/>blobset.blobs._iot_endpoint_config = null (unset)
+    D->>E:STATE MESSAGE<br/>system.last_update = t0<br/>blobset.blobs._iot_endpoint_config = null (unset)
```
grafnu, line 28: These first two messages really are just "quiescent state" before the sequence, but actually aren't part of the sequence itself, so I'm not sure if they should be included (just because "more" is often worse than "elegant" :-).  If they follow the if-and-only-if guidelines then this is implied (so doesn't need to be explicitly stated), and I'm not sure there'd be much confusion...

docs/specs/categories.md
```diff
@@ -46,7 +46,7 @@ implicit expected `level` values, indicated by '(**LEVEL**)' in the hierarchy be
   * _blob_: Conditions specific to an individual blob
     * _receive_: (**DEBUG**) About receiving a blob update
     * _fetch_: (**DEBUG**) Fetching a blob update
-    * _apply_: (**NOTICE**) Applying a blob update
+    * _apply_: (**NOTICE**) Applying a blob update. **ERROR** When there is a failure applying a blob update
```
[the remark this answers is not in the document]
noursaidi: Got it, undone change.

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -0,0 +1,150 @@
...
+    autonumber
+    participant D as Device
+    participant E as Original Endpoint
+    participant E' as New Endpoint
+    E->>D:CONFIG MESSAGE<br/>timestamp = t0<br/>blobset.blobs._iot_endpoint_config.blob = <ENDPOINT><br>blobset.blobs._iot_endpoint_config.blob.phase = "final"
+    D->>E:STATE MESSAGE<br/>system.last_update = t0<br/>blobset.blobs._iot_endpoint_config.phase = "apply"
+    critical: Total duration < 30 seconds
+    D-->>E':CONNECTION ATTEMPT
+    note over D: Failure, e.g. endpoint cannot be reached, incorrect credentials...
+    D-->>E:CONNECTION ATTEMPT
```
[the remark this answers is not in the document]
noursaidi: The second connection attempt (back to the original working endpoint) moved out of the critical section (also now called a loop box)

docs/specs/sequences/endpoint_reconfiguration.md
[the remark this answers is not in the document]
noursaidi: `(unset)` removed

docs/specs/sequences/endpoint_reconfiguration.md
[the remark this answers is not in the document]
noursaidi: I was using `t0`, `t1`, etc.. to indicate a new different config message. The specific intent was to show the state message sent after a reconnect is in response to the config message sent from the new endpoint

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -0,0 +1,150 @@
...
+The [endpoint configuration blob](https://github.com/faucetsdn/udmi/blob/master/tests/configuration_endpoint.tests/simple.json) is a JSON object defined by
+[configuration_endpoint.json](https://faucetsdn.github.io/udmi/gencode/docs/configuration_endpoint.html), which is base64 encoded in the config message.
+
+
+## Tests
+
+### Valid Endpoint (Successful Reconfiguration)
+
+**Notes**
+- `<ENDPOINT>` is a **base64** encoded endpoint object
```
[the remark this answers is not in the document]
noursaidi: `<ENDPOINT>` changed to `<NEW_ENDPOINT>`

docs/specs/sequences/endpoint_reconfiguration.md
noursaidi: missed that - correcteD

docs/specs/sequences/endpoint_reconfiguration.md
noursaidi: Put into a loop box

docs/specs/sequences/endpoint_reconfiguration.md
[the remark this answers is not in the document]
noursaidi: Maybe just a note to say _`last_update` matches timestamp of the config message received from the new endpoint_?

bin/gencode_docs_examples
```diff
@@ -0,0 +1,143 @@
...
+  parser = argparse.ArgumentParser()
+
+  parser.add_argument('check', nargs='?', default=False,
+            help='Check in-line examples match examples in tests directory')
+
+  return parser.parse_args()
+
+
+def validate_code_blocks(file_contents):
+  """ Run some very primitive checks to validate templated placeholder code
```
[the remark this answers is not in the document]
noursaidi: I noticed I was doing both, so I wasn't sure if this was for consistency or preference. The examples in [Google style guide](https://google.github.io/styleguide/pyguide.html) seems to suggest same line for Python, so I've changed all to match, but can change it either way

docs/specs/sequences/endpoint_reconfiguration.md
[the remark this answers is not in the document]
grafnu: But it's entirely redundant. Every config message is followed by a state message with a matching timestamp.  What problem are you trying to "fix" by having that information there? What loss of information is there if it's taken out?  My point is that 1) it causes more questions than it answers, and 2) is just more "noise" without much signal behind it. Every other line in the sequence is indicating something that's specific/unique to endpoint sequences... except for that.

bin/gencode_docs_examples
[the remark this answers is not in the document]
grafnu: Really it's about consistency!  So, that's fine as you have it now.

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -0,0 +1,150 @@
...
+[configuration_endpoint.json](https://faucetsdn.github.io/udmi/gencode/docs/configuration_endpoint.html), which is base64 encoded in the config message.
+
+
+## Tests
+
+### Valid Endpoint (Successful Reconfiguration)
+
+**Notes**
+- `<NEW_ENDPOINT>` is a **base64** encoded endpoint object
+- `blobset.blobs_iot_endpoint_config` is present in a device's state message if, and only if,   the last received config message has a `blobset.blobs_iot_endpoint_config` block.
```
grafnu, line 18: missing . after blobs (two places)

docs/specs/sequences/endpoint_reconfiguration.md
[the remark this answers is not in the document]
noursaidi: @grafnu I was trying to address this comment - https://github.com/faucetsdn/udmi/pull/447#discussion_r971173499

grafnu:
Ah, well it seemed like a better idea when I said it the first time :-).
What might work better would be a note over the "new endpoint" that says
something like "state.timestamp matches new endpoint config.timestamp" --
to be explicit about it.  As it stands now, although you're 100% right, it
kinda goes unnoticed (even by me!) because it does look just like normal
message flow. We don't really indicate all the usual checks (like for the
phase, etc...) and I don't think we should, but this seems like a special
case?  Sorry for all the confusion and back/forth...

grafnu (approved):
Just one minor comment about the position of the state comparison note, but otherwise all looks good to me!

grafnu:
I think it would just be one diagram that shows the overall flow. I'm not
sure of the utility in breaking it out into smaller things unless the
smaller things have some meaning behind them (e.g. ability to restart is
useful independent of redirection).  I added in some fields I think would
work well for this -- they're borrowed from the restart mechanism (mode and
last_start).  Everything *underlined* is new, everything else is the same.
Some of the system config messages (e.g. mode and last_start) aren't part
of the redirect sequence, but they're necessary things that need to be here
at this point (but could be earlier) -- so not 100% sure how to diagram
them.

CONFIG MESSAGE
*system.mode = "active"*
*system.last_start = <LAST_START>*
blobset.blobs._iot_endpoint.blob = <ENDPOINT>
blobset.blobs._iot_endpoint.blob.phase = "final"
STATE MESSAGE
blobset.blobs._iot_endpoint.blob.phase = "apply"
*system.mode = "reconnect"*
CONNECTION ATTEMPT
*CONNECTION* SUCCESS
CONFIG *MESSAGE*
*system.last_start = <LAST_START>*
STATE MESSAGE
blobset.blobs._iot_endpoint.blob.phase = "applied"
*system.mode = "active"*

PR #448, opened 2022-09-08
Add tool to view device GCP cloud logging entries

noursaidi:
Adding link to some of the scripts I put together which helped me diagnose devices

### DIFF
--- a/docs/tools/gcloud.md
+++ b/docs/tools/gcloud.md
@@ -18,3 +18,15 @@ them to work out-of-the-box without a deeper understanding of what's going on!
 # Update a device's GCP IoT Core configuration
 
 `bin/reset_config`
+
+# Viewing Device GCP Cloud Logging logs by Device ID
+
+Using `log.py` from [noursaidi/devicetrace](https://github.com/noursaidi/devicetrace)
+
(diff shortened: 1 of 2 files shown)


docs/tools/gcloud.md
```diff
@@ -18,3 +18,15 @@ them to work out-of-the-box without a deeper understanding of what's going on!
 # Update a device's GCP IoT Core configuration
 
 `bin/reset_config`
+
+# Viewing Device GCP Cloud Logging logs by Device ID
+
+Using `log.py` from [noursaidi/devicetrace](https://github.com/noursaidi/devicetrace)
```
grafnu, line 24: You shouldn't reference your person repo here -- so, either these need to be moved into the main faucetsdn/ one, or, well... they shouldn't be here. Is there anything special about these files that we can't move into the main udmi repo somewhere?

noursaidi: No reason they can't be included.

I've updated this PR to include the utility to view GCP which I think is useful. The others need a little of work before they can be included if they're needed.

grafnu (approved):
Nice!

PR #450, opened 2022-09-09
Correct writeback_state sequence test

noursaidi:
Correct behaviour of writeback_state sequence test


noursaidi:
@grafnu what needs to be done to fix the generated.md as below:

Looking at the other sequences, it should at least include the change config, e.g.:
```
1. Update config for writeback
 * Set "pointset.points.filter_alarm_pressure_status.set_value" = false
 * Set "pointset.points.filter_differential_pressure_setpoint.set_value" = 60
 * Set "pointset.points.filter_differential_pressure_sensor.set_value" = 15
```

And ideally it would include some of the metadata around the tests
```
invalidTarget.target_point = invalidTarget.target_value
failureTarget.target_point = failureTarget.target_value
appliedTarget.target_point = appliedTarget.target_value
```

grafnu:
When you say "include some of the metadata" -- do you mean that the values
are showing up in the config sent to the device, but not in the
generated.md?  Or is it that the values aren't even showing up in the
config?  Basically, what is "it" in that sentence?

noursaidi:
At the moment the values/config diffs don't show up in generated.md at all. 

The specific values would not be documented (as they're device specific), rather what they represent should 

So looking at one line from the example in my comment, Instead of showing something like 
` * Set "pointset.points.filter_alarm_pressure_status.set_value" = false`. 

The point `filter_alarm_pressure_status.set_value` and the value `false` both come from `invalidTarget.target_point ` and `invalidTarget.target_value` in the Java code respectively. 

Ideally, we would document in generated.md something generic e.g. ` * Set "pointset.points.INVALID_POINT.set_value" = INVALID_VALUE` instead

noursaidi:
Thanks Trevor for you PR, seems to have fixed the bug so `generated.md` now accurately represents

But it's introduce an error with the `single_scan` and `periodic_scan` - `Test failed: Unexpected non-semantic Date in semantic value calculation` - relevant diffs and logs in this test run -> https://github.com/noursaidi/udmi/actions/runs/3044131852/jobs/4904263182#step:5:11915

grafnu (commented):
Oy, I think this needs a merge with master before review!  (GitHub algorithms seem to be bad??!)

noursaidi:
@grafnu made a mistake when I did a merge hence the issues you saw (I think GitHub's algos were fine)

There are still 249 modified files, but the bulk of these are the sequencer outputs in `validator/sequences`, I copied in my latest run. All other changed files are as expected and should be reviewed.

bin/clone_model
```diff
@@ -5,7 +5,7 @@ mkdir -p $ROOT_DIR/sites
 cd $ROOT_DIR/sites 
 
 MODEL_DIR=udmi_site_model
-MODEL_VER=1.7
+MODEL_VER=808aa06fbe922b9f749d2a6b52ff03b6f9074b48
```
grafnu, line 8: needs 1.8 tag

validator/sequences/broken_config/config_pointset.attr
```diff
@@ -6,5 +6,5 @@
   "subType" : "config",
   "category" : "commands",
```
grafnu, line 7: I think these should be added to .gitattributes to ignore the diffs so it's clear to the reviewer!

noursaidi: What files need to be included in this directory? Without any redaction, all the files will change with every run except for the `sequence.md`? Should we only include `sequence.md` and ignore the rest?

grafnu:
Ok, tag 1.8 should be ready!

noursaidi:
Great, CI test all clear now so should be mergeable. I've also made the requested to `.gitattributes`.

I was just stating all the files will be modified when they're copied over (because timestamps, etc),  including for those tests which did not change. I was just wondering if all the files were needed or some redaction was needed, but as you said its good to have the complete set

grafnu:
Nour, can you separate out all the "I don't care" log files into a separate PR and then either push that through or have this PR against that one?  Even though they're not visible by default, it's near impossible to review this PR because of the sheer number of "must ignore" files, and there's ways around that!

validator/src/main/java/com/google/daq/mqtt/util/ConfigDiffEngine.java
```diff
@@ -40,16 +37,24 @@ public List<String> computeChanges(Config deviceConfig) {
...
+    if (thing == null) {
+      return ImmutableMap.of();
+    }
+    if (thing instanceof Map) {
+      Map<String, Object> asMap = (Map<String, Object>) thing;
+      return asMap.keySet().stream()
+          .collect(Collectors.toMap(key -> key, key -> convertSemantics(asMap.get(key))));
+    }
+    return Arrays.stream(thing.getClass().getFields())
+        .filter(field -> isNotNull(thing, field)).collect(
```
noursaidi, line 51: @grafnu this change appears to be causing a stack overflow on the endpoint reconfig tests and I can't figure it out -  https://github.com/noursaidi/udmi/actions/runs/3090008247/jobs/4998355934#step:5:13153

.github/workflows/testing.yml
```diff
@@ -111,6 +111,9 @@ jobs:
           more /tmp/sequencer.out
           diff -u /tmp/sequencer.out etc/sequencer.out
           diff -u /tmp/generated.md docs/specs/sequences/generated.md
+          ls -1 validator/sequences | xargs -I% diff -u \
```
grafnu, line 114: be warned, what I discovered the other day is that the shell in GitHub Actions doesn't always process failure conditions correctly. Did you check that this actually fails the run when there's a problem? ( You could quickly check by tweaking the validator/sequences/%/sequence.md file)

grafnu (approved):

.github/workflows/testing.yml
noursaidi: Thanks, I've checked and it does fail when there is a difference or missing file.

Issue #443, opened 2022-09-06
Registrar - Running against a single device blocks all devices in a registry


noursaidi:

grafnu:
Addressing with https://github.com/faucetsdn/udmi/pull/451

PR #451, opened 2022-09-11
Don't block registar devices, and refactor a bit to enable Registrar unit testing

grafnu:


noursaidi (approved):
Quickly checked that it still works and that it fixed blocked devices bug #443

validator/src/main/java/com/google/daq/mqtt/util/IotProvider.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.google.api.services.cloudiot.v1.model.Device;
+import java.util.AbstractMap.SimpleEntry;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Abstraction for a cloud-based IoT provider.
+ */
+public interface IotProvider {
```
johnrandolph, line 11: Doc strings that explain the interface?

grafnu: updated

johnrandolph (approved):

PR #456, opened 2022-09-16
Add web-server invoker and test refactoring

grafnu:


validator/src/main/java/com/google/daq/mqtt/util/SimpleWebServer.java
```diff
@@ -0,0 +1,80 @@
...
+  private final HttpServer server;
+  private final Map<String, Consumer<Map<String, String>>> mockHandlers;
+
+  protected SimpleWebServer(List<String> args) {
+    try {
+      if (args.size() < 1) {
+        throw new RuntimeException("Need server port argument");
+      }
+      int port = Integer.parseInt(args.remove(0));
+      if (port > 0) {
```
johnrandolph, line 35: Suggest > 1023 for sanity

johnrandolph (approved):

validator/src/main/java/com/google/daq/mqtt/util/SimpleWebServer.java
grafnu: Done

PR #459, opened 2022-09-19
Fix for disabled CI integration testing and endpoint sequence test

grafnu:


validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceRunner.java
```diff
@@ -98,7 +98,9 @@ public abstract class SequenceRunner {
   private static final File resultSummary;
   private static final IotReflectorClient client;
   private static final String VALIDATOR_CONFIG = "VALIDATOR_CONFIG";
-  private static final String CONFIG_PATH = System.getenv(VALIDATOR_CONFIG);
+  public static final String DEFAULT_CONFIG = "/tmp/validator_config.json";
```
johnrandolph, line 101: future_nit: Eventually this relying on /tmp thing won't be sufficient

johnrandolph (approved):

PR #463, opened 2022-09-21
add "commit" to endpoint reconfiguration 

noursaidi:
Preview link - https://noursaidi.github.io/udmi/docs/specs/sequences/endpoint_reconfiguration.html

Essentially normal redirect with a box that says reboot .. One question is, given the first sequence diagram checks all "phases" does the second diagram need to do it too?

### DIFF
--- a/docs/specs/sequences/endpoint_reconfiguration.md
+++ b/docs/specs/sequences/endpoint_reconfiguration.md
@@ -67,19 +67,9 @@ sequenceDiagram
     participant D as Device
     participant E as Original Endpoint
     participant E' as New Endpoint
-    E->>D:CONFIG MESSAGE<br/>blobset.blobs._iot_endpoint_config.base64 = <NEW_ENDPOINT><br/>blobset.blobs._iot_endpoint_config.phase = "final"
-    D->>E:STATE MESSAGE
-    loop Total duration < 30 seconds
-    D-->>E':CONNECTION ATTEMPT
-    end
-    E'->>D:CONFIG MESSAGE
-    D->>E':STATE MESSAGE
-    loop UDMI restart sequence
-    E'->>D:CONFIG MESSAGE<br/>(to trigger UDMI restart)
-    note over D: Device restarts
-    D-->>E':CONNECTION ATTEMPT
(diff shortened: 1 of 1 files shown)


grafnu:
So I would say no -- it ultimately adds a lot of extra stuff with no value.
I'd just start with the state (phase == final), and then enough afterwards
to indicate that it's sticky (e.g. connects to the new endpoint not the old
one).  I would expand the "REBOOT DEVICE" box to indicate that it's a
sequence from the new config (not sure if we want to add in an arrow for
that), and then a corresponding box like before that says the state last
update should match the last config (from reboot).  Not 100% sure, but at
least that's what I'd try as the next iteration.

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -57,6 +57,21 @@ sequenceDiagram
...
+
+```mermaid
+%%{wrap}%%
+sequenceDiagram
+    autonumber
+    participant D as Device
+    participant E as Original Endpoint
+    participant E' as New Endpoint
+    D->>E':STATE MESSAGE<br/>blobset.blobs._iot_endpoint_config.phase = "final"
+    note over D,E': Reboot sequence from new endpoint
```
grafnu, line 71: nit: New Endpoint (capital) since it's a specific thing

docs/specs/sequences/endpoint_reconfiguration.md
```diff
@@ -57,6 +57,21 @@ sequenceDiagram
...
+    D->>E':STATE MESSAGE<br/>blobset.blobs._iot_endpoint_config.phase = "final"
+```
+
```
grafnu, line 74: Maybe add a simple note after that just says "This just means the new endpoint connection is remembered after a system restart" I dunno -- it's one of those "it's so simple I'm not sure people will see it!" situations...

docs/specs/sequences/endpoint_reconfiguration.md
grafnu, line 71: nit: I'm calling it "restart" not "reboot" not sure what the exact difference would be -- but I was going with the terminology my laptop uses :-).

grafnu (approved):

PR #464, opened 2022-09-21
Add blob nonce and endpoint_config_connection_success test

johnrandolph:


pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -837,6 +837,13 @@ private void extractEndpointBlobConfig() {
       String iotConfig = extractConfigBlob(IOT_ENDPOINT_CONFIG.value());
       extractedEndpoint = iotConfig == null ? null
           : OBJECT_MAPPER.readValue(iotConfig, EndpointConfiguration.class);
+      if (extractedEndpoint != null) {
+        // TODO: Refactor extractConfigBlob() to get any blob meta parameters like nonce.
+        if (deviceConfig.blobset.blobs.get(IOT_ENDPOINT_CONFIG.value()) != null) {
```
grafnu, line 842: nit: there's a .containsKey() method that's preferred (according to "them") over the .get() != null check...

schema/configuration_endpoint.json
```diff
@@ -20,6 +20,9 @@
     },
     "client_id": {
       "type": "string"
+    },
+    "nonce": {
+      "type": "string"
```
grafnu, line 25: matching pattern for nonce?

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/BlobsetSequences.java
```diff
@@ -46,4 +64,26 @@ public void endpoint_config_connection_error() {
     });
   }
 
+  @Test
+  @Description("Push endpoint config message to device that results in success.")
+  public void endpoint_config_connection_success() {
+    BlobBlobsetConfig config = new BlobBlobsetConfig();
+    config.phase = BlobPhase.FINAL;
+    config.base64 = endpointConfigBase64PayloadHostname(ENDPOINT_CONFIG_HOSTNAME);
```
grafnu, line 72: method name seems off -- should the "Hostname" be there?

johnrandolph: Right, for a while I had a different function that generated a payload with client_id. But now, it's removed. So I also renamed the function.

schema/configuration_endpoint.json
johnrandolph: Copied from blob "nonce" regex.

grafnu (commented):
Code looks ok, is there a passing CI sequencer run of this somewhere?

PR #471, opened 2022-09-26
gittools update and venv version check

grafnu:


johnrandolph (approved):
Worth a try--let's see how it improves (or breaks) our workflow

PR #472, opened 2022-09-27
UDMIF web app console updates

mspratt-biot:
Updates include displaying site and device validation messages in the UI. The webs Angular version has been bumped to version 14, and the Angular Social Login package has been bumped to the latest to accommodate the usage of the new GIS APIs.


bin/pubber
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/usr/local/bin/bash -e
```
grafnu, line 1: I don't think this is a universally valid path...

pubber/build.gradle
```diff
@@ -69,6 +69,7 @@ dependencies {
     }
     implementation 'joda-time:joda-time:2.10.10'
     implementation 'org.eclipse.paho:org.eclipse.paho.client.mqttv3:1.2.5'
+    implementation 'com.bugsnag:bugsnag:3.6.3'
```
grafnu, line 72: does this do anything without additional code changes?

grafnu (commented):
I know you didn't request a review yet, but two small things jumped out at me...

PR #475, opened 2022-09-29
Fix for broken pubber proxy device and corresponding integration test

grafnu:


bin/test_validator
```diff
@@ -15,13 +15,11 @@ shift
...
+proxy_id=AHU-22
 mkdir -p out
 
-serial_no=validator-$RANDOM
-echo Using pubber with serial $serial_no
-
-PUBBER_OUT=pubber.out
-VALIDATOR_OUT=validator.out
+PUBBER_OUT=out/pubber.out
+VALIDATOR_OUT=out/validator.out
```
johnrandolph, line 22: Eventually I wonder if we won't be able to use static filenames and need to do something more robust. e.g.

VALIDATOR_OUT=$(mktemp -p out/ validator.XXXX.out)

bin/test_validator
```diff
@@ -60,24 +62,36 @@ if [[ $i -eq $WAITING ]]; then
...
+    outfile=$PUBBER_OUT.$did
+    serial_no=validator-$RANDOM
+    echo Writing pubber output to $outfile, serial no $serial_no
+    cmd="bin/pubber $site_path $project_id $did $serial_no $@"
+    echo $cmd
+    $cmd > $outfile 2>&1 &
+
+    # Give a little bit of time to settle before deterministic check
+    sleep 10
+
```
johnrandolph, line 77: Should we check if the process is still alive after the sleep?

bin/test_validator
grafnu: I often find that actually makes it more difficult overall, because then when you (or a script) wants to automatically process the output you have to figure out which one-of-many files you're dealing with. There's also constraints about multiple-things running at the same time (as in, it won't work because of the UDMI-REFLECTOR registry behavior) -- so practically speaking it shouldn't be an issue in this case.

bin/test_validator
grafnu: The lines immediately following make sure that it started up ok (reaches quiescent state), and any check for did-it-die would be pretty arbitrary after 10 seconds... It's not a failure mode I've seen (successfully complete startup but have died within 10 seconds).  Sometimes I put a check in at the end of the script to say "hey, this died unexpectedly so probably invalid results"... but that's slightly different.

johnrandolph (approved):

PR #477, opened 2022-10-06
Create upversioning utility to upgrade schema and test messages version

noursaidi:
A tool to change the version in test messages/examples.

Introduces the `$id` annotation into schema files, and uses it so schema files are tagged with the version.

Uses a combination of a list of files to upgrade and a list of files to ignore, so any new files must explicitly be added into either at the time of running the utility.

Usage: `bin/upversion NEW_VERSION`


grafnu:
What's the functional intent behind it (the $id field you have) -- what is
it intended to be *used* for?  B/c if it has anything to do with
"versioning" specific to (e.g.) 1.4.0 I'd want it more explicit. An $id
would only be useful as a unique identifier, not with semantics. E.g., if
we replaced it with the sha512 hash of the string, would it still work for
its intended purpose?

noursaidi:
The intention behind the addition was to have some reference of the version of UDMI (e.g. `1.4.0`) when looking at the schema, because without doing anything git related, it's not possible to tell which version it relates to. This is trying to address some issues people have had.

Removing the semantic information from the value of `$id`, it would no longer meet this specific purpose.

grafnu:
Ok, in that case I'd suggest something like "$udmi_version" -- the reason
is I wouldn't want to have to parse the larger $id string to try and
extract the semantically meaningful value -- as it makes that too brittle.
So, if we know the thing we're looking for (i.e. UDMI version) then let's
just call that out explicitly!

bin/upversion
```diff
@@ -0,0 +1,92 @@
...
+while read file ; do
+    new_uri=$(printf $BASE_URI $NEW_VERSION $file)
+    jq ".[\"\$id\"] |= \"$new_uri\"" $file | sponge $file
+done < <( find schema -type f -name "*.json" )
+cat $YAY | xargs gsed -i -E "s/$TESTS_REGEX/\1\"$NEW_VERSION\"\3/g"
+bin/gencode_docs_examples 
+echo Done!\n
+
+echo Commit changes with:
+echo "git add tests/*.tests/*.json"
```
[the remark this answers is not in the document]
noursaidi: Originally did the tool did the tag and commit, but I ran into X reasons why that was a terrible idea so turned into echo. But they're removed now

schema/configuration_endpoint.json
```diff
@@ -1,4 +1,5 @@
 {
+  "$id": "https://raw.githubusercontent.com/faucetsdn/udmi/1.4.0/schema/configuration_endpoint.json",
```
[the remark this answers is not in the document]
noursaidi: As per discussion below changed to $udmi_version

etc/upversion_nay.txt
```diff
@@ -0,0 +1,25 @@
+tests/event_discovery.tests/empty.json
```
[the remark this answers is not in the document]
noursaidi: Comment incorporated

bin/upversion
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
noursaidi: Added -e

noursaidi:
CI Run [here](https://github.com/noursaidi/udmi/actions/runs/3330185987)

I can split the tool and the upversion to 1.4.0, but thought to leave it as an example of what exactly it does

bin/upversion
```diff
@@ -0,0 +1,155 @@
...
+#     all files must be added to etc/upversion.txt only files preceeded
+#     with a 'y' are updated. Only works with "pretty"/non-minified JSON
+#   - documentation in-line exammples
+#   - specific java files (Pubber.java, ConfigUtil.java, LocalDevice.java)
+
+UPVERSION_LIST=etc/upversion.txt
+TESTS_REGEX="^(\s{0,4}\"version\"\s*:\s*)([0-9.\"]*)(,?)"
+SCHEMA_VERSION_IDENTIFIER=\$udmi_version # must start with $
+
+# Updates existing $udmi_version in a JSON file. Called by
```
grafnu, line 21: NB: Typically avoid "called by" comments because all they do is get out of date and become inaccurate. It's easy enough to search for "update_existing_schema_version" so it's not hard to find where it's used...

bin/upversion
```diff
@@ -0,0 +1,155 @@
...
+# update_schema_version
+# Usage: update_existing_schema_version FILE_TO_UPDATE NEW_VERSION
+function update_existing_schema_version(){
+    file=$1
+    version=$2
+    gsed -i -E "s/\"\\$SCHEMA_VERSION_IDENTIFIER\"\s*:\s*\"[0-9.]*\"/\"$SCHEMA_VERSION_IDENTIFIER\": \"$version\"/" $file
+}
+
+# Updates an existing $udmi_version, or adds if missing
+# Usage: update_schema_version FILE_TO_UPDATE NEW_VERSIO
```
grafnu, line 31: Missing trailing N

bin/upversion
```diff
@@ -0,0 +1,155 @@
...
+if [[ "$#" != 1 ]]; then
+    echo Usage: $0 NEW_VERSION
+    exit 1
+fi
+NEW_VERSION=$1
+shift 1
+
+if ! [[ $NEW_VERSION =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+    echo ERROR Invalid version: $NEW_VERSION
+    # exit 1
```
grafnu, line 60: Remove comments for dead code. In the long run "# exit 1" isn't useful for anything...

bin/upversion
```diff
@@ -0,0 +1,155 @@
...
+fi
+
+bin/check_version
+if [[ -n $(git tag -l $NEW_VERSION) ]]; then
+    echo ERROR New version $NEW_VERSION already exists in upstream repo.
```
grafnu, line 65: if the system keeps going not sure these count as errors... so, either WARNINGS, or it should fail on these conditions... ?

bin/upversion
```diff
@@ -0,0 +1,155 @@
...
+for test_file in tests/*.tests/*.json; do
+    if [[ $(ggrep -P -c "^[^#] $test_file" $UPVERSION_LIST) != 1 ]]; then
+        echo $test_file
+        err_upversion_list=1
+    fi 
+done
+
+if [[ -n $err_upversion_list ]]; then
+    echo ERROR .. above files not found in upversion list or listed more than once
+    # exit 1
```
grafnu, line 84: what you can do is "touch .has_an_error" or something as a marker, and then at the end of the bunch of code, you just have one place that exits if that file exists.  (You can also just set a bash variable as long as you're careful of subshell scoping.)

bin/upversion
```diff
@@ -0,0 +1,155 @@
...
+#   - schema/*.json (all files)
+#   - etc/categories.json 
+#   - tests/**/*.json and validator/traces/**/*.out - 
+#     all files must be added to etc/upversion.txt only files preceeded
+#     with a 'y' are updated. Only works with "pretty"/non-minified JSON
+#   - documentation in-line exammples
+#   - specific java files (Pubber.java, ConfigUtil.java, LocalDevice.java)
+
+UPVERSION_LIST=etc/upversion.txt
+TESTS_REGEX="^(\s{0,4}\"version\"\s*:\s*)([0-9.\"]*)(,?)"
```
grafnu, line 18: can you use jq instead of this?  Seems very unstable... e.g. you can set a value with something like (from elsewhere in scripts):

bin/reset_config:    jq .system.testing.sequence_name=\"${config_file%.json}\" > ${dst_config}

etc/upversion.txt
```diff
@@ -0,0 +1,101 @@
...
+n tests/configuration_endpoint.tests/delta.json
+n tests/model_pointset.tests/example.json
+n tests/envelope.tests/lgtw.json
+n tests/envelope.tests/empty.json
+n tests/envelope.tests/errors2.json
+n tests/envelope.tests/example.json
+n tests/envelope.tests/gateway.json
+n tests/envelope.tests/example2.json
+n tests/envelope.tests/errors1.json
+n tests/state.tests/makemodel_upgrade.json #tests message upgrade from v 1
```
grafnu, line 30: space after #

Is the # even necessary, or just convenience?  What happens if there's other words after the file... ? Seems like having the # is better overall, just kinda curious... ?

noursaidi: Used it to improve readability, but is not necessary,  passed through bash `read action file comment` so everything after the second space is a comment

bin/upversion
noursaidi: Tried `jq` but opted for sed because:
- For files in test/*,  `jq` does not support comments, so cannot be used for these files.
- For files in schema/* `jq` 
    - adding a field puts it at the bottom, whereas it makes more sense for this to be at the top
    - reformats (although very little) content, namely enforces one item per line in lists.
For all changes, there are checks that there is one of whatever it is expecting to replace for sanity

bin/upversion
noursaidi: The `exit 1` shouldn't have been commented out,  so it should fail at each stage.

bin/upversion
noursaidi: all the `exit 1` shouldn't have been commented, they have been uncommented

bin/upversion
noursaidi: added

bin/upversion
noursaidi: Good suggestion, incorporated this for the errors which aren't for a duplicate version or unclean git

grafnu (approved):

PR #486, opened 2022-10-21
Fixes for continuous validation

grafnu:


noursaidi (approved):
fixed issues with validation (errors set to empty list rather than null and validation schema errors)

PR #488, opened 2022-10-26
deprecate legacy dashboard and rename "dashboard" to "udmis"

noursaidi:
Deprecates  (delete) legacy UDMI firebase dashboard and old "udms" (basically the DAQ-front end) directory, and rename "dashboard" to "udms", e.g. `udms/deploy_udms_gcloud PROJECT_ID`

Merge will be coordinated with learning material update


grafnu:
Just having a quick think of "UDMIS" vs. "UDMS" (should the "I" be included or not)... Right now, we're calling the other bit UDMIF (front-end), while this would be S for (service or system). What were your final thoughts about that part of the name?

noursaidi:
I had a likely imaginary memory that it was called "UDMS", and also we have "UDMS-REFLECT". But UDMI-S makes more sense and is consistent with "UDMI-F". I can make that change

noursaidi:
Renamed to udmis now

grafnu (approved):

PR #496, opened 2022-11-03
Update mapper output schema to allow multiple devices per message

gnmerritt:
This adds the additional required fields to the mapper output schema to allow for multiple translations to be issued at once (via a map of guid -> entity info) and adds an extra level to stash entity information like `type`, `missing_telemetry_fields`, and `connections`.


schema/event_mapping_entities.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$udmi_version": "1.4.0",
+  "title": "Mapping Event Entities",
+  "description": "Mapper output describing a group of devices by their guids",
+  "type": "object",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "additionalProperties": false,
+  "existingJavaType": "java.util.HashMap<String, MappingEventEntity>",
+  "patternProperties": {
+    "": {
```
grafnu, line 10: what are these keys?  There should be some format to them... "anything goes" isn't quite what we want.

schema/event_mapping_entity.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$udmi_version": "1.4.0",
+  "title": "Mapping Event Entity",
+  "type": "object",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "additionalProperties": false,
+  "properties": {
+    "code": {
```
grafnu, line 8: "code" isn't what I'd expect here... "name"?  "id"?    I wouldn't call something like "fcu-8_1" a code...

schema/event_mapping_entity.json
```diff
@@ -0,0 +1,30 @@
...
+      "description": "",
```
grafnu, line 9: empty description -- either have it described or leave the field out

schema/event_mapping_entity.json
```diff
@@ -0,0 +1,30 @@
...
+    },
+    "type": {
+      "description": "",
+      "type": "string"
+    },
+    "cloud_device_id": {
+      "description": "",
+      "type": "string"
+    },
+    "connections": {
```
grafnu, line 20: should not be empty

schema/event_mapping_entity.json
```diff
@@ -0,0 +1,30 @@
...
+    },
+    "links": {
```
grafnu, line 22: should not be empty

schema/event_mapping_entity.json
```diff
@@ -0,0 +1,30 @@
...
+    },
+    "translation": {
+      "$ref": "file:building_config.json"
+    },
+    "missing_telemetry_fields": {
```
grafnu, line 27: should not be empty

schema/event_mapping_entity.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$udmi_version": "1.4.0",
+  "title": "Mapping Event Entity",
+  "type": "object",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "additionalProperties": false,
+  "properties": {
+    "code": {
+      "description": "",
+      "type": "string"
```
grafnu, line 10: type should be an enum

schema/event_mapping_entity.json
```diff
@@ -0,0 +1,30 @@
...
+    "cloud_device_id": {
+      "description": "",
+      "type": "string"
+    },
+    "connections": {
+    },
+    "links": {
+    },
+    "translation": {
+      "$ref": "file:building_config.json"
```
grafnu, line 25: the referenced thing isn't quite a building config if it's under a "translation" bucket, maybe it was off before but we should clean it up as we go along.  These look like "translations" (plural), and then the thing under it should have the word "translation(s)" in it.

schema/event_mapping_entity.json
bsimmons-onboard: "code" is from the DBO definition "Code: The human readable identifier for the entity. This should be unique in document scope." https://github.com/google/digitalbuildings/blob/master/ontology/docs/building_config.md

@grafnu does this work or would you still prefer a change to something like "name", "id", or even "humanReadableId" ?

grafnu:
no worries, as long as it's not out of left field.  The usual use for
devices is either "device name" or "device id" -- so just slightly
different.  code works well enough.

schema/event_mapping_entity.json
gnmerritt: happy to fix this, but not sure I understand what the guidelines are. For example I see empty schema maps checked in at https://github.com/grafnu/udmi/blob/7755958bf6801557929a0955cf30b008813f1c49/schema/building_translation.json#L13 and in other places. 

Do the automated tests need to be updated to reflect this guideline? Am I missing or mis-reading the extant code?

schema/event_mapping_entity.json
gnmerritt: I don't believe so, this field will receive values like `"US-SVL-MP2"`

schema/event_mapping_entity.json
gnmerritt: Plural translations to me implies multiple different takes rather than a singular work - Lattimore's Illiad is a translation (singular). 

Not sure what you're asking re: the clean up, do you just want me to rename the `building_config.json` file?

schema/event_mapping_entities.json
grafnu: you can just use that one you see in a few places -- I just want something that's reasonably accurate in there!

schema/event_mapping_entity.json
grafnu: code is fine -- just wasn't sure where it came from!

schema/event_mapping_entity.json
grafnu: Those are just preliminary sketches of things, but we should figure out more what they are at this stage.

schema/event_mapping_entity.json
grafnu: hmmm, ok -- I think we'll have to follow-up with this team about this because that definition seems to be a mess... "US-SVL-MP2" is not a type.

bsimmons-onboard:
@grafnu we did a second pass based on your feedback, ready for another look.

gnmerritt:
Okay - added patterns in all the spots you requested, thank you for the suggestions @grafnu. The only one that I didn't use was `code` - I see a bunch of mixed case values in our examples and it's a human readable field so I think we should be fine on comparisons. 

I rebased and rebuilt the gencode hash but otherwise there are no changes in the earlier force-pushed commits

schema/event_mapping_entity.json
```diff
@@ -7,24 +7,41 @@
   "properties": {
     "code": {
       "description": "Human readable code for the entity, should be unique in the document scope",
-      "type": "string"
+      "type": "string",
+      "examples": ["fcu-8_1"]
```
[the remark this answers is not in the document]
gnmerritt: mixed case + `-` and `_` looks like it works

grafnu:
Mostly looks good, but there's still two lingering references to things as
just "object" when they can/should be more specific!

schema/event_mapping_entity.json
```diff
@@ -7,24 +7,41 @@
...
     },
     "cloud_device_id": {
       "description": "Opaque identifier for the entity",
-      "type": "string"
+      "type": "string",
+      "examples": ["3607425"]
     },
     "connections": {
+      "description": "Entity connections, keyed by guid",
+      "type": "object",
```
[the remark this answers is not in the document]
gnmerritt: I believe it is addressed, please see the `patternProperties` below on lines 29-31. This field is a map of the form `{ "guid": "CONTAINS" }` so it should be `type: object` as I understand it, with validation at the property level.

schema/event_mapping_entity.json
```diff
@@ -8,21 +8,27 @@
...
       "description": "Opaque identifier for the entity",
       "type": "string",
+      "pattern": "^[0-9]+$",
       "examples": ["3607425"]
     },
     "connections": {
       "description": "Entity connections, keyed by guid",
       "type": "object",
+      "patternProperties": {
+        "^[-0-9a-z]+$": { "enum": ["CONTAINS"] }
```
gnmerritt, line 30: the github threading is getting confused so I want to call this change out explicitly here, and I believe this is a link back to the previous thread: https://github.com/faucetsdn/udmi/pull/496/commits/8db9c3a40d6022b6ef1f34928e24886b5510aa48#r1040125597

schema/virtual_links.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$udmi_version": "1.4.0",
+  "title": "Virtual Equipment Links",
+  "description": "Virtual equipment mapping, keyed by guid",
+  "type": "object",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "additionalProperties": false,
+  "patternProperties": {
+    "^[-0-9a-zA-Z$]+$": {
+      "type": "object",
```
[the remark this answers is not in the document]
gnmerritt: yes - this is a two level map of the form `{ guid: { "dbo-type": "dbo-type" } }` so here I am using the pattern properties at the top level to validate the guid keys and then again on line 12 to validate the dbo types

gnmerritt (commented):
Hi Trevor - I tried to flag my changes explicitly in this review, hopefully that will make it easier for you/GitHub's UI

grafnu (approved):
Looks good! I'm sure there will be iterations at some point, but nothing is jumping out me as wrong at this point...

gnmerritt:
@grafnu I don't have write permissions on the repo - is there anything else that needs to happen before this merges?

grafnu:
Nothing to be done other than letting me know, that works!  Submitted...

PR #498, opened 2022-11-08
Split writeback tests

noursaidi:
Split writeback test into:
* `writeback_success_apply` can the device update the value of a point through writeback
* `writeback_success_state` can the device update the value of a point through writeback and sets the `value_state` to `applied`
* `writeback_invalid_state` does the device report `invalid` in a points `value_state` when the value is known to be invalid
* `writeback_failure_state` does the device report `failure` in a points `value_state` when the value is known to be invalid

Also add a utility `bin/test_sequencer_solo` to test a single sequence test against a pubber instance with given options (to test failures or edge cases .. similar to the one DAQ had). For some reason this works locally, [but in CI it goes through the first line (irrespective of how many lines the file has) .. haven't figured that bit out](https://github.com/noursaidi/udmi/actions/runs/3422535071/jobs/5700058776)


etc/sequencer_solo.out
```diff
@@ -0,0 +1,4 @@
+TESTING writeback_success_state with pubber opts: noWriteback
```
grafnu, line 1: I would turn this into an in/out file, that is also the input to the testing, e.g.:
```
TEST writeback_success_state noWriteback RESULT fail writeback_success_state timeout waiting for point filter_differential_pressure_setpoint to have value_state applied
```
as one line could automatically be parsed to extract the first part so it's not replicated in the test script

bin/test_sequencer_solo
```diff
@@ -0,0 +1,54 @@
...
+
+SITE_PATH=sites/udmi_site_model
+DEVICE_ID=AHU-1
+
+RESULTS_OUT=/tmp/sequencer_solo.out
+PUBBER_OUT=/tmp/pubber.log
+SEQUENCER_OUT=/tmp/sequencer.log
+TEST_LIST=/tmp/solo_sequencer_tests.txt
+
+cat > $TEST_LIST <<EOF
```
grafnu, line 14: as per other comment, I think this should be parsed out of the in/out file, not as a separate list maintained here

bin/test_sequencer_solo
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash -e
```
grafnu, line 1: I'd prefer a name like "test_itemized" or something to 1) avoid the test_sequencer prefix, and then 2) something a bit more descriptive than _solo.

etc/sequencer_solo.out
noursaidi: Changed as requested

bin/test_sequencer_solo
noursaidi: Changed to read the tests from the "golden file"

bin/test_sequencer_solo
noursaidi: renamed to test_itemized

bin/test_itemized
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash -e
+
+PROJECT_ID=$1
+shift 1
+
+REGEX="^TEST ([a-z_]+)([A-Za-z ]+)?(RESULT .*)?$"
```
grafnu, line 6: What's this used for? I don't see it anywhere in the file... ??

grafnu (approved):

PR #502, opened 2022-11-12
Endpoint topic prefix configuration

grafnu:
Required for the testing agnost other non-IoT-Core endpoints (e.g. Hono or AWS), that use a different topic structure.  Doesn't impact the entire topic (part including the event/config/state type), but just a prefix that includes (potentially among other things) the device_id.


johnrandolph (commented):
Can we please add test classes for all the classes?

johnrandolph (approved):

PR #503, opened 2022-11-14
Add shell exit traps for EXIT cleanup

grafnu:
Makes it so that ^C or a "kill SIGINT" against any of the processes exits cleanly. (There's the core java program, the java wrapper, and test runner.)


johnrandolph (approved):
Tried to find a way to make java call setsid() but seems not readily available? So I'm ok with this for now,.

PR #504, opened 2022-11-14
Enable metadata schema validation

grafnu:
Simple flag flip, and then an extra schema test file just for fun.

### DIFF
--- a/tests/metadata.tests/nosys.json
+++ b/tests/metadata.tests/nosys.json
@@ -1,0 +1,41 @@
+{
+  "version": "1.4.0",
+  "timestamp": "2018-08-26T21:39:29.364Z",
+  "description": "Generic test example metadata file",
+  "cloud": {
+    "auth_type": "ES256"
+  },
+  "pointset": {
+    "points": {
+      "return_air_temperature_sensor": {
+        "units": "Degrees-Celsius",
+        "baseline_value": 20,
+        "baseline_tolerance": 2
+      },
+      "room_setpoint": {
+        "writable": true,
+        "units": "Degrees-Celsius",
+        "baseline_value": 20,
+        "baseline_state": "applied"
+      }
+    }
+  },
+  "discovery": {
+    "families": {
+      "bacnet": {
+      }
+    }
+  },
+  "testing": {
+    "targets": {
+      "invalid": {
+        "target_point": "return_air_temperature_sensor",
+        "target_value": -20
+      },
+      "failure": {
+        "target_point": "room_setpoint",
+        "target_value": -20
+      }
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


noursaidi (approved):

bin/test_registrar
```diff
@@ -18,7 +18,7 @@ clean_devices=$(fgrep 'Z"' ${TEST_SITE}/registration_summary.json | wc -l)
 exit_status=0
 
 echo Found ${clean_devices} clean devices.
-[ "${clean_devices}" == 4 ] || exit_status=1
```
johnrandolph, line 21: Can we make this automatic?

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -301,23 +301,28 @@ private Metadata readMetadataWithValidation(boolean validate) {
       return null;
     }
 
+    JsonNode mergedMetadata = getMergedMetadata(instance);
```
johnrandolph, line 304: Add unit tests for this function please. It might need to be broken up into pieces?

bin/test_registrar
grafnu: What do you mean by automatic?

johnrandolph: Automatic = Can we derive the 4 from a calculation based on the test inputs

grafnu:
I'm really not sure it's worth it in this case. The logic to calculate that
for just this one case would be relatively complicated (find all the files
that don't match a very specific pattern that just matches one file), and
then likely would end up bring more fragile than just having the raw
expected value -- and then you'd also have to worry about an inadvertent
change in the input data changing the expected and actual value for the
test...

bin/test_registrar
johnrandolph: I'm semi ok with this if we make it less of a hunt to figure out why it's 3, instead of it just being a bare int by itself in a file that calculates the rest.  For example

# Clean devices are: SNS-4 GAT-123 AHU-22
[ "${clean_devices}" == 3 ].....

grafnu: I'm really not sure it's worth it in this case. The logic to calculate that
for just this one case would be relatively complicated (find all the files
that don't match a very specific pattern that just matches one file), and
then likely would end up bring more fragile than just having the raw
expected value -- and then you'd also have to worry about an inadvertent
change in the input data changing the expected and actual value for the
test...

grafnu: Ok -- added named constant with detail about why it is what it is!

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
grafnu: Added class LocalDeviceTest with coverage for the two major caes

grafnu:
Ok, added some tests for LocalDevice (for the new function), as well as comments at the high-level about clean devices.

PR #505, opened 2022-11-15
Utility to stream messages from a given device from Pub/Sub

noursaidi:
Feature desired by several people who are interested in "live streaming" data from a device.


pisuke (approved):
Great work! I tried it and it works very well. 
We might want to add the possibility to stream from all devices or a selection of devices with a more compact view of the stream in the future.

PR #506, opened 2022-11-15
Update stream_device_messages to add multiple devices

noursaidi:


noursaidi:
Draft because I realised the datetime is not UTC and need to fix/confirm

PR #509, opened 2022-11-16
fix events prefixed as null when writing traces with validator

noursaidi:


noursaidi:
@grafnu tagged you because I'm certain I ran into this problem before and then you fixed it, so I don't know I'm imaging or there was a change you made elsewhere which impacted this

grafnu (approved):

grafnu:
There's likely just multiple places that it needs to be caught/handled,
so... not surprising there's other cases out there!

PR #515, opened 2022-11-23
fix udmis deploy

noursaidi:


noursaidi:
Merging because fixes incorrect path name in deploy script

PR #516, opened 2022-11-23
Add system_mode_restart

johnrandolph:
Relates to fixes in Trevor's "myreset" branch


pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -575,12 +579,15 @@ private void maybeRestartSystem() {
     if (systemConfig == null) {
       return;
     }
+    info("maybeRestartSystem " + deviceState.system.mode + " " + systemConfig.mode
```
grafnu, line 582: I think this is likely intermediate debug info, so should be removed or downgraded to debug?

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/BlobsetSequences.java
```diff
@@ -99,4 +102,47 @@ public void endpoint_config_connection_success_reconnect() {
           && stateStatus.level == Level.NOTICE.value();
     });
   }
+
+  @Test
+  @Description("Restart and connect to same endpoint and expect it returns.")
+  public void endpoint_config_connection_success_restart() {
```
grafnu, line 108: this test seems misnamed because it has nothing to do with endpoint config -- it's purely restart. As per the discussion in the doc, we're really looking at two separate tests for "restart" -- one it just "does the system restart" and shouldn't be tied to anything to do with endpoint, and the other is "does endpoint redirect survive restart" -- which would be some extra logic on top of this one.  Anyway, I would suggest just renaming/redescribing this test to the simpler "restart only" one, and then we can add the other one later.

validator/src/main/java/com/google/daq/mqtt/util/ConfigDiffEngine.java
```diff
@@ -30,34 +30,44 @@ public ConfigDiffEngine() {
    */
   @SuppressWarnings("unchecked")
   public List<String> computeChanges(Config deviceConfig) {
-    Map<String, Object> updated = convertSemantics(deviceConfig);
+    Map<String, Object> updated = extractDefinitions(deviceConfig);
```
grafnu, line 33: What's our current stance with "unit testing" coverage?  I'd say that if we want to improve increasing our unit testing in general that now would be the time to do it (i.e. at least a >0 incremental improvement in test overage)... ?

pubber/src/main/java/daq/pubber/Pubber.java
johnrandolph: Agree, debug

validator/src/main/java/com/google/daq/mqtt/util/ConfigDiffEngine.java
johnrandolph: We should have tests as much as possible for new or changed code. AFAIK this change comes from a different change. Do you want to add tests in a different PR?

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/BlobsetSequences.java
johnrandolph: So rename it to "_restart_only" ?

grafnu:
At this point I think we're mixed up enough that it's not safe to assume
that something is coming from a different change -- if it's not in
faucetsdn/master then it doesn't exist and is essentially part of this
change. The PR you're thinking of was likely into _your_ working branch, so
doesn't have any permanence to it other than what you have. If we're going
to do it, I think tests should be added in the PR that introduces the
change, otherwise it just adds more work in the long run...  If you want to
shard out the work to others that's fine, but we should do it explicitly.

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/BlobsetSequences.java
johnrandolph: Yes I meant endpoint_config_connection_success_restart_only.

grafnu:
The ability to restart a system is completely independent/separate from
endpoint configuration so we need a test just for that -- irrespective of
any other endpoint-restart tests that we might have.

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/BlobsetSequences.java
johnrandolph: If it's system_mode_restart is there somewhere else to put the test?

PR #517, opened 2022-11-23
Merge various fixes into master

johnrandolph:
TBD


grafnu:
How is #517 different from #516?  They seem to have a lot of the same
content in them... ?

PR #518, opened 2022-11-23
sequencer tests for sample rate

noursaidi:
Sequence test for changable sample rate (sample_rate_sec) and changable publish interval (change both sample_rate and sample_limit)

Status for invalid sample rates will follow in a new PR


bin/test_itemized
```diff
@@ -14,6 +14,11 @@ GOLDEN_FILE=etc/test_itemized.out
 rm -f $RESULTS_OUT $SEQUENCER_OUT $PUBBER_OUT 
 touch $RESULTS_OUT $SEQUENCER_OUT $PUBBER_OUT 
 
+#pubber/bin/build
```
grafnu, line 17: don't leave dead code

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -858,19 +858,22 @@ private void configHandler(Config config) {
...
+      actualInterval = updatePointsetConfig(config.pointset);
+      updatePointsetPointsConfig(config.pointset);
       updateDiscoveryConfig(config.discovery);
       extractEndpointBlobConfig();
     } else {
       info(getTimestamp() + " defaulting empty config");
       actualInterval = DEFAULT_REPORT_SEC * 1000;
     }
+    if (configuration.options.fixedSampleRate != null) {
+      actualInterval = configuration.options.fixedSampleRate * 1000;
```
grafnu, line 875: a better coding style is to avoid mutable variables, and instead intoruce a new variable with the new maning, eith.

`useInterval = configuration.options.fixedSampleRate == null ? actualInterval : configuration.options.fixedSampleRate * 1000`

In the big picture it makes the code more readable because you don't need to keep track of when a variable might change. Mutable variables are only really necessary in loops, etc... (and this is why the variable was initially indicated as final, which forces that it's written ones, and exactly once)

validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
```diff
@@ -82,7 +82,7 @@ public abstract class SequenceBase {
   public static final String EVENT_PREFIX = "event_";
   public static final String SYSTEM_EVENT_MESSAGE_BASE = "event_system";
   public static final int CONFIG_UPDATE_DELAY_MS = 2000;
-  public static final int NORM_TIMEOUT_MS = 120 * 1000;
+  public static final int NORM_TIMEOUT_MS = 300 * 1000;
```
grafnu, line 85: Is this change still necessary?  I'm worried about this because it starts to drastically increase the overall time.  Is this because the new tests that you added have longer times to them but there's no way to individually specify an increased timeout?

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
```diff
@@ -32,7 +40,7 @@ public void system_last_update() {
     });
   }
 
-  @Test
+  @Test(timeout = 120000)
```
grafnu, line 43: why special-case the shorter time?

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
```diff
@@ -116,10 +129,151 @@ public void extra_config() {
...
+  private String samplingMessagesCheckMessage(SamplingRange samplingRange) {
+    return String.format("time period between successive pointset events is %s",
+        samplingRange);
+  }
+
+  /**
+   * Tests both sample_rate_sec and sample_limit_sec by defining two non-intersecting narrow
+   * ranges of both parameters, and ensuring telemetry is within this range.
+   */
+  @Test(timeout = 240000)
```
grafnu, line 204: is it possible to make the 240000 a NAMED_CONSTANT_MS?  I'm not sure syntactically if it's allowed, but would help readability overall!

validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
noursaidi: This test needed a longer time to run, but the global timeout overrides the test level timeout. So I increased the overall time, then put a timeout on every test to keep the time low

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
noursaidi: As above, increased the global timeout so I can have a longer test, and then reinstated the shorter timeout on all the other tests

bin/test_itemized
noursaidi: Accidental, was meant to be I commented, thanks

pubber/src/main/java/daq/pubber/Pubber.java
noursaidi: Thanks for the feedback, changed as recommended

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
noursaidi: Yes it works, added a class of constants e.g. `FOUR_MINUTES_MS`. Let me know that needs changing

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -859,19 +859,22 @@ private void configHandler(Config config) {
 
   private void processConfigUpdate(Config config) {
     final int actualInterval;
+    final int useInterval;
```
grafnu, line 862: can declare at point of use in this case since there is no branch involved.

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -1283,9 +1286,11 @@ private void updatePointConfig(AbstractPoint point, PointPointsetConfig pointCon
     updateState(point);
   }
 
-  private int updateSystemConfig(PointsetConfig pointsetConfig) {
+  private int updatePointsetConfig(PointsetConfig pointsetConfig) {
     final int actualInterval;
     boolean hasSampleRate = pointsetConfig != null && pointsetConfig.sample_rate_sec != null;
+    boolean hasSampleLimit = pointsetConfig != null && pointsetConfig.sample_limit_sec != null;
+    //if(hasSampleRate && hasSampleLimit && pointsetConfig.sample_rate_sec < )
```
grafnu, line 1293: ??

validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
grafnu: Well, I hate this logic but it's not your fault! It should be the other way around... default is shorter and then if you want longer it's a special case!

grafnu: (Nothing needs to change here, I'm just ranting)

pubber/src/main/java/daq/pubber/Pubber.java
noursaidi: Changed

pubber/src/main/java/daq/pubber/Pubber.java
noursaidi: accidentally committed, removed

grafnu (approved):

grafnu:
Closing for now since it seems stale. Please re-open when it's ready to address again.

PR #519, opened 2022-11-28
Various fixes to pubber and sequencer 

johnrandolph:


grafnu (approved):
Only thing missing would be unit tests for some of the classes (SemanticValue or ConfigDiffEngine) -- if we want to increase test coverage for these then I'd recommend doing it now as part of this PR.

PR #531, opened 2022-12-06
system_mode_restart

johnrandolph:


validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/BlobsetSequences.java
```diff
@@ -99,4 +102,47 @@ public void endpoint_config_connection_success_reconnect() {
...
+
+  @Test
+  @Description("Restart and connect to same endpoint and expect it returns.")
+  public void system_mode_restart() {
+    // Prepare for the restart.
+    final Date dateZero = new Date(0);
+    untilTrue("last_start is not zero", () -> deviceState.system.last_start.after(dateZero));
+
+    deviceConfig.system.mode = SystemMode.ACTIVE;
+    updateConfig();
```
grafnu, line 114: note upateConfig() is redundant here (won't hurt anything) because it's implicit in the untilTrue() call.

grafnu (approved):

PR #532, opened 2022-12-07
generateEndpointConfigBase64Payload needs to use generateEndpointConfigClientId

johnrandolph:
Note this means that pubber was potentially a false positive on restart?

### DIFF
--- a/validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/BlobsetSequences.java
+++ b/validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/BlobsetSequences.java
@@ -96,11 +96,11 @@ public class BlobsetSequences extends SequenceBase {
       BlobPhase phase = deviceState.blobset.blobs.get(
           SystemBlobsets.IOT_ENDPOINT_CONFIG.value()).phase;
       // Successful reconnect sends a state message with empty Entry.
-      Entry stateStatus = deviceState.blobset.blobs.get(
+      Entry blobStateStatus = deviceState.blobset.blobs.get(
           SystemBlobsets.IOT_ENDPOINT_CONFIG.value()).status;
       return phase != null
           && phase.equals(BlobPhase.FINAL)
-          && (stateStatus == null);
+          && blobStateStatus == null;
     });
   }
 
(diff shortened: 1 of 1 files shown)


validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/BlobsetSequences.java
```diff
@@ -47,7 +47,7 @@ private String generateEndpointConfigClientId() {
 
   private String generateEndpointConfigBase64Payload(String hostname) {
     String payload = String.format(
-        ENDPOINT_CONFIG_HOSTNAME_PAYLOAD, ENDPOINT_CONFIG_CLIENT_ID, hostname);
+        ENDPOINT_CONFIG_HOSTNAME_PAYLOAD, generateEndpointConfigClientId(), hostname);
```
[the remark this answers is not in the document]
johnrandolph: See below

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/BlobsetSequences.java
```diff
@@ -99,7 +99,8 @@ public void endpoint_config_connection_success_reconnect() {
       return phase != null
           && phase.equals(BlobPhase.FINAL)
           && stateStatus.category.equals(SYSTEM_CONFIG_APPLY)
-          && stateStatus.level == Level.NOTICE.value();
+          && stateStatus.level == Level.NOTICE.value()
```
grafnu, line 102: I meant that the status in the blob block was as intended. The system status is not always indicative of what the blob is doing... the specific problem here was that the blob was reporting an error that was not checked by this condition -- this change doesn't fix that.

johnrandolph: Changed this to check blob state message for success.

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/BlobsetSequences.java
```diff
@@ -92,14 +92,15 @@ public void endpoint_config_connection_success_reconnect() {
     deviceConfig.blobset.blobs = new HashMap<>();
     deviceConfig.blobset.blobs.put(SystemBlobsets.IOT_ENDPOINT_CONFIG.value(), config);
 
-    untilTrue("blobset entry config status is success", () -> {
+    untilTrue("blobset phase is FINAL and stateStatus is null", () -> {
       BlobPhase phase = deviceState.blobset.blobs.get(
           SystemBlobsets.IOT_ENDPOINT_CONFIG.value()).phase;
-      Entry stateStatus = deviceState.system.status;
+      // Successful reconnect sends a state message with empty Entry.
+      Entry stateStatus = deviceState.blobset.blobs.get(
```
grafnu, line 99: nit: probably blobStatus is a better variable name, to differentiate from the state-system-status

PR #537, opened 2022-12-09
Generate code coverage reports locally

johnrandolph:
Emits HTML which is immediately usable while coding, and XML.

The CodeCov plugin for Github can read the XML output (being generated here but not yet in the PR)

### DIFF
--- a/bin/gencovg
+++ b/bin/gencovg
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+function emitfileurl() {
+  local report="$1"
+  local path="$2"
+  local fullpath=$(realpath ${ROOT})
+  echo "Coverage Report"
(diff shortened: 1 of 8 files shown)


bin/gencovg
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
grafnu, line 1: -e?

bin/gencovg
```diff
@@ -0,0 +1,25 @@
...
+
+function emitfileurl() {
```
grafnu, line 3: emit_file_url or emitFileUrl?

johnrandolph: emit_file_url

grafnu (approved):

PR #543, opened 2022-12-17
Updates to schema components and some test improvements

grafnu:
1) Converts blob updates to be URL only (so data URLs for base64)
2) Adds a bad-hash test to verify sha256 for blobs
3) Cleans up some testing libraries to improve debugging
4) Cleans up some system metrics fields as per discussions with Delta


schema/config_blobset_blob.json
```diff
@@ -8,29 +8,20 @@
...
-      "format": "uri"
+      "type": "string"
     },
     "sha256": {
       "type": "string",
       "description": "Expected hash of the retrieved resource",
       "pattern": "^[0-9a-f]{64}$"
     },
-    "nonce": {
+    "generation": {
```
johnrandolph, line 19: How about "timestamp" which is less ambiguous and consistent to other format=date-time fields?

schema/configuration_endpoint.json
```diff
@@ -62,9 +66,11 @@
         }
       }
     },
-    "nonce": {
+    "generation": {
```
johnrandolph, line 69: How about "timestamp" which is less ambiguous and consistent to other format=date-time fields?

schema/configuration_endpoint.json
```diff
@@ -21,6 +21,10 @@
     "hostname": {
       "type": "string"
     },
+    "error": {
```
johnrandolph, line 24: What's this?

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -1017,14 +1036,27 @@ private String extractConfigBlob(String blobName) {
...
+      endpointConfiguration.error = e.toString();
+      return JsonUtil.stringify(endpointConfiguration);
+    }
+  }
+
+  private String acquireBlobData(String url, String sha256) {
+    if (!url.startsWith(DATA_URL_JSON_BASE64)) {
+      throw new RuntimeException("URL encoding not supported: " + url);
+    }
+    byte[] dataBytes = Base64.getDecoder().decode(url.substring(DATA_URL_JSON_BASE64.length()));
```
johnrandolph, line 1054: Tests

johnrandolph (commented):
Also sending you a doc

schema/state_blobset_blob.json
```diff
@@ -10,6 +10,12 @@
     },
     "status": {
       "$ref": "file:common.json#/definitions/entry"
+    },
```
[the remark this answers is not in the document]
grafnu: I'm not sure what this is referring to?  I think you might have this comment on a wrong line.

schema/config_blobset_blob.json
grafnu: The semantic meaning of "generation" is different than "timestamp" -- e.g. the value of the status message should match the generation of the config message -- it's NOT the timestamp of the status message/entry itself. The purpose of this is to tie the fields together, not indicating the time of the thing itself.  It's the same usage as the discovery subblocks use which have a similar semantic meaning.  We can sort through if that's the right behavior and/or term, but that's the intent behind it (and why the term is different).

schema/configuration_endpoint.json
grafnu: It's something for capturing errors during parsing/handling -- I updated the description slightly to reflect this.

schema/config_blobset_blob.json
johnrandolph: I created a doc so we can discuss this somewhere more nimble than here

schema/state_blobset_blob.json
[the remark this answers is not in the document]
johnrandolph: Yeah right below here, generation vs timestamp

bin/test_redirect
```diff
@@ -65,6 +65,7 @@ cat <<EOF > out/endpoint.json
 EOF
 
 base64=$(base64 -w 0 out/endpoint.json)
+sha256=$(cat out/endpoint.json | sha256sum | sed 's/ .*//')
```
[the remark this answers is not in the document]
grafnu: Generally we don't support OSX, we're just not setup for it -- so if people want to fix it we can, but it's not something I'm going to worry about too much.  That code you have seems to work so I'll put it in...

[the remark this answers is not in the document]
johnrandolph: I agree but if we can quickly chip away things

pubber/src/main/java/daq/pubber/Pubber.java
grafnu: unit tests added

grafnu:
Updated/addressed/discussed comments. PTAL.

bin/test_redirect
```diff
@@ -65,6 +65,7 @@ cat <<EOF > out/endpoint.json
 EOF
 
 base64=$(base64 -w 0 out/endpoint.json)
```
johnrandolph, line 67: Sigh this is similarly not OSX friendly. Can you comment OSX needs -b 0 for future archeologists?

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -152,13 +154,14 @@ DiscoveryEvent.class, getEventsSuffix("discovery")
...
   private final CountDownLatch configLatch = new CountDownLatch(1);
-  private final State deviceState = new State();
+  final State deviceState = new State();
   private final ExtraPointsetEvent devicePoints = new ExtraPointsetEvent();
   private final Set<AbstractPoint> allPoints = new HashSet<>();
   private final AtomicBoolean stateDirty = new AtomicBoolean();
   private final Semaphore stateLock = new Semaphore(1);
   private final String deviceId;
-  private Config deviceConfig = new Config();
+  private final List<Entry> logentries = new ArrayList<>();
```
johnrandolph, line 163: logEntries?

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -1448,12 +1482,12 @@ private String traceTimestamp(String messageBase) {
   private void augmentDeviceMessage(Object message) {
     try {
       Field version = message.getClass().getField("version");
-      assert version.get(message) == null;
       version.set(message, UDMI_VERSION);
       Field timestamp = message.getClass().getField("timestamp");
-      assert timestamp.get(message) == null;
-      timestamp.set(message, getCurrentTimestamp());
-    } catch (Exception e) {
+      if (timestamp.get(message) == null) {
```
johnrandolph, line 1487: Tests?

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -941,12 +945,29 @@ private void maybeRedirectEndpoint() {
       return; // No need to redirect anything!
     }
 
+    if (extractedEndpoint != null) {
```
johnrandolph, line 948: Tests?

schema/config_blobset_blob.json
johnrandolph, line 19: To be honest I'm still not a fan of this naming. The best case for it is that we already have it in "discovery", so it's consistent to have a field with this name that is a date-time value.  The other fields in the schema tell you what they are by their names, and you could reasonably understand the value. This naming gives you no such hint. "Timestamp" would be more clear. WDYT?

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -890,24 +893,25 @@ private void processConfigUpdate(Config config) {
...
     if (deviceConfig.blobset == null) {
       extractedEndpoint = null;
-      return;
+      return null;
     }
     try {
       String iotConfig = extractConfigBlob(IOT_ENDPOINT_CONFIG.value());
       extractedEndpoint = fromJsonString(iotConfig, EndpointConfiguration.class);
       if (extractedEndpoint != null) {
         // TODO: Refactor extractConfigBlob() to get any blob meta parameters like nonce.
```
johnrandolph, line 905: Update comment

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -1017,16 +1038,29 @@ private String extractConfigBlob(String blobName) {
...
+      endpointConfiguration.error = e.toString();
+      return JsonUtil.stringify(endpointConfiguration);
     }
   }
 
+  static String acquireBlobData(String url, String sha256) {
+    if (!url.startsWith(DATA_URL_JSON_BASE64)) {
+      throw new RuntimeException("URL encoding not supported: " + url);
+    }
+    byte[] dataBytes = Base64.getDecoder().decode(url.substring(DATA_URL_JSON_BASE64.length()));
```
johnrandolph, line 1056: Do we like how this will test if we send malformed data urls etc?

bin/test_redirect
grafnu: I really don't want to get into supporting OSX unless there's an official plan to do so -- e.g. proper QA and whatever behind it. It's just an unnecessary distraction for the core dev work we need to do.  If somebody wants to do OSX that's fine but then it needs to be on them to tweak the tools to work.

pubber/src/main/java/daq/pubber/Pubber.java
grafnu: added

pubber/src/main/java/daq/pubber/Pubber.java
grafnu: done

schema/config_blobset_blob.json
grafnu: but it's not just a timestamp -- the term "generation" has the semantics of grouping several things together (e.g. multiple state changes that are all part of the same overall operation), which is different than the basic timestamp of each individual state change. It also is a timestamp in that it can be used to generally sort things into rough categorization of when things happen, which is useful for debugging in a distributed system (which is not something then you can easily do with a random nonce).

pubber/src/main/java/daq/pubber/Pubber.java
grafnu: logentries is congruent with the use (where the full string is `deviceConfig.system.logentries` or something like that) -- I suppose it doesn't really matter, but I was just going with consistency at that level.

pubber/src/main/java/daq/pubber/Pubber.java
grafnu: Not sure what you mean -- this isn't a test. If there data URL is malformed then it will complain, either through the decoding, or the hash will not match.

schema/config_blobset_blob.json
johnrandolph: As we discussed in GVC, I agree the value may become a date-time format. I disagree it should be generation. 

The maximal improvement would be to name it and make it a timestamp, because the value will give you the human debug utility you request, and the name neither overly enhances it (since we aren't going to enforce any kind of grouping or data consistency across the "generation" fields) nor under-describes it (it's clear the field is a date/time).

If you'd like to proceed go ahead but I think the more general "timestamp" is optimal.

pubber/src/main/java/daq/pubber/Pubber.java
johnrandolph: (Is there a commit to push here?)

johnrandolph (approved):

schema/config_blobset_blob.json
grafnu: > and the name neither overly enhances it (since we aren't going to enforce any kind of grouping or data consistency across the "generation" fields)

But it does -- the "generation" will remain consistent across multiple phases of an operation. So, if there's "apply" and "final" they would have the same _generation_ but occur at different individual times (_timestamp_). This is why I don't want to call it "timestamp" since it has a semantically different meaning.

pubber/src/main/java/daq/pubber/Pubber.java
grafnu: added test

grafnu:
Updated with another test or two, and pushed the missed commit with the comment change.

schema/config_blobset_blob.json
johnrandolph: I don't think timestamp needs to change just because it's called timestamp, but I do think generation could also be "3rd" or (int)0 or "nonce". My suggestion is generation_timestamp since it retains what I understand to be both our highest level concerns.

johnrandolph (approved):

grafnu:
*generation_timestamp* is redundant -- it's Hungarian notation that
generally goes against style guides.  The *type* of the field is a
*timestamp*, so it's not really necessary to add that in again and those
other values you cite can't be used (by design).  In java it manifests as a
Date() object so there's no ambiguity there.

johnrandolph:
it's nothing that fancy (Hungarian notation), rather my suggestion merges our topmost concerns and is consistent with the rest of the blob. We have neither extremely high level names e.g. "data_validity_check" // sha256, nor hungarian "sha256_string".  That said I have approved either yours or my rename.

PR #545, opened 2022-12-20
Refactor system operations config/state bucket

grafnu:
1) Refactoring of the system operations config/state bucket to provide more semantic structure.
2) Add cloud functions mechanism to auto-detect when an upgrade is necessary.

THIS PR WILL REQUIRE UPDATING CLOUD FUNCTIONS (but should be backwards compatible).


validator/src/main/java/com/google/daq/mqtt/util/MessageUpgrader.java
```diff
@@ -48,30 +48,53 @@ public MessageUpgrader(String schemaName, JsonNode message) {
...
+    final JsonNode original = message.deepCopy();
+    boolean upgraded = false;
+
+    if (forceUpgrade || minor < 0) {
+      minor = 0;
+      patch = 0;
+      upgraded = true;
+    }
+
     if (minor < 3) {
```
noursaidi, line 69: if (major == 1 && minor < 3)

validator/src/main/java/com/google/daq/mqtt/util/MessageUpgrader.java
```diff
@@ -48,30 +48,53 @@ public MessageUpgrader(String schemaName, JsonNode message) {
...
+
     if (minor == 3 && patch < 14) {
+      JsonNode before = message.deepCopy();
       upgrade_1_3_14();
+      upgraded |= !before.equals(message);
+      patch = 14;
     }
-    if (message.has(VERSION_PROPERTY_KEY)) {
+
+    if (minor < 4) {
```
noursaidi, line 81: I think  the new operational block needs to be in a new version `1.4.1`, and the upgrade is then from 1.4.0 to 1.4.1?

grafnu:
As far as I can tell, 1.4.0 was never released.

validator/src/main/java/com/google/daq/mqtt/util/MessageUpgrader.java
grafnu: Major is always 1 (assertion at top of code block)

PR #546, opened 2022-12-21
Adding basic pylint

grafnu:


bin/test_pylint
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash -e
+#
+# Run pylint on all the detected python files and compare against the expected list.
+# This is far from perfect, but the goal is just to prevent horrible things and to get
+# people to at least think about the style of python code. Adding a new python file will
+# require people to add another line in the test_pylint.out golden file.
```
noursaidi, line 6: There are only a handful of lint errors - it's probably worth marking them as ignored  in the file or resolving them to save the additional step of adding the file to the

bin/test_pylint
```diff
@@ -0,0 +1,31 @@
...
+source venv/bin/activate
+
+LINT_OUT=out/test_pylint.out
+LINT_GOLDEN=etc/${LINT_OUT#out/}
+
+rm -f $LINT_OUT
+mkdir -p out
+
+for file in $files; do
+    linty=$(pylint $file | fgrep "rated at" | awk '{print $7}' | sed -e 's/\..*//')
```
noursaidi, line 26: Can we include the Google pylintrc somewhere as the configuration file? 
https://google.github.io/styleguide/pylintrc

grafnu:
Can you try that again -- I don't understand what you mean. There are lots
of lint errors in the current files.  Marking them as ignored or resolving
them is certainly the right thing to do, but not sure what that has to do
with this PR... ?  Do you mean "done include files that have no lint errors
in the list"?

noursaidi:
If the current lint errors are resolved or ignored, then there is no need for the golden `test_pylint.txt`  and the need to add any python files into the list. The CI check becomes for any lint errors. 

I have resolved most pylint errors, can PR into this?

grafnu:
Sure -- if you have something to update I can take a look.  Right now I'm
seeing LOTS of python lint errors, so not quite sure what you're really
getting at -- so looking at the same version of things would likely help.

johnrandolph (commented):
Please add to bin/precheck

noursaidi:
https://github.com/grafnu/udmi/pull/4 resolves the pylint errors (combination of actually fixing, ignoring errors with good reason, or skipping (`bin/test_py`).

Also suggest the pylintrc is named `.pylintrc` in the root directory, then it's picked up automatically by pylint or an IDE on the users machine

noursaidi (approved):

PR #560, opened 2023-01-13
Refactor BlobsetSequences to add endpoint_redirect_and_restart

johnrandolph:


johnrandolph:
This PR suggests that we stop comparing these sequence markdown files. 

They are problematic to keep in sync while we have a different environment from generation (local) to running the tests.

PR #562, opened 2023-01-19
Add site_metadata schema

noursaidi:
Interactive viewer - https://noursaidi.github.io/udmi/gencode/docs/site_metadata.html 

Add site_metada schema for collecting information about a site. Renamed old site_metadata into site_defaults. (this change needs to be done in the faucetsdn/udmi_site_model repo too)


grafnu:
You'll notice that this is failing unit tests since there's a unit test that runs registrar on a test site, see tests/downgrade.site -- updating that will make it clear what the proposed change is too, since it's hard to tell now since there is no "example" to go along with the code changes...

schema/site_metadata.json
```diff
@@ -0,0 +1,42 @@
+{
+    "$udmi_version": "1.4.0",
+    "title": "Site Metadata",
+    "description": "Information about the ",
```
grafnu, line 4: about the _______

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
```diff
@@ -718,14 +718,14 @@ private void loadSiteMetadata() {
...
       return;
     } catch (Exception e) {
-      throw new RuntimeException("While validating " + SITE_METADATA_JSON, e);
+      throw new RuntimeException("While validating " + SITE_DEFAULTS_JSON, e);
     }
 
     try {
-      System.err.printf("Loading " + SITE_METADATA_JSON + "\n");
+      System.err.printf("Loading " + SITE_DEFAULTS_JSON + "\n");
       this.siteMetadata = OBJECT_MAPPER.readValue(siteMetadataFile, Metadata.class);
```
grafnu, line 726: so -- if we're changing the schema then we need to change the class it's using as well... So rather than loading this into Metadata.class it should load it into SiteMetadata.class

grafnu:
Initially the design behind the site_defaults.json file was a way to specify default values for any device -- and the default value could be _anything_. With this change, the default values that can be supplied are only things that are defined in the SiteMetadata schema.  Is that the intended behavior?  Just want it to be explicit to make sure we're thinking the same thing.  Then, also, it means there might be a bit of extra work in the code (literally could be a single line, though) to add the site SiteMetadata to the defaults of the device -- not something I think is there currently...

grafnu:
What's the status of this?  Either we should update, finalize, or obliviate.

PR #563, opened 2023-01-19
add serial_no to metadata

noursaidi:


grafnu (approved):

grafnu:
how is this used?  If we put it in, then we should make the tests actually test it?

PR #566, opened 2023-01-25
Add @Feature annotation with categories

grafnu:


validator/src/main/java/com/google/daq/mqtt/sequencer/Feature.java
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Defines the feature score for this test, in AU.
+   *
+   * @return feature score value
+   */
+  int score() default DEFAULT_SCORE;
+
+  /**
+   * Enum of allowed stages.
+   */
```
johnrandolph, line 48: Is there a doc that pinpoints what each of these stages means to the UDMI project? For example ALPHA vs PREVIEW

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/BlobsetSequences.java
```diff
@@ -198,7 +198,7 @@ public void endpoint_connection_success_alternate() {
 
   @Test
   @Description("Restart and connect to same endpoint and expect it returns.")
-  @FeatureStage(PREVIEW)
+  @Feature(stage = PREVIEW)
```
johnrandolph, line 201: DO we need category?

validator/src/main/java/com/google/daq/mqtt/sequencer/Feature.java
noursaidi: Also occurs to me, should "REQUIRED" be renamed "RELEASE"?

grafnu: I'd like to scope this PR at the mechanism for specifying the properties, but not the properties themselves (leave changing that to a separate PR). Once this PR is in place, it's easy to add/change the values themselves, and/or the properties associated with any given feature.

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/BlobsetSequences.java
grafnu: Right now, if no category is defined then it just defaults to "unknown."  Basically it comes down to the behavior we want for a test that has no particular category defined.  Either we say 1) it has to have a category, 2) for "no category" then we leave out the @Feature annotation, or 3) defaults to something.  Doesn't really matter to me what behavior it has - it's all relatively easy to change.  The only enforcement there is right now is that the end result has to have A category in it, but it's undetermined exactly how/when that is specified in all cases.

johnrandolph: I meant should we fill in category values that Nour is generating. But we can revisit this when we settle the enums, too.

validator/src/main/java/com/google/daq/mqtt/sequencer/Feature.java
johnrandolph: If you change the enum to just one placeholder value I'm good with that. Then we can go back and edit the real values in place.

grafnu:
Eventually yes, but I'd like to keep that as a separate activity.  After
this PR, I want to have some generated NAMED_CONSTANTS for the categories
to prevent string typos, etc...  Similar to the error categories! All just
staging steps to the eventual goal...

validator/src/main/java/com/google/daq/mqtt/sequencer/Feature.java
grafnu: How about two -- I'd like to keep default and non-default since the code/output is different in those two cases!  Updated, PTAL.

johnrandolph: Ok...as long as we can change REQUIRED to a different label if needed lgtm

johnrandolph (approved):

PR #574, opened 2023-01-30
Adding feature bucket documentation and generated enum

grafnu:
Also renames the code use of "category" to "buckets" to avoid confusion with the already existing error categories.


bin/gencode_buckets
```diff
@@ -0,0 +1,85 @@
...
+WILDCARD_SUFFIX = r'.[a-z]+(_[a-z]+)*'
+
+BUCKET_REGEX = r' *\* _([a-z]+)_: (.*)'
+JSON_FORMAT = '%s%s{ "pattern": "^%s$" }'
+DEVICE_PREFIX = 'device'
+
+JAVA_DESCRIPTION = '\n%s// %s\n'
+JAVA_TARGET = '%s%s("%s")%s\n'
+
+doc_in = os.path.join('docs/specs/buckets.md')
```
johnrandolph, line 18: Make doc_in, java_in, java_out CONSTANTS, or, move them to main() and pass them as arguments to the functions (better for testing too). Prefer latter

Also, not sure on the need for os.path.join() ?

bin/gencode_buckets
```diff
@@ -0,0 +1,85 @@
...
+def read_buckets():
+  buckets = []
+  prefix = []
+  previous = -1
+  group = None
+  with open(doc_in, 'r', encoding='utf-8') as doc:
+    while line := doc.readline():
+      indent = line.find(BUCKET_MARKER)//2
+      wildcard = re.match(WILDCARD_REGEX, line)
+      if wildcard:
```
johnrandolph, line 31: Where is this used?

johnrandolph (commented):
Since this script will write java code,

bin/gencode_buckets
grafnu: It's not -- good catch (removed).

bin/gencode_buckets
grafnu: Did both actually (since they're CONSTANTS that should be passed in). And removed os.path.join()

johnrandolph (approved):

PR #577, opened 2023-02-02
Skip ALPHA tests unless explicitly specified

grafnu:


validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceRunner.java
```diff
@@ -167,4 +165,35 @@ private void process(List<String> targetMethods) {
...
+      return true;
+    }
+    Feature annotation = method.getAnnotation(Feature.class);
+    Stage stage = annotation == null ? Feature.DEFAULT_STAGE : annotation.stage();
+    return stage.processGiven(getDesiredStability());
+  }
+
+  private Stage getDesiredStability() {
+    String stability = SequenceBase.validatorConfig.stability;
+    return Strings.isNullOrEmpty(stability) ? DEFAULT_STABILITY : Stage.valueOf(stability);
```
johnrandolph, line 193: This stage vs stability naming does not seem congruent to me. 
I think we should pick one term, and it should be related to the enum decl. 
I think stage is better. We are choosing to test at a min or max threshold or level of stage. Alpha tests are in a stage of alpha. They might be quite stable, just preliminary. 
WDYT

johnrandolph (commented):
The impact of only generating sequence.md for tests that pass a certain STAGE threshold, instead of just generating them all, is potentially problematic IMHO.

1) If an eng changes the STAGE level/threshold that sequencer runs test at--perhaps even multiple times in a row--then some different sequence.md files are generated. 
  1a) The eng may end up with only file exists/does not exist difference from whatever upstream you're comparing to.  It would not be possible to compare the *contents* of the file which I believe is a feature request of this sequence.md generation.
 1b) It generates additional noise in the client

I think we may just need to generate them all. WDYT?

grafnu:
My idea of restricting the output to "only non-Alpha tests" was two-fold:
1) Reduce the churn for developers (aka us) for Alpha features.  So, if
you're working on something as Alpha then you don't really want to be
bothered with the generated markdown being exact or changing. It's still
there if you want to look at it, but it wouldn't trigger any of the
automatic-change mechanisms that were giving problems.
2) Make it clear to external people about what was "solid" in terms of what
they might want to consider implementing (BETA/STABLE)

I don't understand the concern about the generated files. Yes, the system
only generates sequence.md files for tests that are run.  The contents of
the files are only compared for BETA+ tests, so it doesn't matter if the
ALPHA tests run or not -- it won't trigger the CI comparisons. The files
are only compared by test_sequencer, so just running the sequence test
doesn't do anything specific with the generated files. Can you be more
specific about the particular error case you're referring to?   I get the
part about "switch from running normal BETA+ to running ALPHA+" -- but not
what the impact of that would be -- it shouldn't make any difference to the
flow.

"I think we may just need to generate them all. WDYT?" -- depends what you
mean by "generate" -- they are *generated *for every test that actually is
run (this is true now). This PR is adding that they are only *combined* into
the master docs file if they are BETA+.

What client and noise are you talking about?

validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceRunner.java
grafnu: How about "feature_stage" -- same as Feature.Stage (as a class). I panicked and put "stability" because I didn't just want to use "stage" (too generic in the wider context) -- but I agree that's wasn't congruent.

johnrandolph: Relating to stage would be great. What about min_stage or something? I'm actually not wildly happy about min_ but it at least implies we are filtering from a range, not matching.

johnrandolph (approved):

grafnu:
I'm updating the stage level to "min_stage" which is parallel to e.g. how
logging handles it  (min_ prefix).  Not perfect, but I think it gets the
meaning clear enough.

PR #585, opened 2023-02-13
Feature Enumeration

grafnu:


tests/event_discovery.tests/features.json
```diff
@@ -7,12 +7,25 @@
   "timestamp": "2018-08-26T21:42:12.237Z",
   "generation": "2018-08-26T21:37:12Z",
   "features": {
-    "system": {
-      "baseline": {
-        "last_config": {}
+    "endpoint": {
+      "features": {
+        "config": {
+          "stage": "alpha"
```
noursaidi, line 13: Is this the stage of the device's implementation of this feature? In the case that all the tests under endpoint.config were stable, and/or endpoint.config itself was "stable" (if we assign stages to buckets), a device can still be "alpha"?

grafnu:
Yes, basically it's saying that the device believes that it implements this
feature to the indicated level (which would determine if it should be
qualified, or used in a production building). So basically, something
like @Alpha == "implemented but ignored", @Beta == "should be run through
official qualification but not used in production buildings'', and
then @Stable == "Can be used in production buildings'' -- essentially the
base "meaning" of alpha/beta/stable are the same, but the application is
slightly different (tests are not devices).

"all the tests under endpoint.config were stable" isn't directly relevant
here.  The device needs to pass the tests that are @Stable, not all of them
-- and if that's sufficient for the feature to be used then it's GA.

noursaidi (approved):

PR #586, opened 2023-02-14
Adding provision to clean out alternate config

grafnu:
Nukes the not-in-use config to prevent it inadvertently being used to bounce around from a not-used endpoint.

### DIFF
--- a/validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
+++ b/validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
@@ -42,7 +42,6 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
-import java.nio.file.Files;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
```diff
@@ -1242,8 +1240,19 @@ protected void withAlternateClient(Runnable evaluator) {
     }
   }
 
-  protected void mirrorDeviceConfig() {
-    String receivedConfig = actualize(stringify(receivedUpdates.get(CONFIG_SUBTYPE)));
+  protected void mirrorToOtherConfig() {
```
johnrandolph, line 1243: Could we have some line by line comments here? It's getting fairly dense...

validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
```diff
@@ -1242,8 +1240,19 @@ protected void withAlternateClient(Runnable evaluator) {
...
+    updateConfig("mirroring config " + useAlternateClient);
+    Config target = (Config) receivedUpdates.get(CONFIG_SUBTYPE);
+    // Since this is updating the mirror, use the opposite of what it would be locally.
+    target.system.testing.endpoint_type = useAlternateClient ? null : "alternate";
+    updateMirrorConfig(actualize(stringify(target)));
+  }
+
+  protected void clearOtherConfig() {
+    updateMirrorConfig("{}");
```
johnrandolph, line 1252: Comments please. Should this run in all conditions or are there any useful assertions to add that help document what's going on.

johnrandolph (changes requested):

grafnu:
Updated with comments.

johnrandolph (approved):

PR #595, opened 2023-02-22
Add proper UDMIS reflect envelope, and clean up startup configuration check

grafnu:


grafnu:
Merged without approval since it was blocking some other immediate work. If there's any comments/issues to be addressed I can do them in a new PR.

PR #596, opened 2023-02-22
Add feature stages and buckets to existing tests

noursaidi:


noursaidi:
Can't get the CI Tests to pass, suspect because of ongoing changes

grafnu:
Now now, don't go blaming the system for things too hastily!  I pulled your
PR into my repo and it works just fine:
https://github.com/faucetsdn/udmi/actions/runs/4251328598

I think the problem is likely that your functions aren't adequately
updated. As per b/270581316 there's some known problems with it, but...
basic answer is "try updating your cloud functions"

grafnu (approved):

PR #599, opened 2023-02-27
Fix processing for hidden or temporary device files

grafnu:
Addresses situations where there's a hidden folder in the devices/ directory (e.g. .DS_store from mac) and/or temporary edit files (i.e. trailing ~). Filters them out so they're not considered valid "devices" for the tools.


noursaidi (approved):

johnrandolph (changes requested):
Can you please describe the scenario this fixes in the comments of this PR?
Also the generic variable name...

PR #605, opened 2023-03-02
update instruction for udmi_reflect

noursaidi:


noursaidi:
merge because documentation mistake correction

PR #606, opened 2023-03-02
Add Feature Stages docs

johnrandolph:


docs/guides/feature_stages.md
```diff
@@ -0,0 +1,39 @@
...
+# Feature Stages
+
+Sequence tests use the @Feature annotation to tag test functions with their stage of release.
+
+* ALPHA
+    * Definition
+        * The feature (definition, code, schema, tests, etc) is still expected to change, maybe even substantially, while in ALPHA stage.
+        * The feature may not work as expected.
+        * Ready for external testing by manufacturers who are trusted testers/we have a close relationship with.
+        * One CI test needs to pass.
```
grafnu, line 12: What does this mean?  What is "one CI test" that's not the same as an @Feature?   (i.e. what information is lost without this?)

docs/guides/feature_stages.md
```diff
@@ -0,0 +1,39 @@
...
+* BETA
+    * Definition
+        * Development team has finished initial development of feature.
+        * The feature is defined in device qualification plans.
```
grafnu, line 16: As per my comment in the doc the UDMI specs can't reference qualification -- the dependency goes the other way.

docs/guides/feature_stages.md
```diff
@@ -0,0 +1,39 @@
...
+        * Ready for external testing by many manufacturers.
+        * The feature will only change if significant bugs or feature mismatch is found from the industry, partners, or testers.  (Note that feature mismatch should be a very rare occurrence if the feature was fully vetted in the ALPHA stage.)
+* STABLE
+    * Definition
+        * Feature is complete, and should not change.
```
grafnu, line 21: I think the qualifier here is that it's been known stable with n>=1 devices in a live environment, so this is really about "ready for wider adoption"

bin/sequencer
```diff
@@ -7,7 +7,8 @@
 ROOT_DIR=$(realpath $(dirname $0)/..)
 TEST_LOG=/tmp/test_log.txt
 log_level=INFO
-UDMI_ALT_REGISTRY=ZZ-REDIRECT-NA
+UDMI_ALT_REGISTRY=ZZ-TRI-FECTA-ALT-REGISTRY
```
grafnu, line 10: unintended change?

johnrandolph:
I'll move these comments back over to the original doc.

PR #607, opened 2023-03-03
Adding structured report output to sequencer

grafnu:
Result ends up in (when doing bin/test_sequencer): sites/udmi_site_model/out/devices/AHU-1/sequencer_report.json


noursaidi:
My only question is where the feature stage (alpha/beta/stable) is recorded?

noursaidi:
And also, should the descriptions be in there too? I'm thinking there either needs to be some file with all the available sequences and their stage and descriptions for every tagged version that a consuming application can read for this information, or they're included in the results?

noursaidi (approved):

noursaidi (approved):

PR #611, opened 2023-03-07
Upversion files to 1.4.1

noursaidi:


bin/upgrade_version
```diff
@@ -17,7 +17,7 @@
 UPVERSION_LIST=etc/upversion.txt
 TESTS_REGEX="^(\s{0,4}\"version\"\s*:\s*)([0-9.\"]*)(,?)"
 SCHEMA_VERSION_IDENTIFIER=\$udmi_version # must start with $
-ERROR_MARKER=.upversion_error
```
grafnu, line 20: does this change actually do anything?   I have no problem with it -- more just a curiosity

grafnu (approved):

bin/upgrade_version
noursaidi: Nope, was testing to see if it did and just left it

PR #618, opened 2023-03-13
Enabling baseline registry sharding for decreasing test run time.

grafnu:


docs/tools/sharding.md
```diff
@@ -0,0 +1,48 @@
...
+
+# Sharding Sequencer
+
+It's possible to run sequencer on an alternate set of project registries, called shards. This is
+useful for testing things from the same site model in parallel, since currently there can only
+be one instance of any tool using the reflector running on each registry.
+
+## Operation
+
+Basic operation is determined by a specified _registry suffix_ (e.g. `%A`) that triggers all (ideally) tools
```
johnrandolph, line 11: %A reads tokenization will occur and be replaced with something like a shard number or something. How about an example  like "-SUFFIX" or "-A"?

docs/tools/sharding.md
```diff
@@ -0,0 +1,48 @@
+[**UDMI**](../../) / [**Docs**](../) / [**Tools**](./) / [Sharding](#)
+
+# Sharding Sequencer
+
+It's possible to run sequencer on an alternate set of project registries, called shards. This is
```
johnrandolph, line 5: If I'm understanding correctly there is one alternative registry/shard. It would be great if we had many shards, not just one, as long as we are here. What do you think?

docs/tools/sharding.md
grafnu: The choices are `+ . % - _ ~` but I want to avoid `-` (dash) since that's what's used as the allowable separator in the normal site/registry name, so I just want something different. I changed it to `_` (underscore) since it's visually obvious, or I can do anything else you prefer from `+ . ~` (the remaining three options).

docs/tools/sharding.md
grafnu: Yes (current setup), and yes (have many). It's why I called everything a "shard" rather than just suffix or something. But, I don't have the capacity to do that right now (it's not trivial) so not something I can prioritize and work into this PR.

johnrandolph (approved):

PR #644, opened 2023-05-11
Add example HAProxy with logging

noursaidi:


misc/proxy/haproxy/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
grafnu, line 1: nb: the .sh extension on the filename is unnecessary/redundant. Ican just be "build" as it's an executable file, and it doesn't matter if internally it's shell or python, etc...!

grafnu (approved):

grafnu:
What's the status of this?  It's been here for a while, so we should either finalize or close it for now...

PR #648, opened 2023-05-19
Validator fix for points in state messages

grafnu:


grafnu:
bump. @noursaidi would like to get this closed out soonish -- PTAL.

PR #652, opened 2023-05-25
Update for feature enumeration BETA

grafnu:


validator/sequences/feature_enumeration/sequence.md
```diff
@@ -11,5 +11,6 @@
     * Remove `discovery.generation`
 1. Wait for cleared enumeration generation
 1. Check that no family enumeration
-1. Check that feature enumeration feature is stable
+1. Check that feature enumeration matches metadata
```
noursaidi, line 14: Should the "match metadata" be a seperate test? I'm thinking about the process here:
1) feature enumeration
2) validate using sequencer that the device supports each feature it says it supports
3) put results into metadata
4) run test again?

validator/sequences/feature_enumeration/sequence.md
```diff
@@ -1,5 +1,5 @@
 
-## feature_enumeration (ALPHA)
+## feature_enumeration (BETA)
```
noursaidi, line 2: Our current criteria for BETA says a manufacturer must have implemented it. Has any yet? Maybe the requirement should be flexible

tests/event_discovery.tests/features.json
```diff
@@ -3,29 +3,17 @@
...
   "generation": "2018-08-26T21:37:12Z",
   "features": {
-    "endpoint": {
-      "features": {
-        "config": {
-          "stage": "alpha"
-        }
-      }
+    "endpoint.config": {
+      "stage": "alpha"
```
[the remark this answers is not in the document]
grafnu: This indicates the _device_ level of support for the feature.  So, the device might indicate that it supports a feature but only at "alpha" level meaning you can technically use it, but shouldn't use it in production, etc...  Alpha/Beta/Stable are very generic terms so they can't really be limited to only one specific use. Yes, the individual sequencer tests also have this designation, and the semantics are the same, but the scope is different.

tests/event_discovery.tests/features.json
```diff
@@ -3,29 +3,17 @@
...
   "timestamp": "2018-08-26T21:42:12.237Z",
   "generation": "2018-08-26T21:37:12Z",
   "features": {
-    "endpoint": {
-      "features": {
-        "config": {
-          "stage": "alpha"
-        }
-      }
+    "endpoint.config": {
```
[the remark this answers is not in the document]
grafnu: If it's alpha then ¯\_(ツ)_/¯, if it's BETA/STABLE then we shouldn't rename it.  Generally there should be provisions to ignore anything "alpha" so a misnamed bucket shouldn't cause any systemic problems.

docs/specs/sequences/generated.md
```diff
@@ -27,9 +27,26 @@ Some caveats:
...
+* [feature_enumeration](#feature_enumeration)
 * [system_last_update](#system_last_update): Check that last_update state is correctly set in response to a config update.
 
+## empty_enumeration (BETA)
+
+1. Update config before enumeration not active:
+    * Add `discovery` = { "enumerate": {  } }
+1. Wait for enumeration not active
+1. Update config before matching enumeration generation:
+    * Add `discovery.generation` = `generation start time`
```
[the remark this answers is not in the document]
grafnu: discovery is the superset of { scan, enumeration }, as [roughly described in the docs](https://faucetsdn.github.io/udmi/docs/specs/discovery.html).

validator/sequences/feature_enumeration/sequence.md
grafnu: I'm not sure what you're getting at with your process... the metadata should indicate what features it supports, or you can give it the -a flag if you don't know yet. This test is specifically that the feature discovery enumeration returns the right set of things that you think you're testing against.  Or you run the feature enumeration test, it fails, and then you capture that data. I'm not sure why anything in there would need to be a separate test?

validator/sequences/feature_enumeration/sequence.md
grafnu: Yah -- I was thinking about this.  One option might be to have a "PREVIEW" stage or something that says "this is not Alpha anymore but not quite BETA" -- which would be that interstitial phase of "we're working with somebody on this". So the progression would be ALPHA -> PREVIEW -> BETA -> STABLE... thoughts?  Easy enough to add in to the code!

validator/sequences/feature_enumeration/sequence.md
noursaidi: I had a misunderstanding which was cleared up in another comment (what the stage was).

So I think this test is validating that the device supports the features the tester knows it supports - i.e. are the results of feature enumeration accurate. The new test could be "does the device support feature enumeration and return a list of features". In any case, this test is fine I think.

noursaidi (approved):

PR #654, opened 2023-05-26
Preliminary code to read live GCP Logs and emit metrics

johnrandolph:
Nothing here should be considered finalized. For example, some part of the bucketing and submitting at a threshold reflected a direct dependency on GCM submit RPC.
However this gets us started looking at live logs and emitting something.
This captures just errors that are generated by GCP Cloud IoT.


schema/monitoring_metric.json
```diff
@@ -0,0 +1,39 @@
...
+    "type": "object",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "additionalProperties": false,
+    "properties": {
+      "timestamp": {
+        "type": "string",
+        "format": "date-time"
+      },
+      "severity": {
+        "type": "string",
```
grafnu, line 14: enum?

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
```diff
@@ -0,0 +1,253 @@
...
+   */
+  protected class LogTailEntryList extends LinkedList<LogTailEntry> {
+
+  }
+
+  /*
+   * All the LogTail items currently processing.
+   * Hold the log entries in buckets for upstream processing in chunks.
+   */
+  protected class LogTimeSeries extends TreeMap<Long, LogTailEntryList> {
```
grafnu, line 139: This seems like a lot for an inner class.  Can it be made standalone?  Otherwise can it be static?  What from the containing class is it referencing?

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
```diff
@@ -0,0 +1,253 @@
...
+          "(severity=ERROR AND protoPayload.serviceName=\"cloudiot.googleapis.com\")";
+
+  public LogTail(String projectId) {
+    this.projectId = projectId;
+    this.logsTimeSeries = new LogTimeSeries();
+  }
+
+  public static void main(String[] args) {
+    String projectId = "essential-keep-197822";
+    LogTail lt = new LogTail(projectId);
```
grafnu, line 42: style nit (maybe it's Java) -- avoid abbreviations... e.g. consider logTail or tail

here and elsewhere

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
```diff
@@ -0,0 +1,253 @@
...
+          "(resource.labels.function_name=\"udmi_config\") OR " +
+          "(severity=ERROR AND protoPayload.serviceName=\"cloudiot.googleapis.com\")";
+
+  public LogTail(String projectId) {
+    this.projectId = projectId;
+    this.logsTimeSeries = new LogTimeSeries();
+  }
+
+  public static void main(String[] args) {
+    String projectId = "essential-keep-197822";
```
grafnu, line 41: I'd make this an argument at least... next step is going to be some kind of integration testing or something, so it definitely won't be this...

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
```diff
@@ -0,0 +1,253 @@
...
+    int statusCode;
+    String statusMessage;
+    Severity severity;
+    int value = 1;
+  }
+
+  /*
+   * Linked list of LogTailEntry.
+   */
+  protected class LogTailEntryList extends LinkedList<LogTailEntry> {
```
grafnu, line 131: private static?

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
```diff
@@ -0,0 +1,253 @@
...
+    // Be careful in here. Unwrapping this is a careful sequence.
+    Payload.ProtoPayload payload = log.getPayload();
+    if (!payload.getData().getTypeUrl().equals("type.googleapis.com/google.cloud.audit.AuditLog")) {
+      System.err.println("Not AuditLog");
+      return;
+    }
+    try {
+      AuditLog auditLog = AuditLog.parseFrom(payload.getData().getValue().toByteArray());
+      LogTailEntry entry = new LogTailEntry();
+      entry.timestamp = log.getInstantTimestamp();
```
grafnu, line 86: field needs to be converted to an iso timestamp -- there's some utility functions in GeneralUtils or JsonUtils that do this (look for "timestamp" or something like that).   Or an Instant might have the right thing, I forget.

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
```diff
@@ -0,0 +1,253 @@
...
+      entry.methodName = auditLog.getMethodName();
+      entry.serviceName = auditLog.getServiceName();
+      entry.resourceName = auditLog.getResourceName();
+      entry.statusCode = auditLog.getStatus().getCode();
+      entry.statusMessage = auditLog.getStatus().getMessage();
+      entry.severity = log.getSeverity();
+      logsTimeSeries.add(entry);
+      logsTimeSeries.maybeEmitMetrics();
+    } catch (InvalidProtocolBufferException e) {
+      System.err.println(e.toString());
```
grafnu, line 96: extend the component base class and then call debug() or error() or something

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
```diff
@@ -0,0 +1,253 @@
...
+    int LOG_ENTRIES_PER_SUBMIT = 100;
+
+    private void emitJsonTimeSeries(PrintStream output) throws JsonProcessingException {
+      ObjectMapper objectMapper = new ObjectMapper();
+      for (long k : keySet()) {
+        for (LogTailEntry log : this.get(k)) {
+          Monitoring monitoring = new Monitoring();
+          monitoring.metric = new MonitoringMetric();
+          loadMetricFields(monitoring.metric, log);
+          output.println(objectMapper.writeValueAsString(monitoring));
```
grafnu, line 152: JsonUtils.stringify() -- sets some standard properties on the output mapper.

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
```diff
@@ -0,0 +1,253 @@
+package com.google.bos.udmi.monitoring;
```
grafnu, line 1: seems like this class is trying to do two things?  One is gather metrics, the other is emit them -- I'd separate them out so they can be repurposed (e.g. emit other sources of metrics, or gather metrics to a different sink).

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
```diff
@@ -0,0 +1,253 @@
...
+
+  public void tailLogs() {
+    LogEntryServerStream stream = getCloudLogStream(LOG_FILTER);
+    processLogStream(stream);
+  }
+
+  /*
+   * Fields from GCP Logging stored in a neutral class.
+   */
+  protected class LogTailEntry {
```
grafnu, line 116: make static if possible. also private if possible

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
```diff
@@ -0,0 +1,253 @@
...
+    } else {
+      stream = logging.tailLogEntries(TailOption.filter(filter));
+    }
+    return stream;
+  }
+
+  /*
+   * Top level processor of incoming GCP log entries.
+   */
+  private void processLogEntry(LogEntry log) {
```
grafnu, line 61: unit test?

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
johnrandolph: Switched to using JsonUtil.stringify

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
johnrandolph: Yeah, this was just to get started i.e. draft PR, I split this into pieces.

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
johnrandolph: Will start writing them. Sending the class refactoring in parallel.

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTail.java
johnrandolph: Missed this. Will address.

johnrandolph: Just checking, you mean at the output stage, right? The JSON stringify already wrote in something like ISO8601.

schema/monitoring_metric.json
```diff
@@ -0,0 +1,41 @@
+{
+    "$udmi_version": "1.4.1",
+    "title": "Monitoring metric",
+    "description": "One metric",
+    "type": "object",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "additionalProperties": false,
+    "properties": {
+      "timestamp": {
```
grafnu, line 9: NB: There's an existing structure in common.json#entry that has many of the same fields and is commonly used in other parts of the system.  Having a uniform structure just means that it'll be easier for some tools to consume this in a more standardized way.  The format of that was something that could be easily consumer, e.g., CloudLogging.  It's used for status and log messages generated by devices and tools. Slightly different than a metric, but something to at least consider.

schema/monitoring_metric.json
```diff
@@ -0,0 +1,41 @@
...
+        "type": "string",
+        "format": "date-time"
+      },
+      "severity": {
+        "enum": [
+          "ERROR",
+          "NONE"
+        ]
+      },
+      "project": {
```
grafnu, line 19: Most of the constructs in UDMI use the message envelope (container for the message) to hold the project, location, registry, and device_id -- see envelope.json  -- this is essentially the structure as described here https://cloud.google.com/iot/docs/how-tos/mqtt-bridge  -- again, just something for us to consider to use common constructs when they make sense.

schema/monitoring_metric.json
```diff
@@ -0,0 +1,41 @@
+{
```
grafnu, line 1: maybe take a look at event_system.json  -- which is the current "proposed" (@Alpha) definition for the messages emitted by devices to report system metrics -- just to see how it's structured.  Not the same, of course, but just trying to make sure we're considering common constructs where appropriate.

johnrandolph: Ok I merged ptal

schema/common.json
```diff
@@ -30,45 +30,7 @@
...
-          "examples": [600]
-        }
-      },
-      "required": [
-        "message",
-        "category",
-        "timestamp",
-        "level"
-      ]
+      "$ref": "file:entry.json"
```
grafnu, line 33: this seems like an unnecessary redirection for all the existing uses -- does a s/x/y replace work to update existing uses to not have this reference in common.json?

schema/monitoring_metric.json
johnrandolph: Yeah, yesterday in our 1:1 I said I thought it might be a bit of a stretch to co-mingle the two. But we are visualizing it now. My other idea is to move just deviceId deviceNumId etc fields into a separate schema and refer to them in envelope.json.  (Not as a nested map, but as first level properties).  Then we share the definition in both places but break the association with messaging. WDYT?

grafnu:
I think for now I'd leave it as you have it here, I don't think the "right"
answer will become apparent until we actually try to integrate with some
other system. We could discuss things back and forth for a bit and circle
around something... but I think at this stage it's more important to get
something "kinda close" out the door...

schema/monitoring_metric.json
johnrandolph: I toyed around with moving the fields into a resource.schema or similar but I couldn't get the $ref to work (and keep the depth/structure of envelope). I'm just going to keep it how it is for now.

schema/common.json
johnrandolph: I think this $ref vs inline decl is equivalent depth? i.e. field access is still entry.message.

grafnu: It introduces an unnecessary redirection, and multiple ways of specifying the same thing. Other places have

` "$ref": "file:common.json#/definitions/entry"`

which then maps directly to 

`"$ref": "file:entry.json"`

I see no reason to keep the original "#/definitions/entry" version around.

johnrandolph: Oh I get you now. Done

johnrandolph (commented):
Chipping away at this. Not done yet.

common/src/test/java/com/google/bos/udmi/monitoring/LogTailTest.java
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  @Before
+  public void openMocks() {
+    closeable = MockitoAnnotations.openMocks(this);
+  }
+
+  @After
+  public void releaseMocks() throws Exception {
+    closeable.close();
```
grafnu, line 36: what's the purpose/impact of this closeable.close()?  Asking because I don't have anything similar in the code I've done so I don't know if I'm missing something or not...

gencode/java/udmi/schema/Envelope.java
```diff
@@ -25,7 +26,10 @@
     "deviceRegistryLocation",
     "projectId",
     "payload",
+    "source",
+    "gatewayId",
     "transactionId",
+    "publishTime",
```
grafnu, line 32: these changes seem sus -- is your working tree up to date?

johnrandolph:
Have more test coverage now. Please review. Not sure what the issue is with schema suddenly.

grafnu:
You changed the schema (properties that used to be required are no longer
required (as per the change you made in schema/common.json) and so the
tests are picking that up. Logentries are generally expected to have a
message and timestamp, so in this case I'd suggest it's WAI (at least as
far as the tests are concerned)... I'm not sure it's semantically correct
to make those fields no longer required. I'm not sure if that adversely
affects your code/logging, but at least that's what's going on with the
rest of the system.

udmis/build.gradle
```diff
@@ -28,12 +28,14 @@ sourceSets {
         java {
             srcDirs '../gencode/java'
             srcDirs '../common/src/main/java'
+            srcDirs 'src/main/java'
```
grafnu, line 31: these lines (here and under test) shouldn't be necessary -- they are implicit

udmis/build.gradle
```diff
@@ -56,7 +58,7 @@ jacocoTestReport {
 
 checkstyle {
     ignoreFailures = false
-    maxWarnings = 0
+    maxWarnings = 32
```
grafnu, line 61: cleanup warnings

udmis/build.gradle
```diff
@@ -94,12 +97,12 @@ dependencies {
...
 
     implementation 'com.google.cloud:google-cloud-pubsub:1.66.0'
+    implementation 'com.google.cloud:google-cloud-logging:3.14.9'
     implementation 'org.jetbrains:annotations:20.1.0'
     
     implementation 'org.eclipse.paho:org.eclipse.paho.client.mqttv3:1.2.5'
 
-    implementation group: 'junit', name: 'junit', version: '4.13.2'
-
+    // implementation group: 'junit', name: 'junit', version: '4.13.2'
```
grafnu, line 105: remove if not needed

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTailEntry.java
```diff
@@ -0,0 +1,28 @@
...
+  Instant timestamp;
+  String methodName;
+  String serviceName;
+  String resourceName;
+  int statusCode;
+  String statusMessage;
+  Severity severity;
+  int value = 1;
+
+  public boolean equals(LogTailEntry other) {
```
grafnu, line 17: There's a new thing in Java called a "record class" that makes some of this boiler-plate unnecessary -- see https://docs.oracle.com/en/java/javase/20/language/records.html -- don't care if you use it or not, but I was just reading about it so had to mention it!

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTimeSeries.java
```diff
@@ -0,0 +1,179 @@
...
+      metric.envelope.projectId = matcher.group("project");
+      metric.envelope.deviceRegistryLocation = matcher.group("location");
+      metric.envelope.deviceRegistryId = matcher.group("registry");
+      metric.envelope.deviceId = matcher.group("device");
+    } // else do something?
+  }
+
+  private LogSeverity severityToLogSeverity(Severity severity) throws IOException {
+    // Subtle: These are two different types with different enum value ranges. Translate manually.
+    switch (severity) {
```
grafnu, line 114: (other common ways to do this are with an (immutable) map or with a string conversion) -- just FYI, no need to change

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTimeSeries.java
```diff
@@ -0,0 +1,179 @@
...
+        } else {
+          reInsert.add(log);
+        }
+      } else {
+        reInsert.add(log);
+      }
+      this.put(k, reInsert);
+      numberOfLogs++;
+    }
+    // return true;
```
grafnu, line 161: remove dead code

udmis/src/test/java/com/google/bos/udmi/monitoring/LogTailTest.java
```diff
@@ -0,0 +1,329 @@
...
+
+    AuditLog auditLogMock = Mockito.mock(AuditLog.class);
+    try {
+      Mockito.doReturn(auditLogMock).when(logTailMock).getAuditLogParseFrom(any(byte[].class));
+    } catch (InvalidProtocolBufferException e) {
+      fail(e);
+    }
+
+    LogTailEntry entryTest = getLogTailEntryTest();
+    Mockito.doReturn(entryTest.timestamp).when(logEntryMock).getInstantTimestamp();
```
grafnu, line 130: (FYI you can make "doReturn" a static import and then all the leading Mockito. things go away -- doesn't make much difference, but I find it visually much cleaner)

udmis/src/test/java/com/google/bos/udmi/monitoring/LogTailTest.java
```diff
@@ -0,0 +1,329 @@
...
+    } catch (IOException e) {
+      fail(e);
+    }
+
+    logTailMock.processLogEntryError(logEntryMock);
+    assertTrue(argumentCaptor.getValue().equals(entryTest));
+  }
+
+  /*
+   This test fails because IOException is not thrown from maybeEmitMetrics().
```
grafnu, line 155: but what's the problem?  Is it that it should throw an exception, or that the test is wrong?

johnrandolph: (I sent this to generate your next round of comments in parallel of working on this) Yeah the test doesn't function -- the test code does not function so that the code UNDER test throws a IOException from maybeEmitMetrics.

udmis/src/test/java/com/google/bos/udmi/monitoring/LogTailTest.java
johnrandolph: Sure, I made that the case for everything, to be consistent.

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTailEntry.java
johnrandolph: Nothing this as a TODO. e.g. If the equals() becomes more complex let's revisit.

udmis/src/main/java/com/google/bos/udmi/monitoring/LogTailBase.java
```diff
@@ -0,0 +1,10 @@
+package com.google.bos.udmi.monitoring;
+
+import com.google.bos.udmi.service.pod.ContainerBase;
+
+/**
+ * Extend ContainerBase with UDMI Monitoring specific functions.
+ */
+public class LogTailBase extends ContainerBase {
```
grafnu, line 8: NB: I don't understand the purpose of this class, but it's not hurting anything.

grafnu (approved):

PR #657, opened 2023-06-01
Fix for validator NPE with missing pointset metadata

grafnu:


bin/test_combos
```diff
@@ -0,0 +1,21 @@
...
+for test in $tests; do
+    echo '################' Running combo test $test
+    traces=$(jq -r .traces $test)
+    site_model=$(jq -r .site_model $test)
+    validator_config=$(jq -r .validator $test)
+    if [[ $validator_config != null ]]; then
+        validator/bin/validate -- schema trace $traces $site_model
+        check=$(echo $validator_config | jq -r .check)
+        echo $check
+        eval $check
```
johnrandolph, line 17: Bare/unprotected eval (related: using bare 'source', in PR 561) continues to be a red flag to me.

johnrandolph (approved):

bin/test_sites
```diff
@@ -30,6 +30,7 @@ function redact_files {
     for file in $@; do
         sed -E -i \
             -e 's-While loading.*udmi/tests/-REDACTED/-' \
+            -e 's/at .*main\(.*\.java:[0-9]+\)/REDACTED/' \
```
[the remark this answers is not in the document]
grafnu: > What are you redacting, here? It looks like part of a stack trace?

See tests/sites/downgrade/devices/DWN-1/expected/exceptions.txt -- the stack trace is already redacted, this is just then adding in an obvious REDACTED marker... I was getting confused because the output was just removed before so just looking at that file was confuzel. Now, at least there's a big REDACTED marker to clarify.

bin/test_combos
grafnu: > Bare/unprotected eval (related: using bare 'source', in PR 561) continues to be a red flag to me.

What would you suggest instead?  This is just implementing execution of the contents of tests/combos/missing_pointset.json -- which is trying to find the right balance of flexibility with structure.

johnrandolph: My simplest suggestion, i.e. least delta from what we have now, versus changing the test rig overall, just make it more explicit:

- in the test definition write something structured like "validator": {  { "fgrep_check": { "path":... , "pattern":...}}   and here when you perform the fgrep, quote the arguments

grafnu:
That seems unnecessarily complicated... The problem is there's a LOT of
different ways to check something, fgrep is just one of them... I don't see
the utility of doing all that overhead for something that's really just
mimicking what is on the command line... ?  What specifically are you
worried about with the "eval" part?  It's really just tunneling code from
one place to another.

bin/test_combos
grafnu: This is more or less modeled after the way github actions does things --https://github.com/faucetsdn/udmi/blob/master/.github/workflows/testing.yml#L59 -- which just runs the command line as given. Seems to work reasonably well in the context...

johnrandolph: I don't see the line you're comparing to. The testing yaml is just running commands that were already written, right? My concern with eval is that if you can change the shell or invoke something else.  I realize we are invested in the shell script tests but we could at least be as careful as possible here. Adding a fgrep handler shouldn't be too hard?

grafnu:
https://github.com/faucetsdn/udmi/blob/master/.github/workflows/testing.yml#L59

        run: bin/run_tests registrar_tests

There's no functional difference there between that and what I have -- it's
a field of the test runner script that is executed in the given
environment. it's not "adding a fgrep handler" that's hard, it's "adding a
handler for every kind of check that we might want."  E.g., what to do with
[[ $(jq .version input_file.json) == "1.4.1"]] -- or what-not. I understand
the concern with eval in an open environment, e.g. where there's some input
that is provided in a running system that somebody could type something
sneaky into, but that's not this.

PR #658, opened 2023-06-02
Schema validation through sequencer

grafnu:
This PR is just about doing the validation and generating the resulting SCHEMA output lines, and adding synthetic RESULT lines at the end to summarize.


grafnu:
Current output is something like https://github.com/grafnu/udmi/actions/runs/5158002651/jobs/9291129797#step:10:12

noursaidi (approved):

PR #662, opened 2023-06-09
Fix IDE testing and naked eval

grafnu:


bin/test_combos
```diff
@@ -14,8 +17,19 @@ for test in $tests; do
         validator/bin/validate -- schema trace $traces $site_model
         check=$(echo $validator_config | jq -r .check)
         echo $check
-        eval $check
+        code=0
+        result=$(eval $check) || code=$?
```
johnrandolph, line 21: Note: This protects the shell the script is running in but not anything else. Still it's progress.

johnrandolph (approved):

PR #663, opened 2023-06-09
Fixing issues with gateway creation

grafnu:
Addresses the issue where a device switches from being a non-gateway to a gateway.


noursaidi (approved):
Haven't tested whether this also needs fixing in the UDMIS/Reflector registrar workflow, but works normally

PR #665, opened 2023-06-13
Add credentials file and run script

johnrandolph:


bin/monitoring
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash -e
```
grafnu, line 1: I'd put this in udmis/bin -- just trying to push some things down to more specific levels when we can.

grafnu (approved):

PR #669, opened 2023-06-21
UDMIS dynamic switching between ClearBlade and GCP for IoT Core

grafnu:


noursaidi:
Tested and works, some config changes needed as described below
Notes:
- Docker run command needs to change to `["/root/bin/run", "/udmi/pod-config.json"]`
- UDMIS needs to be built (udmis/bin/build) before building docker container
- `CLEARBLADE_CONFIGURATION` env var needs to be defined in the pod and key needs to be passed in as a secret (see below) to use clearblade/dynamic
- pod_config should be sourced from `udmis/src/test/configs/dynamic_pod.json`, which itself needs to be updated for the projects
- pod_config should ideally be a [ConfigMap](https://kubernetes.io/docs/concepts/configuration/configmap/), but that's more a future note and shouldn't need to be incorporated in this PR 

My pod config:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: udmi-test-pod
spec:
  containers:
    - name: test-pod
      image: gcr.io/PROJECT_ID/udmis
      env:
      - name: CLEARBLADE_CONFIGURATION
        value: "/udmi/clearblade.json"
      imagePullPolicy: Always
      resources:
        requests:
          cpu: 100m
          memory: 100Mi
      volumeMounts:
      # name must match the volume name below
      - name: secret-volume
        mountPath: /udmi
        readOnly: true
  # The secret data is exposed to Containers in the Pod through a Volume.
  volumes:
  - name: secret-volume
    projected:
      sources:
      - secret:
          name: pod-config.json
      - secret:
          name: clearblade.json
```

grafnu:
Thanks -- I made some of the changes, but will address some others later. Will surely need more testing and iteration before declared complete!  I want to get this in ASAP because I want to avoid some merge issues with other work I'm doing...

PR #679, opened 2023-07-06
adding possibility to push to alternative image registry

pisuke:
We need to be able to push to artifacts registries, as gcr.io is getting deprecated in 2024

### DIFF
--- a/udmis/bin/container
+++ b/udmis/bin/container
@@ -6,17 +6,25 @@ DROOT=.
 
 function usage {
     echo Error: $*
-    echo Usage: $0 { build, shell, run, push, deploy, update, status, logs, stop }
+    echo Usage: $0 { build, shell, run, push, deploy, update, status, logs, stop } [repo]
     echo Project: $PROJECT
     exit 1
 }
 
-PROJECT=$(gcloud config get project)
-REPOSITORY=gcr.io/$PROJECT
-IMAGE=udmis
-
 cmd=$1
 shift || usage missing command
+repo=$1
+
+PROJECT=$(gcloud config get project)
+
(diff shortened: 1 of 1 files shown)


udmis/bin/container
```diff
@@ -6,17 +6,25 @@ DROOT=.
...
+shift || usage missing command
+repo=$1
+
 PROJECT=$(gcloud config get project)
-REPOSITORY=gcr.io/$PROJECT
+
+if [[ -z $repo ]]; then
+    REPOSITORY=gcr.io/$PROJECT
+else
+    REPOSITORY=$repo
```
grafnu, line 23: Just curious what "$repo" looks like (e.g., what's an example?) -- my only comment/question here is if it would be better to just have the base effected (so replace gcr.io with something else) and then always append the $PROJECT (which is dynamic).  Otherwise, it's possible to confuzel because the project used for REPOSITORY might be different than the PROJECT as determined automatically.  But again, I'm not 100% sure what that value looks like, which is why I ask!

pisuke: In artifacts registry the repo name includes the name of the project, region, type of repo and optional folders.
Here is an example referring to a `docker` type of repository project named `udmi-project` located in the `us-central1` region:

us-central1-docker.pkg.dev/udmi-project/udmi/udmis:latest

Other types of repositories are also possible (Maven, npm, python, etc.)

Artifact registry specifically uses the name "Repository" for this sort of things.

Screenshot from 2023-07-12 12-13-56

grafnu: Ok -- yes, definitely I was thinking too narrow here with the usage!

grafnu (approved):

PR #689, opened 2023-07-19
Add sequencer tests for software, hardware, config logging, pointset_publish

noursaidi:
Adding tests for:
- state `software` matches metadata
- State `make` & `model` matches metadata
- Device implements minimum expected logging events associated with receiving a config message
- device publishes pointset (not really a test, more of an information about the device/configuration)

CI testing and cleanup of generated `sequences.md` needed - but thought to get general feedback.


validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/SystemSequences.java
```diff
@@ -0,0 +1,82 @@
...
+   *  For example, if firmware v1.1 is installed, the device may report the version of all packages
+   *  within this firmware
+   */
+  @Test(timeout = THREE_MINUTES_MS)
+  @Feature(stage = BETA, bucket = SYSTEM)
+  @Summary("device publishes correct software information in state messages")
+  public void state_software() {
+
+    Map metadata_software = catchToNull(() -> deviceMetadata.system.software);
+    if ( metadata_software == null ) {
```
grafnu, line 65: no spaces after (  -- or before )

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/SystemSequences.java
```diff
@@ -0,0 +1,82 @@
...
+      throw new AssumptionViolatedException(
```
grafnu, line 66: keep on same line?

grafnu, line 66: we should also consider creating a helper function (akin to ifNotNullGet)

ifNullThen(metadata_software, () -> throw new AssumptionViolationException())

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
```diff
@@ -153,4 +153,24 @@ public void extra_config() {
     untilLogged(SYSTEM_CONFIG_APPLY, SYSTEM_CONFIG_APPLY_LEVEL);
   }
 
+  @Test(timeout = ONE_MINUTES_MS)
+  @Feature(stage = BETA, bucket = SYSTEM)
+  @Summary("Check that the device publishes minimum required log entries when receiving config")
+  public void config_logging() {
+    deviceConfig.system.min_loglevel = Level.DEBUG.value();
```
grafnu, line 160: how is this test functionally different than what's covered already?  Some of the other tests essentially do exactly this along with some other stuff... ?  Each tests adds an overhead, so just wondering what the underlying motivation is...

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
```diff
@@ -153,4 +153,24 @@ public void extra_config() {
...
+    updateConfig("send device config set loglevel to debug");
+
+    // Wait for device to receive the config
+    // We can untilLogged the system.config.apply event, but following the natural ordering
+    // of receive --> parse --> apply may be more helpful
```
grafnu, line 165: may?  I don't understand this explanation.  Seems unnecessary...

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
```diff
@@ -153,4 +153,24 @@ public void extra_config() {
     untilLogged(SYSTEM_CONFIG_APPLY, SYSTEM_CONFIG_APPLY_LEVEL);
   }
 
+  @Test(timeout = ONE_MINUTES_MS)
+  @Feature(stage = BETA, bucket = SYSTEM)
+  @Summary("Check that the device publishes minimum required log entries when receiving config")
+  public void config_logging() {
+    deviceConfig.system.min_loglevel = Level.DEBUG.value();
+    updateConfig("send device config set loglevel to debug");
```
grafnu, line 161: this should be a declarative statement, not an action -- so something like just "loglevel to debug", so the end output reads something like "update config loglevel to debug"

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
```diff
@@ -153,4 +153,24 @@ public void extra_config() {
...
+
+    // Wait for device to receive the config
+    // We can untilLogged the system.config.apply event, but following the natural ordering
+    // of receive --> parse --> apply may be more helpful
+    safeSleep(CONFIG_THRESHOLD_SEC * 2000);
+
+    deviceConfig.system.min_loglevel = Level.TRACE.value(); // Force a new config to be sent
+    updateConfig("device receive config");
```
grafnu, line 169: "loglevel to trace"

docs/specs/sequences/generated.md
```diff
@@ -120,6 +142,18 @@ device publishes pointset events at a rate of no more than config sample_rate_se
 1. Wait for receive at least 5 pointset events
 1. Check that time period between successive pointset events is between 1 and 5 seconds
 
+## state_make_model (BETA)
+
+device publishes correct make and model information in state messages
+
+1. Test skipped: make and model not defined in metadata
```
grafnu, line 149: need a model update!

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/SystemSequences.java
grafnu: check out the latest I submitted, I created the helper function and used it a few places. Look for skipTest(), ifNullThen(), ifNullSkipTest().  Maybe we should also add a ifCatchNullSkipTest() since that's a pattern used in a few places...

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
grafnu: there's also a new function I created, something like forceUpdate() -- does the same thing without arbitrarily changing a value...

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
noursaidi: They aren't really, `extra_field` and `broken_config` do both currently test for the exact same config logging, but as a  dependancy upon those behaviours for testing additional required behaviours. 

There isn't anywhere to say that as a basis "when a device receives a config message it should publish receive, parse, apply system logentries" which is needed. Although, I didn't include anything about the status block

The motivation is better information for both developers and report consumers

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
noursaidi: I tried it - but it doesn't seem to do the trick? 
Using force update - https://github.com/noursaidi/udmi/actions/runs/5740987796/job/15560832585#step:6:468
Setting loglevel to trace to trigger an update - https://github.com/noursaidi/udmi/actions/runs/5740097044/job/15558012902#step:6:518

grafnu:
Huh -- that was totally broken.  This commit fixes it -- I'll push it in
with the PR I'm submitting soon <https://github.com/faucetsdn/udmi/pull/704>
.
https://github.com/faucetsdn/udmi/commit/65b9508c25268ca55cb4b15c88bcb7ec37da9a65

Note that you shouldn't need that safeSleep in there either -- maybe that
was just thrown in because of the other bug? I just tried it without that
sleep and it worked ok.

noursaidi:
Unfortunately the force config still doesn't seem to work for me with the latest PR the `updated` boolean is false because both `sentBlockConfig` and `data` are null?
https://github.com/faucetsdn/udmi/blob/1d8f049b9f3b440191b0819084052c2673a4d354/validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java#L996

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
noursaidi: Removed

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
noursaidi: statement updated

docs/specs/sequences/generated.md
noursaidi: model update done and now referenced in CI testing

docs/specs/sequences/generated.md
```diff
@@ -94,6 +110,12 @@ Check that the device correctly handles an extra out-of-schema field
 1. Check that all enumerated features are official buckets
 1. Check that no point enumeration
 
+## pointset_publish (BETA)
+
+device publishes pointset events
+
+1. Wait for recieve a pointset event
```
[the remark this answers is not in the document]
noursaidi: i after e except after c .... corrected

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/SystemSequences.java
```diff
@@ -0,0 +1,82 @@
...
+    if (expected_make == null || expected_model == null) {
+      throw new AssumptionViolatedException(
+          "make and model not defined in metadata"
+      );
+    }
+
+    String actual_make = catchToNull(() -> deviceState.system.hardware.make);
+    String actual_model = catchToNull(() -> deviceState.system.hardware.model);
+
+    checkThat(format("make in state `%s` matches make in metadata `%s`",
```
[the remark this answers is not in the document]
noursaidi: Removed parameters from output

grafnu:
> @grafnu I did something to the forceUpdate and it works now (for me anyway) - but I don't know if it was right .. [27bf958](https://github.com/faucetsdn/udmi/commit/27bf958ed3dd3587cf9f2b9e90908d352a66d40e)

From what I can tell I think both changes are what's needed for proper theory of operation.

Oh, I just saw this after I did my other commit note.  Now I'm second-guessing myself, because that function (forceConfigUpdate()) -- isn't used anywhere but I thought I had used it.  Oh well -- we should at least get it working.

noursaidi:
That's a handy feature, thanks! Still passes.

You do use the "force config" just not the `forceConfigUpdate` which I added - https://github.com/faucetsdn/udmi/blob/1d8f049b9f3b440191b0819084052c2673a4d354/validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java#L1727

grafnu (approved):

PR #694, opened 2023-07-25
Switch to alpine for UDMIS

grafnu:


udmis/bin/run
```diff
@@ -11,9 +11,12 @@ shift
 ROOT=$(dirname $0)/..
 cd $ROOT
 
+cat /etc/os-release
+
+java -version
```
ohshazbot, line 16: This is smart, I like this

ohshazbot (approved):

PR #699, opened 2023-07-28
Fix periodic_scan sequence

MertCingoz:
~Update sequencer expectation when device supports multiple discovery/localnet families~

~- family_ether_addr~
~- family_ipv4_addr~
~- family_ipv6_addr~
- periodic_scan

### DIFF
--- a/etc/sequencer_report.md.template
+++ b/etc/sequencer_report.md.template
@@ -6,7 +6,7 @@
 
 ## Device Identification
 
-| Device | AHU-1 |
+| Device | {{report.device_id}} |
 |---|---|
 | Site |  |
 | Make | {{report.device_make}} |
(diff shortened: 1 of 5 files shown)


grafnu:
The format for the localnet state families should not have the interface name in them, rather just the name of the address family { ipv4, ether, etc... }. If there are multiple, then the system state should indicate the address for the _primary_ interface (as determined by, e.g., the default route).  The reason being is that backend systems don't really care about the interfaces (and there really should just be one anyway) -- rather they just want to know about _the_ IPv4 address for the device!

MertCingoz:
@grafnu I updated PR to only fix periodic scan sequence for now.
It should be reproducible by discovering 2 (AHU-22 and one more) devices using `virtual` family

grafnu (approved):

PR #701, opened 2023-07-30
new pointset-matches-config tests

grafnu:
1. Add pointset_remove_point and pointset_request_extraneous ALPHA tests.
2. Bunches of cleanup and debugging improvements.
3. Fixes to Pubber for pointset publishing (b/c of tests!)


validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/PointsetSequences.java
```diff
@@ -1,81 +1,169 @@
...
+    untilPointsetSanity();
+
+    deviceConfig.pointset.points.put(EXTRANEOUS_POINT, new PointPointsetConfig());
+
+    try {
+      untilTrue("pointset status contains extraneous point error",
+          () -> ifNotNullGet(deviceState.pointset.points.get(EXTRANEOUS_POINT),
+              state -> state.status.category.equals(POINTSET_POINT_INVALID)
+                  && state.status.level.equals(POINTSET_POINT_INVALID_VALUE)));
+      untilHasInterestingSystemStatus(true);
```
noursaidi, line 79: Why is this necessary if the device has already reported the status of the point as an error?

I do have a concern with the system status  checking in sequencer tests generally that devices may correctly, or incorrectly, use that status to relay some information about the device which may prevail over these errors.

grafnu:
Well, the goal is that the system status is a singular high-order bit of
the status of the device, which is just the worst of something else going
on. If there's some other information that should prevail (e.g. higher
log-level) then that's fine, but if not, then they shouldn't do that!  The
goal is to keep the backend logic a little bit easier just to avoid having
to understand/parse the entire message structure. Although it can go either
way, we should just make sure we're clear.  How about we discuss a bit next
we meet.

noursaidi (approved):

PR #703, opened 2023-07-31
Add markdown sequencer report

noursaidi:
Example:

```
# FCU-504

- Start 2023-07-31T09:49:51Z
- End: 2023-07-31T10:04:21Z
- Status: Run completed

## Device Identification

| Device | AHU-1 |
|---|---|
| Site |  |
| Make |  |
| Model |  |
| Software |  |

## Summary

| | Feature | Stable | Beta | Preview |
| --- | --- | --- | --- | --- |
| - | enumeration | 0/0 | 0/0 | 0/1 |
| - | enumeration.features | 0/0 | 0/0 | 0/1 |
| ✕ | pointset | 0/0 | 1/2 | 0/0 |
| ✕ | system | 1/1 | 1/2 | 0/0 |


## Results

| Bucket | Feature | Stage | Result | Description |
| --- | --- | --- | --- | --- |
| enumeration | empty_enumeration | preview | fail | timeout waiting for enumeration not active |
| enumeration.features | feature_enumeration | preview | fail | timeout waiting for enumeration not active |
| pointset | pointset_publish_interval | beta | pass | Sequence complete |
| pointset | pointset_sample_rate | beta | skip | measured sample rate is too low for automated test |
| system | device_config_acked | beta | pass | Sequence complete |
| system | extra_config | beta | fail | timeout waiting for log category `system.config.parse` level `DEBUG` was logged |
| system | system_last_update | stable | pass | Sequence complete |


## Schema

| Stage | Result | 
| --- | --- |
| stable | ✓ |
| beta | - |
| preview | - |
```


bin/sequencer_report
```diff
@@ -0,0 +1,265 @@
...
+        return "-"
+
+
+@dataclass
+class FeatureScore:
+    """Container for scored points and total points for a feature and stage"""
+
+    scored: int = 0
+    total: int = 0
+    stage: str = ""
```
grafnu, line 59: is this ever used? (stage)

bin/sequencer_report
```diff
@@ -0,0 +1,265 @@
...
+                    SchemaResult(schema, result["stage"], test_name, result["result"])
+                )
+
+        # Ensure all results include at least items from SCHEMAS set of schemas in report
+        schemas = SCHEMAS[:]
+        schemas.extend(x.schema for x in self.schema_results if x.schema not in schemas)
+
+        self.schema_results_by_schema = {
+            k: all_or_none([x.passed() for x in self.schema_results if x.schema == k])
+            for k in schemas
```
grafnu, line 145: this indent seems wrong -- shouldn't the for be +2 or +4 in?

bin/sequencer_report
```diff
@@ -0,0 +1,265 @@
...
+    def add(self, result, score):
+        self.total += score
+        if result == "pass":
+            self.scored += score
+
+    def has(self):
+        """Did the sequencer results have this feature & stage combination?"""
+        return self.total > 0
+
+    def passed(self):
```
grafnu, line 70: this is logic (not just rendering) -- we should consider moving it upstream to the java code, otherwise it might get out of sync with some other code that does the same thing!  Maybe not necessary now, but at least add a TODO that says something to that effect.  E.g., what happens if we decide (pass is getting 80%?)... :-P

grafnu (approved):
A few minor comments, but otherwise good to go!

bin/sequencer_report
noursaidi: I think because it's part of the same list comprehension  as the `k: all_or_none` it shouldn't be indented. The linter is happy with it, and so is the black autoformatter

bin/sequencer_report
noursaidi: It's set but not used. To differentiate between multiple `FeatureScore` (now `FeatureStageScore` objects belonging to the same feature

bin/sequencer_report
noursaidi: True - note added

PR #705, opened 2023-08-02
fix lint error failing CI tests

noursaidi:


noursaidi:
Merging because fixing mistake in PR which was just merged

PR #710, opened 2023-08-09
Handle legacy state messages

grafnu:


udmis/src/test/messages/bad_state.json
noursaidi, line 1: Rename `legacy_state.json`

PR #711, opened 2023-08-12
Add facilities for auto-updating UDMIS deployments in GCP projects

grafnu:


noursaidi (approved):
I haven't tested it, but looked through and LGTM. Is there documentation I can follow to try it, and just a note to say this should be recommended on a standalone cluster, atypical configurations (service accounts/workload identity, namespaces, etc) may cause this to not work

grafnu:
I'm sure all sorts of things will cause it not to work :-).  Namespaces I'm
not too worried about (would just need to add a configuration secret or
something in there to address), but the other things I'm not sure about.

Also note that there might be a few breaking changes with the "container"
script -- I had to fiddle with the order of arguments a bit. Not sure how
much that might break anything else you've been tracking.

PR #736, opened 2023-10-09
Add sequences into report

noursaidi:
Adds sequences into report. Link to example report below. 

CI testing additions just need to be fixed (which is a check that the sequence.md files in validator/sequences are up to date , otherwise there is a risk of comparing to an out of date sequence.md file)

[Sample report (click here) - the sequences.md don't actually match the results, just for illustration](https://gist.github.com/noursaidi/fe571a1b0f62f95ca1b6ae51672ccf3d)


bin/sequencer
```diff
@@ -158,6 +158,6 @@ fi
 fgrep RESULT $site_model/out/devices/$device_id/RESULT.log | \
     fgrep ' schemas ' | sort -k 3 | tee $SCHEMA_OUT
 
-(source venv/bin/activate && bin/sequencer_report $site_model $device_id)
+(source venv/bin/activate && python3 bin/sequencer_report.py $site_model $device_id)
```
grafnu, line 161: rather than make them explicitly python, I think a better solution would be to do something like

PATH=venv/bin:$PATH bin/sequencer_report $site_model $device_id -- so it will pick up the venv as needed.

bin/test_sequencer
```diff
@@ -135,4 +135,10 @@ diff -u out/generated.md docs/specs/sequences/generated.md || (
 echo Checking for duplication in out/sequencer.out and etc/sequencer_planning.txt
 bin/test_sequencer_plan
 
+echo Checking that the sequences in validator/sequences are correct
```
grafnu, line 138: Wouldn't this be checking the sequences in the generated site_model (out/) are correct (verified against validator/sequences)?

I think part of the problem here is the overall generation/updating of these things is a bit... wonky?  I'm not happy with whatever hack I have in place so some of this stuff is confusing me (even before you touched it at all!)

bin/test_sequencer
```diff
@@ -135,4 +135,10 @@ diff -u out/generated.md docs/specs/sequences/generated.md || (
...
+ls -1 sites/udmi_site_model/out/devices/AHU-1/tests/ \
```
grafnu, line 139: My concern here is that this would ultimately end up being too unstable, since it would be checking all the alpha/etc... tests.  Also, we should just make sure this is reconciled with the bit in test_sequencer that checks the generated.md file, since it's functionally supposed to be the same thing.

noursaidi:
@grafnu PTAL again

grafnu (approved):
Oh boy -- this is going to have some fun merge conflicts when I pull it into my current PR (just changing the testing structure)... but I'll sort all that out as needed.

PR #787, opened 2023-12-18
Fix pubber gateway proxy runtime issues

MertCingoz:


.gitignore
```diff
@@ -32,6 +32,7 @@ credentials.json
 /sites/
 /attributes.json
 /venv/
+/venv
```
grafnu, line 35: how is this different than the line above?

pubber/src/main/java/daq/pubber/BasicPoint.java
```diff
@@ -36,7 +36,11 @@ public BasicPoint(String name, boolean writable, String units) {
     dirty = true;
   }
 
-  abstract Object getValue();
+  protected abstract Object getValue() throws IllegalStateException;
```
grafnu, line 39: NB: I generally try to avoid the Java-explicit-exceptions model because, well, I think it's dumb.  You'll see the general style in the code is to try/catch all exceptions locally and re-throw as a RuntimeException with additional context.  So, unless there's a strong reason for it (e.g. something is actually using the IllegalStateException for some specific behavior), I'd prefer to keep it that way (so not have the `throws IllegalStateException` addition to the signature).

pubber/src/main/java/daq/pubber/ManagerBase.java
```diff
@@ -80,10 +80,16 @@ public void error(String message) {
...
     int intervalSec = ofNullable(options.fixedSampleRate).orElse(reportInterval);
+    if (intervalSec < DISABLED_INTERVAL) {
+      error(format("Dropping update interval, negative sample rate %ds", intervalSec));
+      return;
+    }
     if (periodicSender == null || intervalSec != sendRateSec.get()) {
       cancelPeriodicSend();
       sendRateSec.set(intervalSec);
-      startPeriodicSend();
+      if (intervalSec > DISABLED_INTERVAL) {
```
grafnu, line 90: This check seems redundant because of the previous check. Also, the case where intervalSec == DISABLED_INTERVAL likely isn't handled correctly.

pubber/src/main/java/daq/pubber/ManagerBase.java
```diff
@@ -80,10 +80,16 @@ public void error(String message) {
   protected void updateInterval(Integer sampleRateSec) {
     int reportInterval = ofNullable(sampleRateSec).orElse(DEFAULT_REPORT_SEC);
     int intervalSec = ofNullable(options.fixedSampleRate).orElse(reportInterval);
+    if (intervalSec < DISABLED_INTERVAL) {
+      error(format("Dropping update interval, negative sample rate %ds", intervalSec));
```
grafnu, line 84: sample rate doesn't necessarily need to be negative, just less than DISABLED_INTERVAL?  Maybe adjust error message accordingly.

pubber/src/main/java/daq/pubber/ManagerBase.java
```diff
@@ -125,6 +132,9 @@ private void stopExecutor() {
     }
   }
 
+  protected void pause() {
```
grafnu, line 135: @Override?

pubber/src/main/java/daq/pubber/ProxyDevice.java
```diff
@@ -39,15 +40,21 @@ protected void activate() {
...
     deviceManager.updateConfig(config);
+    pubberHost.publisherConfigLog("apply", null, deviceId);
   }
 
+  @Override
   protected void shutdown() {
     deviceManager.shutdown();
   }
 
+  @Override
```
grafnu, line 53: how is resume (unpause) handled? I see where pause() is called, but don't quite grock how un-pause is managed.

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -560,6 +557,7 @@ private void processDeviceMetadata(Metadata metadata) {
       configuration.algorithm = catchToNull(() -> metadata.cloud.auth_type.value());
     }
 
+    isGatewayDevice = false;
```
grafnu, line 560: I strongly prefer either a single-assignment or final style of variable management (unless something explicitly required like in a loop).  One easy fix here would be to declare "isGatewayDevice" as final and then have an else clause for the false assignment.

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -568,12 +566,11 @@ private void processDeviceMetadata(Metadata metadata) {
           configuration.algorithm = authType.value();
         }
       }
+      isGatewayDevice = ifNotNullGet(metadata.gateway.proxy_ids, p -> true, false);
```
grafnu, line 569: I think this can be simplified to just "metadata.gateway.proxy_ids != null"?  Or I think there's a "ifNotNullThen" helper function that does the same without the lambda.

pubber/src/main/java/daq/pubber/Pubber.java
```diff
@@ -696,6 +693,7 @@ protected void startConnection(Function<String, Boolean> connectionDone) {
   private boolean attemptConnection() {
     try {
       isConnected = false;
+      deviceManager.pause();
```
grafnu, line 696: Similar to the other comment -- here's pause() -- where's the matching resume()?

validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
```diff
@@ -800,7 +800,7 @@ protected void resetConfig(boolean fullReset) {
       resetDeviceConfig(false);
       updateConfig("soft reset");
       debug("Done with reset_config");
-      resetRequired = false;
+      // resetRequired = false;
```
grafnu, line 803: ?

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
```diff
@@ -51,7 +51,7 @@ public class ConfigSequences extends SequenceBase {
   private static final Duration CONFIG_QUERY_INTERVAL = Duration.ofSeconds(30);
 
   @Test(timeout = ONE_MINUTE_MS)
-  @Feature(stage = STABLE, bucket = SYSTEM, nostate = true)
+  @Feature(stage = STABLE, bucket = SYSTEM)
```
grafnu, line 54: Likely needs a comment, but this "nostate = true" is intentional since I don't want things marked as "nostate" to just blindly pass -- I wanted it on record that they still essentially fail.  Was there a specific reason you needed this, or was it just because it seemed like a good idea (which is a perfectly reasonable assumption!)?

grafnu (commented):
Overall seems on track, I think most/all of the comments are local considerations.

.gitignore
MertCingoz: I am using venv from another location (symlink) to prevent installing python dependencies each time.

pubber/src/main/java/daq/pubber/ProxyDevice.java
MertCingoz: There is no unpause logic, it is handled by config handler same as regular startup.

validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
MertCingoz: I just changed field to non-static. The previous test config doesn't seem to be cleaned up correctly.

I am getting discovery config on next test if I use more then 3 test class.

`sequencer broken_config periodic_scan too_much_state`

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
MertCingoz: It seems either test declaration or expected device behavior is not correct me. It is expecting state message from the device. But it make sense to fail if it is intentional.

pubber/src/main/java/daq/pubber/ManagerBase.java
MertCingoz: It looks ok to me. Not start periodic senders and only cancel when intervalSec == DISABLED_INTERVAL

grafnu: Ah, got it.

pubber/src/main/java/daq/pubber/ManagerBase.java
grafnu: still missing @Override (?)

pubber/src/main/java/daq/pubber/ProxyDevice.java
grafnu: Ah -- in that case I'd call it stop() -- I think pause() has an implicit assumption that there's a matching resume(). If this is just doing the opposite of "start()" (which is actually the configuration here), I'd name this as the peer of that.

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/ConfigSequences.java
grafnu: I talked w/ Nour about this and we think you're right and I'll return it to the way you had it -- I had done it that way because of development testing but it doesn't make sense to keep it like that.  I'll do that in another PR since it really doesn't have anything to do with your changes and requires some other changes to make it pass CI testing.

grafnu (commented):
Looks like there's still some failing unit tests -- let me know if you have questions about what they mean. The tests can be a bit cryptic, but they are easy to fix.

pubber/src/main/java/daq/pubber/ManagerBase.java
MertCingoz: No need to override here as it is abstract base class.

MertCingoz:
Just reverted schema upgrade. Unit tests require manual upgrade of sample site model and trace files which I think it should be also handled in another PR

grafnu (approved):

PR #815, opened 2024-02-22
add registrar testing utility

noursaidi:
The contents of `test.py` is not complete because this is part of something old and not everything in it is working, however this is a utility I use to validate registrar was successful.

Original intent was to programmatically test cloud registrar, by creating random site models, registering them, validating them, and doing that in parralel

The `test_devices` and  `test_git` methods are working, everything else not yet.

### DIFF
--- a/misc/registrar_validation/test.py
+++ b/misc/registrar_validation/test.py
@@ -1,0 +1,364 @@
+# pylint: skip-file
+# because this is an old partial file
+
+from collections.abc import Callable
+import copy
+import datetime
+import glob
+import json
+import json
+import os
+from pathlib import Path
+import random
+import re
+import shutil
+import shutil
+import signal
+import ssl
+import subprocess
+import sys
+import time
+import time
+from typing import Any
+from typing import Iterator
+import jwt
+import paho.mqtt.client as mqtt
+import pytest
+
+KEYS_TO_REDACT = ["timestamp", "version", "operation"]
+REDACTED_VALUE = "redacted"
+
+SITE_PATH = os.getenv("SITE_PATH") if os.getenv("SITE_PATH") else "."
+devices_list = [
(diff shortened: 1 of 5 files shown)


misc/registrar_validation/test.py
```diff
@@ -0,0 +1,364 @@
+# pylint: skip-file
```
grafnu, line 1: Never ever use "test" as a name for anything, unless it's really just a test of something to be deleted.  What does this actually _do_ for the "test"?  Take it one more level and add at least some description into the name!

misc/registrar_validation/test.py
```diff
@@ -0,0 +1,364 @@
...
+    self.configs = {}
+
+    self.client = self.start_client()
+
+  def subscribe_proxy_config(self, proxy_id):
+    self.client.subscribe(f"/devices/{proxy_id}/config", qos=1)
+
+  def attach_device(self, device_id) -> int:
+    """Returns mid"""
+    attach_topic = f"/devices/{device_id}/attach"
```
grafnu, line 73: We should talk about your general approach here, because although it's "quick and dirty" to crank out tools like this, it doesn't contribute as much in the long run to the overall UDMI ecosystem.  In general with software development our goal is to amortize costs in order to solve problems _efficiently_ in the long run, not just crank out "solutions"!

misc/registrar_validation/test.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
```
grafnu, line 1: NB: .sh extension is redundant, so shouldn't be there.  You only should have that if it *needs* to be 'sourced" -- otherwise it could, e.g. be a binary or python or whatever. There is no need to know this is a script.

grafnu (approved):

PR #851, opened 2024-04-08
Add option for setting static config

noursaidi:


tests/sites/mapping/devices/GAT-123/expected/exceptions.txt
```diff
@@ -1,3 +1,6 @@
+Config
+java.lang.NullPointerException: Cannot read field "min_loglevel" because "this.metadata.system" is null
```
grafnu, line 2: This doesn't seem right...

tests/sites/mapping/devices/GAT-123/expected/errors.map
```diff
@@ -1,6 +1,7 @@
 Exceptions for GAT-123
   While loading credentials for local device GAT-123
     Credential cloud.auth_type definition missing
+  Cannot read field "min_loglevel" because "this.metadata.system" is null
```
grafnu, line 4: I think there's something here that might need some addressing... shouldn't just be exporting a NPE...

tests/sites/missing/devices/AHU-22/expected/errors.map
```diff
@@ -1,4 +1,3 @@
 Exceptions for AHU-22
-  While converting device config
-    While converting point filter_differential_pressure_sensor
-      point ref FAXLE12.present_value does not match expression ([A-Z]{2,4})([0-9]+)\.([a-z_]+)
+  While converting point filter_differential_pressure_sensor
```
grafnu, line 2: Do you know why it lost the "converting device config"... seems like it's losing a bit of important context... ?

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -493,9 +502,14 @@ private byte[] getFileBytes(String dataFile) {
...
 
   private String deviceConfigString() {
+    if (configJson == null) {
+      return null;
+    }
+
     try {
-      JsonNode configJson = OBJECT_MAPPER_STRICT.valueToTree(deviceConfigObject());
-      new MessageDowngrader("config", configJson).downgrade(baseVersion);
+      if (config.canBeDowngraded()) {
```
grafnu, line 510: I'd phrase this as "shouldBeDowngraded" -- I think it could (can) be downgraded, it's just that we may not want to.

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
```diff
@@ -882,6 +882,7 @@ private void sendUpdateMessages(LocalDevice localDevice) {
     }
   }
 
+  //TODO this is now broken if device uses a file rather than automatically generated config
```
grafnu, line 885: Space after //, and colon after TODO... so "// TODO: " -- just for consistency.

validator/src/main/java/com/google/daq/mqtt/util/ConfigManager.java
```diff
@@ -32,25 +39,93 @@
...
 
-  public static ConfigGenerator configFrom(Metadata metadata) {
-    return new ConfigGenerator(metadata);
+  /**
+   * Whether the device config be downgraded.
+   *
+   * @return yes/no
+   */
+  public boolean canBeDowngraded() {
+    return ! isStaticConfig();
```
grafnu, line 102: pretty sure lint error?  No space after !

validator/src/main/java/com/google/daq/mqtt/util/ConfigManager.java
```diff
@@ -32,25 +39,93 @@
...
+  /**
+   * Returns a JSON representation of the device config, allowing for any static config overrides.
+   *
+   * @return JSON Object of Device Config
+   */
+  public JsonNode deviceConfigJson() {
+    if (isStaticConfig()) {
+      String staticFileContents = readStaticConfigFromFile(metadata.cloud.config.static_file);
+      try {
+        return OBJECT_MAPPER_STRICT.readTree(staticFileContents);
```
grafnu, line 114: There code in JsonUtils that does most of this.  You should just be able to compute and calcualte the "configFile" and then pass that into the utility function.  The "readFileToString" and "OBJECT_MAPPER_STRICT parts should be encapsulated in JsonUtil already.

tests/sites/mapping/devices/GAT-123/expected/exceptions.txt
noursaidi: I believe this is unchanged behaviour except the exception is now surfaced. It's expected because the metadata file is invalid, so metadata becomes `null`

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
noursaidi: Should these bits  be removed or fixed to work with generic JSON types? because they also won't work with any downgraded messages I think. They're also redundant now that UDMIS handles the actual config rather than direct through the IoT Core API

grafnu:
Ok -- I think it's fine in the exceptions.txt file, but we should capture
and hopefully make it more understandable for the errors.map file -- if
it's not clear how to do that (it shouldn't be hard) let me know and I can
poke at it.  Sometimes I'm wrong about "it should be easy"

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
noursaidi: Changed

validator/src/main/java/com/google/daq/mqtt/util/ConfigManager.java
noursaidi: Changed

tests/sites/missing/devices/AHU-22/expected/errors.map
noursaidi: Restored this behavior. I had a new config generation function which omitted this text whereas the old did.

grafnu (approved):

PR #864, opened 2024-04-19
Fix message upgrade and config downgrades

noursaidi:


common/src/main/java/com/google/udmi/util/MessageDowngrader.java
```diff
@@ -103,43 +102,114 @@ private Map<String, Object> downgradeState(JsonNode targetVersion) {
...
+  public boolean wasDowngraded() {
+    return !original.equals(message);
+  }
+
+  private void downgradeConfigRaw(String targetVersionString){
+    final SchemaVersion targetVersionEnum = SchemaVersion.fromKey(targetVersionString);
+    final Integer targetVersion = targetVersionEnum.value();
+
+    // downgrade to 1.4.1
+    if (currentVersion() > VERSION_1_4_1.value() && currentVersion() > targetVersion){
```
grafnu, line 130: If we wanted to be really smarmy here we could make a table of sorts that had { versionA, versionB, downgradeFunc, upgradeFun } in it and then all this if statement stuff would be automagic!

Not something I'd suggest for now, but just something to keep in mind for the rainy day in the future when you're bored.

common/src/main/java/com/google/udmi/util/MessageUpgrader.java
```diff
@@ -127,42 +123,47 @@ private Object upgradeRaw(boolean forceUpgrade) {
...
       upgraded |= !before.equals(message);
       patch = 1;
     }
 
     if (upgraded && message.get(VERSION_KEY) != null) {
       message.put(UPGRADED_FROM, originalVersion);
       message.put(VERSION_KEY, String.format(TARGET_FORMAT, major, minor, patch));
+    } else {
+      // Files are now updated to Current version of UDMI
+      if(message.has(VERSION_KEY)){
```
grafnu, line 136: lint won't like this?  Should have space after "if"

Also, comment is unclear or superfluous -- it's not particularly clarifying the why of anything?

common/src/main/java/com/google/udmi/util/MessageUpgrader.java
```diff
@@ -173,20 +174,66 @@ private void upgrade_1_3_14_metadata() {
...
+    }
+    if (METADATA_SCHEMA.equals(schemaName)) {
+      upgradeTo_1_4_1_metadata();
+    }
+  }
+
+  private void upgradeTo_1_5_0_metadata() {
+
+  }
+  private void upgradeTo_1_4_1_metadata() {
```
grafnu, line 192: spaces 'n' stuff

grafnu (approved):
Just one minor real comment and then a few nits and pontifications

common/src/main/java/com/google/udmi/util/SchemaVersion.java
```diff
@@ -12,9 +12,16 @@
 public enum SchemaVersion {
   VERSION_1_5_0("1.5.0", 105000),
   VERSION_1_4_2("1.4.2", 104020),
-  VERSION_1_4_0("1.4.0", 104000);
+  VERSION_1_4_1("1.4.1", 104010),
```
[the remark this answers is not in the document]
noursaidi: You're right - I just followed the existing numbering. They all need changing

grafnu:
There should be another field in the "upgraded_from" to capture the
needful. I agree with the logic that the end result should be consistent
(on current version) even if the source was not changed.

***@***.***:~/udmi$ fgrep upgraded_from -r schema/
schema/events_pointset.json:    "upgraded_from": {
schema/metadata.json:    "upgraded_from": {
schema/state_pointset.json:    "upgraded_from": {
schema/state.json:    "upgraded_from": {
schema/state_system.json:    "upgraded_from": {
schema/events_system.json:    "upgraded_from": {

PR #881, opened 2024-05-07
Fix publisher executor shutdown state

MertCingoz:


pubber/src/main/java/daq/pubber/MqttPublisher.java
```diff
@@ -163,6 +163,9 @@ public void publish(String deviceId, String topicSuffix, Object data, Runnable c
     Object marked =
         topicSuffix.startsWith(EVENT_MARK_PREFIX) ? decorateMessage(topicSuffix, data) : data;
     try {
+      if (!isActive()) {
+        publisherExecutor = Executors.newFixedThreadPool(PUBLISH_THREAD_COUNT);
```
grafnu, line 167: Overall, I think it would just be cleaner to just not shutdown the executor in the close() method. The core of the problem seems to be that close() gets called from non-shutdown situations (like connection lost), so I'd rather keep the changes isolated to that part of the code rather than having the infection spread...

MertCingoz: Shutdown and close logic separated. Seems to be periodic senders failing to publish messages when connection lost but it's not handled by Pubber or corresponding manger. I think we also need to consider triggering reconnect attempt at some point (errorCounter maybe?) so that we can use backoff logic when trying to reconnect.

Currently all proxy devices and gateway trying to publish periodic events and ignoring network problems. This creates error flood on each publish attempt.

grafnu (commented):
Overall technically this looks good, but what kind of testing have you been able to do on this?  I took a quick look at your GitHub and it doesn't seem to be setup to run anything, and the upstream faucetsdn project doesn't run everything on a PR (security reasons, mainly)...

grafnu:
There's the "always runs" CI setup at
https://github.com/faucetsdn/udmi/blob/master/.github/workflows/testing.yml#L249
-- that'll show you the steps the CI testing does to setup a local run.
there's a bin/setup_local script that does the needful.

You can also set your TARGET_PROJET github repo actions variable to
//mqtt/localhost and it should work, as per
https://github.com/grafnu/udmi/settings/variables/actions (if you can see
that, not sure)

There's docs for containerized docker setup --
https://github.com/faucetsdn/udmi/blob/master/docs/udmis/local_docker.md
(hot off the press, just did that yesterday) -- but that doesn't directly
do CI testing yet... that's next.

MertCingoz:
I guess my Github username is not docker registry friendly.
https://github.com/MertCingoz/udmi/actions/runs/10042673525/job/27753661849

grafnu (approved):
LGTM.  Reminder that for merging with master please remove all commit comments (just the PR title) when you squash-commit... the actual commit log (which is the default GitHub fills in) is pretty useless and just pollutes the commit log!

PR #885, opened 2024-05-09
add metadata fields for floor_seq, connection_type, tags

noursaidi:


schema/model_pointset_point.json
```diff
@@ -42,6 +42,16 @@
...
+    },
+    "translation": {
+      "description": "An ontological translation for the point",
+      "type": "string",
+      "examples": ["return_air_temperature_sensor"]
+    },
+    "tags": {
+      "description": "Descriptor for the point",
+      "type": "string",
+      "examples": ["lighting", "lighting,building"]
```
grafnu, line 54: Array of strings seems better than a singular CSV string?  Also consider a (restricted) regex for this (e.g. only [a-z] for actual tag names)... ?  I generally fret about fields like this because they end up getting conscripted to do nefarious things.  Also, what's the future proofing here?  singular string tags are nice but also very limiting, and typically the NEXT thing that people want to do is key=value pairs, so the question comes down to "should we just start with key=value pairs and skip over tags completely" -- rather than support two structures in the future.

schema/model_pointset_point.json
```diff
@@ -42,6 +42,16 @@
       "description": "Mapping for the point to an internal resource (e.g. BACnet object reference)",
       "type": "string",
       "examples": ["AI3", "400070"]
+    },
+    "translation": {
```
grafnu, line 46: erm, not comfortable with this ATM since there's already the concept of translations in the gateway config schema and other places, so I'd want to make sure this is coherent with the overall structure of things.

schema/model_pointset_point.json
grafnu: after our quick chat I understand "why not k=v", but then the question remains about "why a CSV rather than a list/array"

schema/model_pointset_point.json
noursaidi: Changed `translation` into `mapping` and an object type, with one property `name`.

schema/model_pointset_point.json
noursaidi: Changed into a list of alphanumeric string values

schema/metadata.json
```diff
@@ -40,9 +40,13 @@
...
     },
     "tags":{
-      "description": "Free from text entry for a tag associated with the device",
-      "type": "string",
-      "examples": ["green", "electric,total,lighting,building"]
+      "description": "Tags assosciated with the device",
+      "type": "array",
+      "uniqueItems": true,
+      "items": {
+        "pattern": "^[A-Za-z0-9]+$"
```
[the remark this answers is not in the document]
grafnu: didn't we think about excluding A-Z for now?

grafnu (approved):

PR #897, opened 2024-05-24
use site metadata in registrar and publish to udmi_target

noursaidi:


noursaidi:
https://github.com/noursaidi/udmi/actions/runs/9352979939

grafnu (approved):

PR #921, opened 2024-06-27
Bump up version to 1.5.1

noursaidi:


grafnu (approved):
I wish we had a better way of doing this that didn't involve changes to 220 different files! But, I think that's just the result of the schema testing infrastructure...

PR #922, opened 2024-06-27
Promote set of tests to stable

noursaidi:


etc/schema_itemized.out
```diff
@@ -4,23 +4,23 @@
 04 pointset_remove_point
 05 too_much_state
 06 feature_enumeration
-07 valid_serial_no RESULT pass schemas device_state_beta BETA 5/5 Schema validation passed
-07 valid_serial_no RESULT pass schemas events_system_beta BETA 5/5 Schema validation passed
-07 valid_serial_no RESULT pass schemas state_update_beta BETA 5/5 Schema validation passed
+07 valid_serial_no RESULT pass schemas device_state_stable 5/5 Schema validation passed
```
grafnu, line 7: This doesn't seem right -- shouldn't there be a STABLE in there?

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/PointsetSequences.java
```diff
@@ -105,9 +107,9 @@ private boolean validPointEntry(Entry<String, PointPointsetEvents> point) {
     return point.getValue().present_value != null;
   }
 
-  @Test(timeout = TWO_MINUTES_MS)
+  @Test(timeout = SIX_MINUTES_MS)
```
grafnu, line 110: erm? SIX minutes?

validator/src/main/java/com/google/daq/mqtt/util/TimePeriodConstants.java
```diff
@@ -9,4 +9,5 @@ public class TimePeriodConstants {
   public static final int TWO_MINUTES_MS = 2 * 60 * 1000;
   public static final int THREE_MINUTES_MS = 3 * 60 * 1000;
   public static final int FOUR_MINUTES_MS = 4 * 60 * 1000;
+  public static final int SIX_MINUTES_MS = 6 * 60 * 1000;
```
grafnu, line 12: why this?  something's not right...

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/PointsetSequences.java
noursaidi: Following earlier suggestions that this test should support a device which publishes telemetry every five minutes, so six minutes allows the 5 minutes plus setup/teardown.

I just remembered as I was changing these bits, can leave it out of this PR

grafnu:
Yah -- I think for now I'd prefer to leave that out, I forget that
conversation. I think the problem is that this is a pretty blanket change
(all devices), while really we likely need some way for that device to
explicitly declare it's a bad actor.

grafnu (approved):

PR #925, opened 2024-07-09
remove unused dependancy hxtools

noursaidi:


grafnu (approved):
NB for something trivial/mechanical like this, no need for a review!

PR #933, opened 2024-07-19
Create reset_config.md

pisuke:
Documentation for `reset_config`

### DIFF
--- a/docs/tools/reset_config.md
+++ b/docs/tools/reset_config.md
@@ -1,4 +1,6 @@
-## Send a configuration message to a specific device
+[**UDMI**](../../) / [**Docs**](../) / [**Tools**](./) / [Reset Config](#)
+
+# Send a configuration message to a specific device
 
 The `reset_config` tool can be used to send a configuration message to a specific device.
 
--- a/docs/tools/reset_config.md
+++ b/docs/tools/reset_config.md
@@ -1,0 +1,44 @@
+## Send a configuration message to a specific device
+
+The `reset_config` tool can be used to send a configuration message to a specific device.
+
+The `reset_config` message can be invoked from the `udmi` folder in the following way:
+
+```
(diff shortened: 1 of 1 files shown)


docs/tools/reset_config.md
```diff
@@ -0,0 +1,44 @@
+## Send a configuration message to a specific device
+
+The `reset_config` tool can be used to send a configuration message to a specific device.
+
+The `reset_config` message can be invoked from the `udmi` folder in the following way:
```
grafnu, line 5: There's no "reset config" message as such -- it's a standard "config update" message that is used to reset the config. So I'd change this to just "command can be invoked." The distinction is important (technical documentation) because otherwise people will go off looking for a specific message type (which doesn't exist so they'd never find it!).

docs/tools/reset_config.md
```diff
@@ -0,0 +1,44 @@
...
+In this case, it is necessary to store a file including the `config` payload in the `devices/DEVICE_ID/config` folder.
+
+For example, if a file named `config.json` has been created for this purpose in the `devices/TEST-1/config` folder,
+it will be possible to invoke the `reset_config` tool with the following command:
+
+```
+bin/reset_config path/to/site_model my_project TEST-1 config.json
+```
+
+Finally, it is possible to get `registrar` to send a user defined config instead of a generated config file.
```
grafnu, line 32: NIT: This is going to be confusing at some point since this isn't about "reset_config" anymore -- ideally there'd be a separate static_file documentation set somewhere, but doesn't seem to currently exist!  Anyway, just FYI -- this is no worse than some of the other documentation we have so can't complain too much :-)  @noursaidi

pisuke: Cool thank you Trevor

docs/tools/reset_config.md
pisuke: Change done as suggested.

grafnu (approved):

PR #935, opened 2024-07-22
Addition of new keys in site_metadata.json

pisuke:
Addition of new keys in `site_metadata.json` for additional metering and fleet management use cases and to support BAMBI (Bos Automated Metering Building Interface)


schema/site_metadata.json
```diff
@@ -50,6 +58,42 @@
...
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "m2": {
+            "description": "Area of the site in square meters",
+            "type": "number",
+            "examples": [1234, 256, 998765]
+        },
+        "sqf": {
+            "description": "Area of the site in square feet",
```
grafnu, line 72: Redundancy is usually a bad idea if there's really only one thing, because it only means that somebody can get it wrong. Why two fields?  I'd go with one field with a units -- otherwise you're implying that the m2 could be different than sqf...

schema/site_metadata.json
```diff
@@ -50,6 +58,42 @@
...
+      "description": "Folder where the UDMI site model is stored",
+      "type": "string",
+      "examples": ["/home/user/udmi_site_model", "https://drive.google.com/corp/drive/folders/iamadrivelocation"]
+    },
+    "source_repo": {
+      "description": "Source repository where the UDMI site model is stored",
+      "type": "string",
+      "examples": ["https://github.com/faucetsdn/udmi_site_model", "git@github.com:faucetsdn/udmi_site_model.git"]
+    },
+    "device_count": {
```
grafnu, line 88: i'm wondering if we should generalize this a bit into a bucket, just because I can see the utility it some other counts... e.g. registered devices, discovered-but-not-in-model devices, etc...   I was just thinking maybe put these into a general `count` bucket?

```
count.modeled
count.validated
count.registered
count,discovered
```

(related, `device_count` isn't maybe the best name since they're all _device_ counts of various sorts... it's just the adjective that's describing what _kind_ of device it is -- e.g. modeled devices vs. registered devices, so I'd suggest `modeled` or `total` maybe as the adjective instead of `device`)

schema/site_metadata.json
```diff
@@ -50,6 +58,42 @@
...
+            "examples": [1234, 256, 998765]
+        },
+        "sqf": {
+            "description": "Area of the site in square feet",
+            "type": "number",
+            "examples": [13282.67, 2755.56, 10750617]
+        }
+      }
+    },
+    "site_folder": {
```
grafnu, line 78: maybe bucket these into a `source` object?  There's also other things that come up, e.g. the branch, folder, potentially a specific commit. And eventually there might be other mechanisms (e.g. that's not a git repo), so then it'd be a place to contain a type-of-source enum, etc... ?

```
source.folder
source.repo
source.????
```

schema/site_metadata.json
pisuke: We need to consume this information in separate formats. The logic to ensure the information comes from a single field will be in the data entry tool (spreadsheet based, locked field+formula)

schema/site_metadata.json
pisuke: LGTM, happy to update it if there is agreement.

schema/site_metadata.json
pisuke: LGTM, I'll update it.

grafnu:
There should be a presentation/adaptation layer between the underlying DB
structure and user-facing systems -- it doesn't work well to link the two
closely together because people and machines interact with data
differently, and often you want to iterate on one without affecting the
other. On the backend (so in the site_model) there's really only one value
here -- the area of the building -- and then the presentation layer can
translate between them or do whatever it is that people want. If you
already have front-end logic to ensure that the information is
soured/synced from a single data-value.

schema/site_metadata.json
pisuke: Done

pisuke:
> There should be a presentation/adaptation layer between the underlying DB structure and user-facing systems -- it doesn't work well to link the two closely together because people and machines interact with data differently, and often you want to iterate on one without affecting the other. On the backend (so in the site_model) there's really only one value here -- the area of the building -- and then the presentation layer can translate between them or do whatever it is that people want. If you already have front-end logic to ensure that the information is soured/synced from a single data-value.

I've now addressed this, can you please have another look? Thanks!

schema/site_metadata.json
```diff
@@ -32,6 +31,15 @@
       "type": "string",
       "examples": ["Cheesegrater", "Mountain View"]
     },
+    "tags":{
+      "description": "Tags associated with the site",
+      "type": "array",
+      "uniqueItems": true,
+      "items": {
+        "pattern": "^[a-z0-9]+$"
```
grafnu, line 39: shouldn't this pattern have an _ in it?

schema/site_metadata.json
```diff
@@ -50,6 +58,59 @@
...
+      "additionalProperties": false,
+      "properties": {
+        "m2": {
+            "description": "Area of the site in square meters",
+            "type": "number",
+            "examples": [1234, 256, 998765]
+        }
+      }
+    },
+    "source": {
```
grafnu, line 73: "source" doesn't feel right here -- can't remember our conversation about it. Really the bucket seems to be about "external links to places where information is stored" -- and could be used (e.g.) as an _output_ for some of the tools about the side (so definitely not source). Does something like "externals" or "links" make more sense?

e.g. consider the use-case of "where running the registrar tool should output a log of what it does so it can be captured in (e.g.) a Google sheet?"

schema/site_metadata.json
```diff
@@ -50,6 +58,59 @@
...
+          "examples": ["/home/user/udmi_site_model", "https://drive.google.com/corp/drive/folders/iamadrivelocation"]
+        },
+        "repo": {
+          "description": "Source repository where the UDMI site model is stored",
+          "type": "string",
+          "examples": ["https://github.com/faucetsdn/udmi_site_model", "git@github.com:faucetsdn/udmi_site_model.git"]
+        }
+      }
+    },
+    "count": {
```
grafnu, line 89: "counts" (plural, since it's a collection of things) -- also -- do we want to scope this to "device_counts"? (not sure what other kinds of counts we might have).

schema/site_metadata.json
```diff
@@ -50,6 +58,59 @@
...
+      }
+    },
+    "source": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "folder": {
+          "description": "Folder where the UDMI site model is stored",
+          "type": "string",
+          "examples": ["/home/user/udmi_site_model", "https://drive.google.com/corp/drive/folders/iamadrivelocation"]
```
grafnu, line 80: NB: We should figure out the right canonical syntax to this... e.g. should it be file://home/user/udmi_site_model for a file?

schema/site_metadata.json
```diff
@@ -50,6 +58,59 @@
             "type": "number"
         }
       }
+    },
+    "area": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "m2": {
```
grafnu, line 66: I don't think this formulation here is quite right -- why would you ever have more than one specification for area?  This should either be one field with a canonical type ("area_m2") or two separate fields "area_units" and "area_value" or something... Or just "area" and we can default the units to m2 without an explicit field, and then add in a "area_units" or something like that later.

The goal here is to prevent potential ambiguity in the source specification!

schema/site_metadata.json
```diff
@@ -39,15 +48,211 @@
...
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "value": {
+                  "description": "Net Occupied Area of the site",
+                  "type": "number",
+                  "examples": [1234, 256, 997765]
+                },
+                "unit": {
+                  "description": "Unit used for carbon the Net Occupied Area",
```
grafnu, line 135: description seems a bit off -- "for carbon the" seems like it's missing a word or two!

schema/site_metadata.json
```diff
@@ -39,15 +48,211 @@
...
+                  "type": "string",
+                  "examples": ["m2", "sqf"]
```
grafnu, line 137: NB: Not sure if we want to be more picky about this... e.g. an enum?  I'm just worried about the creep where people to something like "f^2" as a unit rather than "sqf" -- or we just need to plan on some other validation step (if/when) this becomes an issue.  Also possible (I think) to impose a regex on this like "(m2|sqf)" which will at least limit the initial options without the pain of an enum.

noursaidi:
`counts` has been removed and a common parameter object has been created. @grafnu can PTAL?

Ignore gencode not running, I can't git commit when I should be able to, but it will let me edit files through the web

schema/parameter.json
```diff
@@ -0,0 +1,25 @@
+{
+    "$udmi_version": "1.5.1",
+    "title": "Parameter",
```
grafnu, line 3: "parameter" is a bit too generic here... I'd suggest maybe "measurement" -- https://english.stackexchange.com/questions/111291/what-is-a-name-for-a-unit-of-measure-and-value -- things like "dimension" or "quantity" also crop up, but usually in conjunction with something like a "dimensional quantity" which can get unwieldy. I just know "parameter" is bad.

schema/parameter.json
```diff
@@ -0,0 +1,25 @@
...
+      "value": {
+        "type": "number"
+      },
+      "unit": {
+        "type": "string",
+        "oneOf": [
+            {"title": "m2", "pattern": "^m2$"},
+            {"title": "sqf", "pattern": "^sqf$"},
+            {"title": "kgCO2e/kWh", "pattern": "^kgCO2e\/kWh"},
+            {"title": "£/m3", "pattern": "^£\/m3$"},
```
grafnu, line 18: This is going to be fun... not sure how £ plays out with all the code :-P

noursaidi:
https://github.com/faucetsdn/udmi/pull/953/

PR #946, opened 2024-08-22
Update device_testing.md

humplik:
Changing "noState" to "nostate"


humplik:
Changing "noState" to "nostate"

PR #950, opened 2024-09-02
add new fields to device metadata

pisuke:
Addition of new fields to device metadata to support the metering program


schema/model_system.json
```diff
@@ -10,8 +10,36 @@
       "description": "Generic human-readable text describing the device",
       "type": "string"
     },
+    "device_version": {
+      "description": "Version of schema supported by the device",
+      "type": "string",
+      "examples": ["1.5.0"]
+    },
+    "device_type": {
```
grafnu, line 18: "type" is way to generic to be using here... "device_type" means too many things. What is this really describing?  Is it "the primary mechanism by which this device is connected into the system"?  Or, "how this device is connected to its _parent_ (next field)"?

Overall this and the following-field are really starting to smell like a description of their local networks, so we should at least consider what it would mean to move them to that subsection.  E.g., "IP" vs. "serial" is a network designation.

noursaidi:
```mermaid
flowchart LR
    Meter--[modbus/rs232]--> Protocol_Translator--[??]--> Headend--[bacnet]-->IoT_Core_Gateway--[mqtt]--> IoT_Core
```

If we are describing the meter, it's parent is Protocol_Translator and it's type is `serial`

grafnu: Ok -- but my point is that you're specifically describing the local network topology of the device, so we should properly consider putting this in the localnet block, not in the system block. It's not really describing the device itself, but rather how it's connected to other things.  But, my other point is that device_type is a bad bad name.

noursaidi: Okay, `localnet.ip = true/false` or something? If I recall correctly the key requirement is to capture whether a device is IP or not.  I believe using the `localnet.families` was not suitable because that required knowing all the addrs, which may not be easily known.

grafnu: Rule #6 is never use booleans (they always seem like a good idea at the time but in the end they aren't). In this case, I'd would in fact suggest we encode it using the same family designation as elsewhere -- so there'd be something like `localnet.parent.family = "serial"` and then `localnet.parent.id = "Protocol_Translator"` be the _id_ of the parent. I don't think we need to use the address for the parent, but rather an internal reference to the ID (aka the IoT address).

noursaidi: I like the localnet.parent suggestion, but I'm not sure about the use of family as a property. Family would imply one of the existing families such as bacnet,ipv4,ethmac, etc, which wouldn't support the intent here to classify things into ip/serial/virtual

How about localnet.parent.connectivity, or protocol? Or dare I mention connectivity_type?

grafnu:
It's meant to be a broad designator for "how one thing is connected to
another thing" -- which is essentially what networks are. Here it's just
describing a more specific parent relationship, but fundamentally not
really different from how a gateway is connected to a proxy device -- just
the inverse.  They're both essentially a hierarchical relationship of
connectivity. IP and SERIAL are both address families for sure, and we were
already talking about "virtual" in the concept of a gateway with a virtual
device.  Not quite sure what a "group" would be, since IP and SERIAL can
have groups -- so might need more understanding of what that was supposed
to be.

I do think the best way to think about it might be as the reciprocal of the
existing gateway/target structure...

schema/model_system.json
pisuke: Let's remove the `device_type` entry for now.
Let's move `parent` to `localnet.parent.*`

grafnu (approved):
I pushed a change renaming "id" to "target" to avoid super generic field names. If we're good with that (tests are fine) then we can submit.  I also added a simple static JSON test just so we have an example of it somewhere.

PR #962, opened 2024-09-13
fix gencode for site_metadata

noursaidi:
Incorrect types for some objects in generated JAVA class for `site_metadata`


schema/site_metadata.json
```diff
@@ -28,6 +28,7 @@
       "examples": ["UK-LON-ABC", "US-NYC-DEF2"]
     },
     "name": {
+      "title": "",
```
pisuke, line 31: Can the title of this be Site Name?

pisuke (changes requested):

PR #964, opened 2024-09-18
addition of danfoss and vayandata results

pisuke:
Addition of Danfoss and VayanData results and some minor tweaks


grafnu (approved):
I'd say in the future that you don't need approval for these kinds of changes!

grafnu (approved):

PR #988, opened 2024-10-11
Add discovery node example with CI test for automapper

noursaidi:
Example of CI test - https://github.com/noursaidi/udmi/actions/runs/11298671275/job/31428114397


.github/workflows/testing.yml
```diff
@@ -355,3 +355,28 @@ jobs:
       - name: Itemized validation
         if: ${{ !cancelled() }}
         run: bin/test_itemcheck
+
+  discoverynode:
+    name: Discovery Node
+    if: ${{ !cancelled() && vars.TARGET_PROJECT != '' }}
+    needs: [ baseline, sequencer, redirect, runlocal ]
```
grafnu, line 362: should get this running local-only for now, to prevent making the overall CI test time too long. Alternatively we can fork out another parallel registry for this (and one of the other three).   (And, this needs: clause is unnecessarily broad -- figure out where it goes in the chain and adjust that accordingly)

For now, I'd be following the pattern of the runlocal: job (which always runs and only uses local testing)

.github/workflows/testing.yml
```diff
@@ -355,3 +355,28 @@ jobs:
...
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+      - name: install pytest
+        run: |
+          pip install pytest
```
grafnu, line 375: Can/should this just be folded into bin/setup_base?  Seems like it would be simpler overall...

misc/discoverynode/README.md
```diff
@@ -8,5 +8,13 @@
 `~/venv/bin/python3 -m pytest tests/`
 
 ## Integration test (no MQTT)
-`sudo testing/integration/integration_test.sh`
+
+```
+sudo testing/integration/integration_test.sh
+
+sudo docker ps -a | grep "discoverynode-test-" | awk '{print $1}' | sudo xargs docker stop
```
grafnu, line 15: just trying to understand this -- it's stopping a different container than it starts?

misc/discoverynode/README.md
```diff
@@ -8,5 +8,13 @@
...
+
+docker run --network=test-network -v  $PWD/dump:/data --rm -ti travelping/pcap
```
grafnu, line 17: Maybe add a specific container name?

You don't need "-ti" if there's no terminal attached.

misc/discoverynode/src/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+echo hello from entry point
+TIMESTAMP=$(date +%s)
```
grafnu, line 2: Usually I use `date -Is -u` to give a standardized format:

`2024-10-13T16:06:56+00:00`

misc/discoverynode/src/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
...
+echo $TIMESTAMP
+tcpdump -ni eth0  -w /pcap/$TIMESTAMP.pcap &
```
grafnu, line 4: extra space after eth0

misc/discoverynode/src/sample_bacnet.py
```diff
@@ -0,0 +1,51 @@
+<<<<<<< HEAD
+"""Sample for calling BACnet scan directly"""
+=======
+>>>>>>> 82d5a00d (wip)
```
grafnu, line 4: this can't be right :-)

misc/discoverynode/src/tests/test_integration.py
```diff
@@ -29,30 +33,36 @@ def test_bacnet_integration():
...
-        m.families["ethmac"].addr for m in messages if "ethmac" in m.families
-    ) >= set(d["ethmac"] for d in docker_config.values())
+    expected_bacnet_ids = set(str(d["bacnet_id"]) for d in docker_config.values())
+    seen_bacnet_ids_toplevel = set(m.scan_addr for m in messages if m.scan_family == "bacnet")
 
-    assert set(
-        m.scan_addr for m in messages if m.scan_family == "bacnet"
-    ) == set(d["bacnet_id"] for d in docker_config.values())
+    # subset because passive scan will find the gateway and device itself
+    assert seen_ethmac_toplevel >= expected_ethmacs
```
grafnu, line 67: if you know exactly what it's doing, can you do `== (expected_ethmacs + 2)`?

misc/discoverynode/src/udmi/core.py
```diff
@@ -19,10 +21,10 @@
 class UDMI:
```
grafnu, line 21: seems like this should be `class core:` to match the filename -- or filename to match? udmi_core?  Just `UDMI` seems a bit problematic because at some point it might conflict with the actual UDMI package/class...

misc/discoverynode/src/udmi/discovery/bacnet.py
```diff
@@ -60,44 +60,46 @@ def stop_discovery(self):
...
-          event.system.hardware.make = vendor_name
-          event.system.hardware.model = model_name
-          event.system.software.firmware = firmware_version
-
-          self.publisher(event)
-          self.devices_published.add(device)
-
-        if self.cancelled:
-          return
+          # defaultdict(<class 'int'>, {('192.168.8.2', 184): 1, ('192.168.8.3', 131): 1})
```
grafnu, line 71: What this trying to say?   If it's not explaining _why_ it's doing something, then likely the comment isn't useful...

misc/discoverynode/src/udmi/discovery/discovery.py
```diff
@@ -66,16 +76,16 @@ class states(enum.StrEnum):
 
   # Discovery has finished the scan
   FINISHED = "finished"
-
+
```
grafnu, line 79: tws

misc/discoverynode/testing/e2e/test.sh
grafnu, line 1: please never call something just `test` it ends up being unnecessarily confusing!  Seems like this should be just `e2e_test` (to match the py file?). Also, the trailing .sh isn't necessary (should omit) since it could theoretically be anything (it's an executable)

noursaidi:
Example run - https://github.com/noursaidi/udmi/actions/runs/11517418899/job/32062058683

misc/discoverynode/testing/e2e/test.sh
noursaidi: renamed and extension removed

misc/discoverynode/src/tests/test_integration.py
noursaidi: Yes!

.github/workflows/testing.yml
noursaidi: Good point, it was already in setup_base

.github/workflows/testing.yml
noursaidi: Updated to use local UDMIs for testing

misc/discoverynode/src/udmi/core.py
grafnu: Thoughts on this?

.github/workflows/testing.yml
```diff
@@ -355,3 +355,48 @@ jobs:
       - name: Itemized validation
         if: ${{ !cancelled() }}
         run: bin/test_itemcheck
+
+  discovery:
+    name: Discovery Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    strategy:
```
grafnu, line 363: fail-fast not necessary b/c no matirx: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

grafnu (approved):

PR #993, opened 2024-10-18
Refactor Pubber to simplify downstream vendor modifications

erik-radixiot:
## Description

When upgrading from upstream latest Pubber version we noticed that it was pretty difficult to adapt to vendor needs, which requires a lot of diff and patching over two commits, with this changes we extracted some common methods into interfaces which will simplify the upgrade process for downstream upgrades to latest upstream UDMI.

## Changes
* Introduced provider interfaces in the `.client` package for every Manager with common methods extracted from the actual Managers that are shared among downstream vendor implementations. 
* Added some new getter and setter methods for private attributes shared with downstream vendor implementation.
* Modified private fields to use interfaces instead of actual implementation of managers.
* Moved some constants from the Manager classes to its parent interface.
* Adapted test for these changes. 
* Fixed default socket factory when no TLS connect is specified.


erik-radixiot:
@grafnu requested changes are ready, please check.

grafnu (approved):

PR #1002, opened 2024-10-25
Add more mapper test scenarios

noursaidi:


grafnu:
Ok, to be fair this was waiting my review. But -- my review is that I don't understand what it's trying to do (but I suppose that doesn't matter since it's in misc/). Either way I'd like to get it off the books so if you still want to submit we can, otherwise we can just let it fade away into a distant memory.

PR #1010, opened 2024-10-30
Add documentation and daemon installer for discovery node

noursaidi:


misc/discoverynode/.gitignore
```diff
@@ -0,0 +1 @@
+venv/*
```
grafnu, line 1: just venv/ -- no need for the trailing star

misc/discoverynode/README.md
```diff
@@ -1,8 +1,121 @@
-# Discovery
+# Discovery Node
 
-## UDMI MQTT Client
+## Notes
 
-`sudo ~/venv/bin/python3 main.py`
+`vendor` discovery family is a counter which increments every second
```
grafnu, line 5: This doesn't make sense?

misc/discoverynode/README.md
```diff
@@ -1,8 +1,121 @@
...
+        "bacnet": {"ip": "192.168.11.251"}
+    }
+    ```
+
+    Local UDMIS (complete)
+
+    ```json
+    {
+        "mqtt": {
+        "device_id": "GAT-1",
```
grafnu, line 45: indent

misc/discoverynode/README.md
```diff
@@ -1,8 +1,121 @@
...
+    venv/bin/python3 -m pip install -r src/requirements.txt
+    ```
+
+2. Create a JSON configuration file
+
+    Sample Clearblade configuration file
+
+    ```json
+    {
+        "mqtt": {
```
grafnu, line 25: It would be dandy if this bit matched the schema of schema/configuration_endpoint.json -- but, we can track that with some future work.  That's fundamentally what this is...

grafnu (approved):

misc/discoverynode/README.md
noursaidi: Yup, I do agree with that and something I plan to do

misc/discoverynode/README.md
noursaidi: It's just a note to say that if the discovery node does a `vendor` discovery is actually a counter

```python
 for i in itertools.count(1):
      if self.cancelled:
        return
      result = DiscoveryEvent(
          generation=self.generation, scan_family=self.scan_family, scan_addr=i
      )
      self.publisher(result)
      time.sleep(1)
```

PR #1011, opened 2024-10-30
add range_min and range_max

noursaidi:


schema/model_pointset_point.json
```diff
@@ -33,6 +33,14 @@
         "failure"
       ]
     },
+    "range_min": {
+      "description": "Represents the lower bound of the error threshold for a point",
+      "examples": [0]
```
grafnu, line 38: type?  should be number?

noursaidi: yup, fixed

grafnu: Did you commit/push all changes?  The schema/ file isn't showing that it's been changed, but the gencode/ file is updated... ??!?

noursaidi: I missed the schema file. Committed now.

grafnu (approved):

schema/model_pointset_point.json
grafnu: The irony is that with a small PR I'd notice something like that, but with a big PR I wouldn't!

Approved.

PR #1025, opened 2024-11-14
Minor fixes to discovery node in misc

noursaidi:
- better logging
-  ethmac -> ether
- include firmware version in bacnet


noursaidi:
merging because misc directory

PR #1032, opened 2024-11-22
Update Command Line Options to use arg_name, not arg_type

KhyatiMahendru:


common/src/main/java/com/google/udmi/util/CommandLineOption.java
```diff
@@ -33,4 +33,9 @@
    * Set the argument type description.
    */
   String arg_type() default NO_VALUE;
```
grafnu, line 35: I would deprecate/remove arg_type() here since there's no current use/need for it.

grafnu (approved):
One small change to make, and then it all looks good to go!

When submitting, the practice we follow is to remove all of the PR comments, so there's no commit history included (which is the default GitHub behavior), and rather just rely on the PR# link of the title.  If you take a look at the git log you'll see what I mean!

common/src/main/java/com/google/udmi/util/CommandLineOption.java
KhyatiMahendru: This is done

KhyatiMahendru:
> When submitting, the practice we follow is to remove all of the PR comments, so there's no commit history included (which is the default GitHub behavior), and rather just rely on the PR# link of the title. If you take a look at the git log you'll see what I mean!

By this, do you mean I should "squash and merge"?

grafnu:
> > When submitting, the practice we follow is to remove all of the PR comments, so there's no commit history included (which is the default GitHub behavior), and rather just rely on the PR# link of the title. If you take a look at the git log you'll see what I mean!
> 
> By this, do you mean I should "squash and merge"?

Yes -- once you start "Squash and Merge" there's a box that'll have the commit history in it -- delete that so there's nothing in the box -- then the Title of the PR is the record of what it was.  Sometimes if it's really important then we can put something in the box, but not just "because"!

PR #1035, opened 2024-11-25
make cloud registrar work again

noursaidi:


validator/Dockerfile.validator
```diff
@@ -1,4 +1,5 @@
-FROM alpine:latest
+#FROM alpine:latest
```
grafnu, line 1: remove comment (#), or add more words explaining why I might care about it...

grafnu (approved):

PR #1041, opened 2024-12-04
Adding complete protocol family (e.g. from BACnet discovery) modeling

grafnu:
… databases


tests/schemas/metadata/bacmodel.json
```diff
@@ -0,0 +1,80 @@
...
+    }
+  },
+  "cloud": {
+    "auth_type": "ES256",
+    "connection_type": "DIRECT"
+  },
+  "pointset": {
+    "points": {
+      "return_air_temperature_sensor": {
+        "units": "Degrees-Celsius",
```
pisuke, line 43: I would change this to `degrees_celsius` to show that this is not a BACnet unit

PR #1042, opened 2024-12-04
Add stricter config rate-limiting

grafnu:


validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
```diff
@@ -384,6 +386,7 @@ private static void setupSequencer() {
     String registrySuffix = exeConfig.registry_suffix;
     altRegistry = SiteModel.getRegistryActual(udmiNamespace, altRegistryId, registrySuffix);
     altClient = getAlternateClient();
+    ifNotNullThen(altClient, IotReflectorClient::activate);
```
KhyatiMahendru, line 389: I also want to understand why we are now calling for client activation and why it worked before without it

validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
```diff
@@ -1252,47 +1255,33 @@ protected void updateConfig(String reason) {
...
+      updateConfig(reason, waitForSync, UPDATE, deviceConfig);
     }
 
     assertConfigIsNotPending();
 
     captureConfigChange(reason);
   }
 
-  private boolean updateConfig(SubFolder subBlock, Object data) {
+  private boolean updateConfig(String reason, boolean waitForSync, SubFolder subBlock,
```
KhyatiMahendru, line 1283: Okay, I kind of get this piece but would still like to share my understanding - We are now waiting for at least CONFIG_BARRIER seconds before making another config update on the device. 

Sequencer needs to validate each config update and that is why we care about not losing any?

When applying config to any device generally, we would not care about losing any intermediate ones, right? Because only the latest one sticks?

grafnu: I made the config barrer logic dependent on the log-level setting (easiest way for now). So if it's TRACE then it will gate the config, but if it's not then it just treats it as per normal.  In the future we can make that more configurable somehow if needed.

PR #1045, opened 2024-12-06
Fix Proxy Device Subscription Issue and Improve Stability

MertCingoz:
The root cause was that proxy devices were unable to recreate configuration handlers after a connection loss.
This PR also includes various stability improvements for the library and the Pubber client.

Steps to reproduce:
- Run a gateway with 50 proxy devices
- Run registrar


pubber/src/main/java/daq/pubber/PubberDeviceManager.java
```diff
@@ -22,23 +25,25 @@ public class PubberDeviceManager extends PubberManager implements DeviceManager
...
     systemManager = new PubberSystemManager(host, configuration);
-    pointsetManager = new PubberPointsetManager(host, configuration);
     localnetManager = new PubberLocalnetManager(host, configuration);
-    gatewayManager = new PubberGatewayManager(host, configuration);
+    pointsetManager = new PubberPointsetManager(host, configuration);
     discoveryManager = new PubberDiscoveryManager(host, configuration, this);
+    gatewayManager = new PubberGatewayManager(host, configuration);
+    subManagers = Arrays.asList(
+            systemManager, localnetManager, pointsetManager, discoveryManager, gatewayManager);
+    Collections.reverse(subManagers);
```
grafnu, line 42: Why reverse (why not just create it reversed initially?) -- also add a brief comment as to why the ordering is important.

pubber/src/main/java/udmi/lib/client/DeviceManager.java
```diff
@@ -70,11 +70,11 @@ default String getTestingTag() {
...
    */
   default void updateConfig(Config config) {
-    getPointsetManager().updateConfig(config.pointset);
     getSystemManager().updateConfig(config.system, config.timestamp);
-    getGatewayManager().updateConfig(config.gateway);
-    getDiscoveryManager().updateConfig(config.discovery);
     getLocalnetManager().updateConfig(config.localnet);
+    getPointsetManager().updateConfig(config.pointset);
+    getDiscoveryManager().updateConfig(config.discovery);
+    getGatewayManager().updateConfig(config.gateway);
```
grafnu, line 77: comment as to why ordering is important

validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
```diff
@@ -246,7 +246,8 @@ public class SequenceBase {
   private static final String FAKE_DEVICE_ID = "TAP-1";
   private static final String NO_EXTRA_DETAIL = "no logs";
   private static final Duration DEFAULT_WAIT_TIME = Duration.ofSeconds(10);
-  private static final Duration CONFIG_WAIT_TIME = Duration.ofSeconds(30);
+  // TODO Config wait time increased for endpoint_failure_and_restart test
```
grafnu, line 249: TODO?  Comment is describing the change not what should be done... ?

pubber/src/main/java/daq/pubber/PubberSystemManager.java
```diff
@@ -116,7 +112,9 @@ public Date getDeviceStartTime() {
 
   @Override
   public void periodicUpdate() {
-    sendSystemEvent();
+    if (host instanceof Pubber) {
```
grafnu, line 115: why this check?

etc/validator.out
```diff
@@ -34,7 +33,6 @@ sites/udmi_site_model/out/devices/GAT-123/state_system.json
 sites/udmi_site_model/out/devices/GAT-123/validation_report.json
 sites/udmi_site_model/out/devices/SNS-4/config.json
 sites/udmi_site_model/out/devices/SNS-4/events_pointset.json
-sites/udmi_site_model/out/devices/SNS-4/events_system.json
```
grafnu, line 37: why are these going away?

MertCingoz: Temporarily disabling system events for virtual proxy devices until we identify useful metrics to send, as they are currently reporting JVM CPU and memory usage of the gateway, which is not beneficial and creates unnecessary network traffic.

pubber/src/main/java/daq/pubber/PubberSystemManager.java
MertCingoz: It's still sending system events for the gateway and direct devices.

validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
MertCingoz: Removed, I was looking to increase the timeout for a specific test and forgot to remove it. This will increase the overall test duration in case of a failure.

etc/validator.out
MertCingoz: Okay, I just re-enabled it and removed only the metrics from the proxy devices. It seems that the sequencer's 1-second delay between event processing is causing a timeout in the 50-proxy-device setup.

pubber/build.gradle
```diff
@@ -55,6 +55,7 @@ jacocoTestReport {
 checkstyle {
     ignoreFailures = false
     maxWarnings = 0
+    configFile = file('../etc/checkstyle.xml')
```
grafnu, line 58: There's already pubber/config/checkstyle/checkstyle.xml -- do you know how this configFile option plays with that?  Overall, I like the idea of a centralized checkstyle.xml, I just want to make sure that this isn't taking an inadvertent step backwards (i.e., ignoring the existing one).

validator/src/main/java/com/google/daq/mqtt/sequencer/SequenceBase.java
```diff
@@ -2203,7 +2203,6 @@ protected int countReceivedEvents(Class<?> clazz) {
 
   protected <T> List<T> popReceivedEvents(Class<T> clazz) {
     SubFolder subFolder = CLASS_SUBFOLDER_MAP.get(clazz);
-    safeSleep(EVENT_WAIT_DELAY_MS);
```
grafnu, line 2206: hmmm, not sure we can do this. The problem is that there's can be a long delay from when events are sent and ultimately received -- and there's no (easy) deterministic way to know it's been "long enough" ... Is there a particular problem this was causing?

MertCingoz: When 50 devices try to send pointset and system events, a 100-second random delay makes it impossible to run the sequencer for proxy devices.

grafnu: As discussed -- we'll leave this for now and then clean it up in a later pass (bug filed).

PR #1048, opened 2024-12-13
Generate config despite schema violation warning

KhyatiMahendru:


validator/src/main/java/com/google/daq/mqtt/util/ConfigManager.java
```diff
@@ -60,6 +65,7 @@ public ConfigManager(Metadata metadata, String deviceId, SiteModel siteModel) {
     this.metadata = metadata;
     this.deviceId = deviceId;
     this.siteModel = siteModel;
+    this.schemaViolations = new ArrayList<>();
```
grafnu, line 68: nit: move constructor in definition (since it's final)

validator/src/main/java/com/google/daq/mqtt/util/ConfigManager.java
```diff
@@ -311,4 +316,42 @@ private DiscoveryConfig getDiscoveryConfig() {
...
+
+  private void recordIfSchemaViolations(Runnable function) {
+    try {
+      function.run();
+    } catch (Exception e) {
+      schemaViolations.add(e);
+    }
+  }
+
+  private <T> void recordIfSchemaViolations(Consumer<T> consumer, T argument) {
```
grafnu, line 331: Are these three variants used? I don't see them anywhere... and, even so, could they be collapsed into two forms (supplier or runnable) and have the argument be encapsulated in the calling lambda?

tests/sites/missing/devices/AHU-22/expected/errors.map
```diff
@@ -1,4 +1,4 @@
 Exceptions for AHU-22
-  While converting device config
```
grafnu, line 2: ideally we'd find a way to preserve this context so that it's easier to diagnose. E.g., something like:

```
Exceptions for AHU-22
  While converting device config
    While converting point filter_alarm_pressure_status
      protocol ref FAXLE12.present_value does not match expression bacnet://[1-9][0-9]*/([A-Z]{2,4}):([1-9][0-9]*)(#[_a-z]+)?
      protocol ref MCN8.present_value does not match expression bacnet://[1-9][0-9]*/([A-Z]{2,4}):([1-9][0-9]*)(#[_a-z]+)?
      protocol ref AI2.differential does not match expression bacnet://[1-9][0-9]*/([A-Z]{2,4}):([1-9][0-9]*)(#[_a-z]+)?
 ```
E.g. consider using the (defined in the code somewhere) `ExceptionList` class, and have that returned by `getSchemaViolations()` (instead of returning the direct list of exceptions), with the appropriate context (converting config, point names, etc...).  I _think_ that should result in the right error output, or at least close enough.  There might need to be a tweak to the error reporting function on what happens when it sees an ExceptionList -- but that should be relatively straightforward.

grafnu (approved):

PR #1052, opened 2024-12-17
Fix mqtt connection issues

MertCingoz:


pubber/src/main/java/udmi/lib/base/ManagerBase.java
```diff
@@ -183,21 +183,24 @@ protected synchronized void startPeriodicSend() {
...
+      periodicSender = null;
     }
     if (initialUpdate != null) {
-      initialUpdate.cancel(false);
+      cancelFeature(initialUpdate);
       initialUpdate = null;
     }
   }
 
+  private void cancelFeature(ScheduledFuture<?> future) {
```
grafnu, line 196: cancelFuture? (not Feature)

pubber/src/main/java/udmi/lib/base/ManagerBase.java
```diff
@@ -183,21 +183,24 @@ protected synchronized void startPeriodicSend() {
...
+    try {
+      future.cancel(false);
+    } catch (Exception e) {
+      throw new RuntimeException("While cancelling executor", e);
```
grafnu, line 200: "cancelling future"

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/DiscoverySequences.java
```diff
@@ -331,12 +331,14 @@ private void scanAndVerify(Date scanStart, DiscoveryScanMode shouldEnumerate) {
 
     checkThat("discovery events were valid", reasons.isEmpty(), CSV_JOINER.join(reasons));
 
+    /* Allow duplicate scan addresses
```
grafnu, line 334: Erm? Why this?  Duplicate addresses should be filtered out client side...

grafnu (commented):
Just a few small comments, and there's some tests failing (should be fixable with a simple file copy as indicated in the logs)

validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/DiscoverySequences.java
MertCingoz: For my case multiple devices running on a single host, requiring them to share the same IPv4 and BACnet addresses. They indeed different virtual gateways sharing same network from underlying host.

MertCingoz:
> Ok -- we're gonna need to work this out. There's an implicit assumption that a scan_address is unique in some places, so two things that share (e.g.) a BACnet address will end up being the same thing. Can you maybe write up a short doc (so not a comment thread in this PR) that outlines what the underlying case for this is?

I'll summarize the underlying case for this and put it together in a short document for clarity.

There are some other issues I noticed; hopefully, you’re already aware of them
- Registrar failing to generate config file for some reason and CI is not stable at this moment. I had run tests multiple times in order to bypass.
- Devices were receiving 2 configuration messages due to the last_start logic. I’ve reverted the commit in this PR, and it’s working for me when I test connecting the devices.
- `RESULT fail enumeration.families enumerate_families PREVIEW 0/10 Failed check that family enumeration size matches: received ether, ipv4, ipv6, vendor, bacnet, expected ether, ipv4, vendor, bacnet` this probably needs to be fixed on site_model.

grafnu:
I know the tests are flaky, but not specifically about registrar failing to
generate a config file as the root cause -- we're tracking a number of such
things and we'll get to them eventually, but if there's something that's
specifically causing you extra headaches then let me know (but, the short
term answer will roughly be the same as "run it again"...)

Nothing wrong with receiving 2 config messages.  The first one is the
initial as-you-subscribe one, and the second is a response to the
last_start logic.  It's a very low-level ACK handshake with the backend
services.  Anyway, it shouldn't be a problem.

Some of the errors in there are intentional (to vet the error handling
stuff) -- so without more context for the "RESULT fail" line you showed I
can't say much about it.

etc/mosquitto_udmi.conf
```diff
@@ -11,3 +11,6 @@ cafile /etc/mosquitto/certs/ca.crt
 certfile /etc/mosquitto/certs/rsa_private.crt
 keyfile /etc/mosquitto/certs/rsa_private.pem
 require_certificate true
+
+listener 1883
```
grafnu, line 15: what's this for?  Were you actively using it somewhere or is it a debugging hook?

grafnu (commented):
Sill pending the question about unique `scan_addr` field.

etc/mosquitto_udmi.conf
MertCingoz: Yes this is for local development with containers until we implement support for CA certificates.

MertCingoz: Port seems to be occupied by CI, reverted this change.

MertCingoz:
Ok, `scan_addr` must be unique for a discovery scan, so I reverted the changes. It should indicate a poorly configured network, such as BACnet addresses being incorrectly assigned to multiple devices, which causes the tests to fail.

What happened was that I have a site model with a mix of Pubber and Mango devices. The Mango devices share the same IP address and BACnet local network addresses. Pubber's BACnet discovery relies only on local network family addresses from metadata files, which caused the tests to fail for Pubber.

grafnu (approved):

PR #1056, opened 2024-12-20
Better logging for sequencer family_addr tests

khyatimahendru:


validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/LocalnetSequences.java
```diff
@@ -23,9 +23,10 @@ private void familyAddr(String family) {
         catchToNull(() -> deviceMetadata.localnet.families.get(family).addr),
         format("No %s address defined in metadata", family));
     HashMap<String, FamilyLocalnetState> families = deviceState.localnet.families;
-    waitUntil(format("localnet family state %s available", family),
+    waitUntil(format("device state localnet family %s is available", family),
         () -> families.containsKey(family) ? null
-            : format("family %s not in %s", family, CSV_JOINER.join(families.keySet())));
+            : format("Because family %s not found in device state localnet family set: %s", family,
```
grafnu, line 28: nit: I think this should start lower-case (`because` not `Because`) to maintain consistency when things are automatically combined.

grafnu (approved):

PR #1057, opened 2024-12-24
Fix flaky tests

khyatimahendru:


grafnu (approved):
Nice catch -- hopefully that wasn't too hard to figure out.  Not what I was expecting!

PR #1058, opened 2024-12-27
Ignore schema for all reply_ subtypes

khyatimahendru:


validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
```diff
@@ -728,7 +728,8 @@ private ReportingDevice validateMessageCore(Object messageObj, Map<String, Strin
       writeDeviceOutCapture(messageObj, attributes, deviceId, schemaName);
 
       String subFolder = attributes.get(SUBFOLDER_PROPERTY_KEY);
-      boolean processSchema = !IGNORE_FOLDERS.contains(subFolder);
+      boolean processSchema = !SubType.REPLY.value().equalsIgnoreCase(schemaName)
```
grafnu, line 731: Why the ignore case?  Generally we should be pedantic about case matching (just avoids confusion elsewhere).

Also, I'm just generally confused how this works... because SubType would be `reply` but the schemaName would be something with a `reply_` prefix? -- so they never would be equal irrespective of case, so this change seems like a no-op?

khyatimahendru: 1. I removed the ignore_case.

1. Particularly for reply_update, there's this condition in function messageSchema() which would return 'reply' as the schemaName - so I had made the comparison with schemaName. **I have fixed it and changed it to compare with subType instead.** 

```
  private String messageSchema(Map<String, String> attributes) {
    String subFolder = attributes.get(SUBFOLDER_PROPERTY_KEY);
    String subType = attributes.get(SUBTYPE_PROPERTY_KEY);

    if (SubFolder.UPDATE.value().equals(subFolder)) { <-------
      return subType;
    }
```

I need some suggestions on how I can test this change. So far, I have created a test schema under tests/traces/simple under an existing device GAT-123 with the name 005_reply_update.json.

tests/traces/simple/expected/devices/GAT-123/reply.out
```diff
@@ -0,0 +1,13 @@
+{
+  "timestamp" : "1999-10-20T01:02:03Z",
```
grafnu, line 2: hmmm -- maybe I'm missing something here but I thought there would not be a reploy.out since it's filtered out?

khyatimahendru: I fixed this and removed these reply files

grafnu (approved):

PR #1059, opened 2025-01-06
Update registrar usage forms in its help function & update docs

khyatimahendru:


common/src/main/java/com/google/udmi/util/CommandLineProcessor.java
```diff
@@ -91,6 +99,10 @@ private void showUsage() {
    */
   public void showUsage(String message) {
     ifNotNullThen(message, m -> System.err.println(m));
+    ifNotNullThen(this.usageForms, () -> {
+      System.err.println("Usage forms:");
+      this.usageForms.forEach(form -> System.err.printf("\t%s%n", form));
```
grafnu, line 104: nit: I generally avoid tabs since they just end up being more confusing (sometimes preserved by cut-and-past, sometimes not, can confuse diff output, etc... ) -- so I'd just expand with spaces...

common/src/main/java/com/google/udmi/util/CommandLineProcessor.java
```diff
@@ -28,6 +28,7 @@ public class CommandLineProcessor {
   private static final int LINUX_ERROR_CODE = -1;
   private final Object target;
   private final Method showHelpMethod;
+  private List<String> usageForms = null;
```
grafnu, line 31: null initializer is redundant

common/src/main/java/com/google/udmi/util/CommandLineProcessor.java
```diff
@@ -91,6 +99,10 @@ private void showUsage() {
    */
   public void showUsage(String message) {
     ifNotNullThen(message, m -> System.err.println(m));
+    ifNotNullThen(this.usageForms, () -> {
```
grafnu, line 102: in `this.usageForms` the `this` isn't necessary (not like python!)

grafnu (approved):

PR #1064, opened 2025-01-07
Spew out model update for BIND and DELETE operations

grafnu:


bin/pull_mqtt
```diff
@@ -31,15 +31,17 @@ sudo mosquitto_sub $SERVER_OPTS -R -F "%j" -t $topic_filter |
...
 
         readarray -d '/' -t array <<< "${topic}/"
-        registryId=${array[2]}
-        deviceId=${array[4]}
-        subType=${array[5]}
-        subFolder=${array[6]}
+        registryId=${array[2]:-}
+        deviceId=${array[4]:-}
+        subType=${array[5]:-}
+        subFolder=${array[6]:-}
```
khyatimahendru, line 37: I am looking to understand how this updated code may be better than the original one. As far as I understood, they would both behave similarly so want to understand why this change may be required. 
```
# original 
subFolder=${array[6]}
subFolder=$(echo $subFolder | xargs)
[[ -n ${subFolder% } ]] || subFolder=update
```
versus
```
# updated
subFolder=${array[6]:-}
subFolder=$(echo $subFolder | xargs)
[[ -n $subFolder ]] || subFolder=update
```

udmis/src/main/java/com/google/bos/udmi/service/core/ReflectProcessor.java
```diff
@@ -152,7 +152,7 @@ private void reflectUdmiLog(Envelope attributes, String message) {
   private Object extractModel(CloudModel request) {
     String metadata = catchToNull(() -> request.metadata.get(MetadataMapKeys.UDMI_METADATA));
     if (metadata == null) {
-      return null;
+      return asModelUpdate(request);
```
khyatimahendru, line 155: I would also like to understand what this change means in terms of behavior. Why does metadata being null translate the request to being a model update?

grafnu: Ah -- so there's a model-update w/ metadata (used for CREATE and UPDATE), since there's a model there (defined my metadata)... and this is handling the case of a model-update for DELETE (or BIND) since there's no actual model involved.  They're both valid model updates, but have a slightly different structure to the messages (which can be decoded by the sender).  I added a brief comment for this!

khyatimahendru (approved):

PR #1072, opened 2025-01-21
Add utility for GSheets logger to stream logs to Sheets

khyatimahendru:


bin/stream_logs_to_sheets.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
grafnu, line 1: no `.sh` at the end of the name since given the #!/ prefix it's a generic executable.

grafnu: This looks like a utility (sourced elsewhere) not an executable, so should go in the etc/ directory.  (Also, then likely wouldn't have the #!/ header, and WOULD have the .sh extension

khyatimahendru: Moved it to etc/

bin/toolrun
```diff
@@ -27,5 +28,6 @@ else
 fi
 
 cmd="java -cp $UDMI_JAR $JDWP $JAVA_CLASS $util_name $@"
+
 echo $cmd
-$cmd 2>&1 | tee $OUT_DIR/$util_name.log
+$cmd 2>&1 | tee $OUT_DIR/$util_name.log | stream_to_gsheets "$util_name" "12leoCVQpdXx6MdKlVCCI4O1mi35dOTw_FxDMBetuwXQ"
```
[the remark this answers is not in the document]
khyatimahendru: Made the spreadsheet id configurable in site_metadata.json

schema/site_metadata.json
```diff
@@ -92,6 +92,11 @@
           "description": "Source repository where the UDMI site model is stored",
           "type": "string",
           "examples": ["https://github.com/faucetsdn/udmi_site_model", "git@github.com:faucetsdn/udmi_site_model.git"]
+        },
+        "logs": {
```
grafnu, line 96: let's think of something more generic than "logs" because I think it will ultimately be used for other things. E.g. error output, validation status dashboard, sequencer. If there value here really is just a link to the spreadsheet, but doesn't (e.g.) specify the exact tab to use, then we could just name it something like "sheet"

etc/stream_logs_to_sheets.sh
```diff
@@ -0,0 +1,21 @@
+# Utility to stream logs from any process to gSheets
+# Will be sourced by other executing scripts
+#
+
+UDMI_ROOT=$(realpath $(dirname $0)/..)
+
+source $UDMI_ROOT/etc/shell_common.sh
```
grafnu, line 7: Should all of this file just go into shell_common.sh itself?  There' no direct side-effects...

khyatimahendru: Moved the function to shell_common.sh

schema/site_metadata.json
khyatimahendru: Renamed logs to sheet

etc/shell_common.sh
```diff
@@ -104,3 +104,11 @@ export UDMI_VERSION=$udmi_version
 export UDMI_COMMIT=$udmi_commit
 export UDMI_TIMEVER=$udmi_timever
 export UDMI_REV=$udmi_rev
+
+stream_to_gsheets() {
+  local tool_name=$1
+  local sheet_id=$2
+  timestamp=$(date +%Y%m%d_%H%M%S)
+  java -cp "$UDMI_JAR" "com.google.udmi.util.SheetsOutputStream" "$tool_name" \
+  "$sheet_id" "$tool_name.$timestamp.log"
```
grafnu, line 113: should be some hanging indent here

schema/site_metadata.json
```diff
@@ -92,6 +92,11 @@
           "description": "Source repository where the UDMI site model is stored",
           "type": "string",
           "examples": ["https://github.com/faucetsdn/udmi_site_model", "git@github.com:faucetsdn/udmi_site_model.git"]
+        },
+        "sheet": {
+          "description": "Link to a spreadsheet to observe real-time output from any server-side tool",
```
grafnu, line 97: it's really "any tool run" -- not necessarily just server side... e.g. if I run it at the CLI I would expect the same thing to happen.

grafnu (approved):

PR #1077, opened 2025-01-28
Skip metadata feature comparison

MertCingoz:


validator/src/main/java/com/google/daq/mqtt/sequencer/sequences/DiscoverySequences.java
```diff
@@ -182,23 +182,24 @@ private boolean shouldEnumerate(Depth depth) {
...
     Set<String> unofficial = enumeratedFeatures.stream()
         .filter(feature -> !Bucket.contains(feature)).collect(Collectors.toSet());
     String format = format("unrecognized { %s }", CSV_JOINER.join(unofficial));
     checkThat("all enumerated features are official buckets", unofficial::isEmpty, format);
   }
 
-  private boolean isTrue(Boolean condition) {
-    return ofNullable(condition).orElse(false);
+  private void checkFeatureMetadata(Set<String> enumeratedFeatures) {
+    ifNotNullThen(deviceMetadata.features, metadataFeatures -> {
```
grafnu, line 193: I think this should be `ifNullSkipTest` -- if metadata is not defined, then it's not that the test passes, it's that it shouldn't be tested at all...

grafnu (approved):

PR #1080, opened 2025-01-31
Draft: Python gencode with datamodel-codegen

khyatimahendru:


grafnu (commented):
Can you add a simple example in here somewhere that just shows how it works?  So, something that does something like 1) parses some JSON, 2) modifies some value (does a read-modify-write), 3) creates a new object and adds it in somewhere, 4) writes it out as JSON?   The "does this work" is less about the thing itself and more about how it's used!

khyatimahendru:
Here's how `bin/gencode_py` works:

1. The script begins by creating a Python virtual environment. It installs two Python packages:
    * `datamodel-code-generator`: This generates Python code from JSON schemas.
    * `poetry`: This tool manages dependencies and packages Python projects.

2. The JSON schemas use `$ref` with the `file:` prefix for referencing other schemas (e.g., `"file:common.json#/definitions/depth"`). **`datamodel-codegen` doesn't support this prefix.** Therefore, we need to preprocess the schemas:
    * It creates a directory `codegen/pyschema`.
    * It uses the `sed` command to remove the `file:` prefix from all `$ref` entries in the schema files and copies the modified files into the `codegen/pyschema` directory.  For example:
  
    ```json
    // Original schema/config_discovery_family.json
    "depth": {
      "description": "Indicates implicit enumeration of discovered devices",
      "$ref": "file:common.json#/definitions/depth"
    }

    // Processed codegen/pyschema/config_discovery_family.json
    "depth": {
      "description": "Indicates implicit enumeration of discovered devices",
      "$ref": "common.json#/definitions/depth"
    }
    ```

3. The script uses the following command to generate Python code from the preprocessed schemas in `codegen/pyschema`. The flags used are explained below:

    ```bash
    datamodel-codegen --input "$SCHEMA_OUTPUT_DIR" --input-file-type jsonschema \
    --output "$PYTHON_OUTPUT_DIR" --use-default --force-optional \
    --use-schema-description --reuse-model --use-exact-imports \
    --output-model-type pydantic_v2.BaseModel --disable-timestamp
    ```

    * `--input "$SCHEMA_OUTPUT_DIR"`:  Specifies the input directory containing the JSON schemas (the `codegen/pyschema` directory).
    * `--input-file-type jsonschema`: Indicates the input files are JSON schemas.
    * `--output "$PYTHON_OUTPUT_DIR"`: Specifies the output directory for the generated Python code.
    * `--use-default`:  Uses default value wherever defined, even if a field is required.
    * `--force-optional`: Makes all fields optional.
    * `--use-schema-description`:  Uses the schema descriptions for docstrings in the generated code.
    * `--reuse-model`: Reuses existing model definitions. (Default behavior is to redefine the model definition in the current model instead of importing it from its existing definition elsewhere in the module)
    * `--use-exact-imports`: Uses exact import paths,, for example: "from .foo import Bar" instead of "from . import foo" with "foo.Bar"
    * `--output-model-type pydantic_v2.BaseModel`: Generates Pydantic v2 models.
    * `--disable-timestamp`: Disables timestamp generation in the docstrings in the generated code. 
    
5. The script adds necessary `__init__.py` files (to make the generated code a Python package) and a `pyproject.toml` file (Poetry's project configuration file) to the generated code directory.

6. Finally, the script uses `poetry build` to create a python wheel file (.whl) from the generated code. This wheel can then be used in the `pyudmi` library and also be published to PyPI, if required.


---
Some interesting examples from the generated code:
1. Handling of `oneOf` with regex patterns - https://github.com/faucetsdn/udmi/blob/ec68320fc93c055de7614e2515593d7a5e03a126/gencode/python/udmi/schema/category.py
2. Multiple enums in a file - https://github.com/faucetsdn/udmi/blob/ec68320fc93c055de7614e2515593d7a5e03a126/gencode/python/udmi/schema/common.py
3. Multiple classes in a file -
https://github.com/faucetsdn/udmi/blob/ec68320fc93c055de7614e2515593d7a5e03a126/gencode/python/udmi/schema/site_metadata.py

grafnu:
> Here's how `bin/gencode_py` works:

Sorry, I could have been more clear... I was looking for an example of how the generated code is used by a client program, not how the generated code is generated. E.g. something with an `import udmi` equivalent in it.

PR #1083, opened 2025-02-05
Downgrade config messages with localnet addrs

grafnu:


khyatimahendru (approved):

grafnu:
It's also required in some other situations -- not just multi-threading!
In this particular case, you can't have a variable reference inside of a
lambda expression unless it's "final or effectively final" (which means it
can't be modified)... the compiler will complain bitterly if you try. So --
in this case, I explicitly want to modify it, and I trick things kinda by
wrapping it in the AtomicReference.  The reference itself (argRef) is final
(not modified), but that doesn't mean I can't modify the contents of what
it points to! There's some underlying reason for this (look up "java memory
fences"), but you don't really need to understand that (but it's good to)
to just do the right thing.  In addition to multi-threading, the Atomic
classes also provide guarantees around memory ordering that can trip you up
even with a single-threaded setup. I think (I'm not 100% sure myself), that
the compiler basically deals with this in the obvious case of execution
through a function, and through function calls -- but that's not possible
when doing a lambda in quite the same way, thus the requirement.

PR #1084, opened 2025-02-06
Draft: UDMI Python Library

khyatimahendru:


misc/pyudmi/src/udmi/client/manager/discovery/discovery_manager.py
```diff
@@ -0,0 +1,382 @@
...
+                            self._set_internal_status(Status.SCHEDULED)
+                            self.state.phase = Phase.pending
+                        else:
+                            # The scan is not repetitive, exit the scheduler
+                            return
+
+            time.sleep(RUNNER_LOOP_INTERVAL)
+
+    @catch_exceptions_to_status
+    def apply_discovery_config(self, config_dict):
```
grafnu, line 317: is there a place in here where there's an example of how this dict is created?  I can guess (likely just a standard json-to-dict parsing), but I wanted to get it codified somewhere to be 100% sure. I saw the `self.publish(event.to_json())` code elsewhere, so I'm looking for the inverse...

(also, as an abstraction thing I'd expect the input/output paths to be roughly similar, so that the concept of "json" is either entirely in this layer, or not... so if JSON is output here then it would also be parsed here. I.e. it produces the same kind of thing that it consumes!)

PR #1089, opened 2025-02-13
Handle device metadata updates & deletes in validator

khyatimahendru:


validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
```diff
@@ -768,6 +794,7 @@ private ReportingDevice validateMessageCore(Object messageObj, Map<String, Strin
 
       if (!device.hasErrors()) {
         outputLogger.info("Validation clean %s", messageTag);
+        handleIfMetadataUpdate(device, attributes, messageObj);
```
noursaidi, line 797: I don't think validator should validate metadata updates (because that should be the role of registrar and is not a message from a device), and there is also no guarantee that metadata messages will pass validation. 

Can this `handleIfMetadataUpdate` move so it's earlier?

khyatimahendru: Moved it earlier

validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
```diff
@@ -646,10 +653,40 @@ private void handleUdmiConfig(UdmiConfig udmiConfig) {
...
+    }
+
+    ReportingDevice device = reportingDevices.get(deviceId);
+    String subFolderRaw = attributes.get(SUBFOLDER_PROPERTY_KEY);
+    String subTypeRaw = attributes.get(SUBTYPE_PROPERTY_KEY);
+
+    if (SubType.MODEL.value().equals(subTypeRaw) && SubFolder.UPDATE.value().equals(subFolderRaw)) {
+      Metadata metadata = convertTo(Metadata.class, messageObject);
+
+      if (metadata == null) {
```
grafnu, line 670: This can only happen if `messageObject` is `null` to begin with, so I think it would be better to have a `requireNonNull()` on `messageObject`, since it should never even get here in that case!  (But if it does somehow when it would be very obvious!)

validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
```diff
@@ -646,10 +654,33 @@ private void handleUdmiConfig(UdmiConfig udmiConfig) {
...
+    String subFolderRaw = attributes.get(SUBFOLDER_PROPERTY_KEY);
+    String subTypeRaw = attributes.get(SUBTYPE_PROPERTY_KEY);
+
+    if (SubType.MODEL.value().equals(subTypeRaw) && SubFolder.UPDATE.value().equals(subFolderRaw)) {
+      if (reportingDevices.containsKey(deviceId)) {
+        ReportingDevice device = reportingDevices.get(deviceId);
+        Metadata metadata = convertTo(Metadata.class,
+            requireNonNull(messageObject, "messageObject is null"));
+
+        if (metadata.cloud != null && metadata.cloud.operation == Operation.DELETE) {
```
grafnu, line 668: observation: you can also do (I think):
```
if (catchToNull(() -> metadata.cloud.operation) == Operation.DELETE) {
  ...
}
```

Functionally they are the same, but just written different -- purely a matter of style.  Totally up to you which you like better (I tend to over abstract things sometimes).

grafnu (approved):

PR #1096, opened 2025-02-20
Add targeted discovery capability

grafnu:


noursaidi (approved):

tests/schemas/config/discovery_targeted.out
jainrocks, line 1: This an empty file which is being added, we can avoid adding this.

tests/schemas/config/discovery_depth.out
jainrocks, line 1: This an empty file which is being added, we can avoid adding this if not needed.

tests/schemas/config/discovery_targeted.json
```diff
@@ -0,0 +1,16 @@
...
+// where {device_id} is that of the discovery node.
+//
+{
+  "version": "1.5.2",
+  "timestamp": "2018-08-26T21:39:29.364Z",
+  "discovery": {
+    "families": {
+      "bacnet": {
+        "generation": "2018-08-26T21:00:13Z",
+        "addrs": [ "9218", "294124", "12933" ]
```
jainrocks, line 12: Shall we add more details around these addresses? 
Or is it specific to tests only?

grafnu:
These empty files are "golden files" in that the testing subsystem uses
them to mark where there is an expected empty file in the output. Empty is
not always the same as missing!

grafnu:
Good idea -- I added some comments.

These are also used as examples some places, so it can be good to annotate
them with things as necessary.

PR #1098, opened 2025-03-03
fix distroless docker container image

noursaidi:


noursaidi:
Merging without review because misc/

PR #1100, opened 2025-03-06
Add depths to bacnet discovery

noursaidi:


noursaidi:
Merging because misc/ directory

PR #1105, opened 2025-03-19
Upgrade pyOpenSSL for discoveryNode

khyatimahendru:
dependabot has upgraded cryptography to version 44.0.1
This is now causing conflicts and leading to failure of the discovery tests job:

```
1.649 The conflict is caused by:
1.649     The user requested cryptography==44.0.1
1.649     pyopenssl 24.2.1 depends on cryptography<44 and >=41.0.5
```


grafnu (approved):
NB: For something small like this feel free to submit without review.

PR #1109, opened 2025-03-24
Enable discovery through non-gateway devices

khyatimahendru:


udmis/src/main/java/com/google/bos/udmi/service/core/ProvisioningEngine.java
```diff
@@ -112,6 +113,42 @@ private boolean shouldProvision(Date generation, CloudModel cloudModel) {
     return TRUE_OPTION.equals(ifNotNullGet(cloudModel.metadata, m -> m.get(UDMI_PROVISION_ENABLE)));
   }
 
+  private void gatewayDiscovery(DiscoveryEvents discoveryEvent, CloudModel gatewayModel,
```
grafnu, line 116: maybe `provisionProxiedDevice`?  I had to read this twice to understand what it was actually _doing_...  Or `provisionWithGateway` or something.

udmis/src/main/java/com/google/bos/udmi/service/core/ProvisioningEngine.java
```diff
@@ -120,26 +157,30 @@ public void discoveryEvent(DiscoveryEvents discoveryEvent) {
...
       }
+
+      Date generation = requireNonNull(discoveryEvent.generation, "missing scan generation");
       String family = requireNonNull(discoveryEvent.family, "missing family");
       String addr = requireNonNull(discoveryEvent.addr, "missing addr");
       String expectedId = format(DISCOVERED_DEVICE_FORMAT, family, addr);
-      if (deviceIds.contains(expectedId)) {
-        debug("Scan device %s/%s target %s already registered", registryId, gatewayId, expectedId);
+
+      if (isGateway(deviceModel)) {
```
grafnu, line 178: Something doesn't smell right here since the only difference between the two cases (I think, and this is the essential question) is that if it is a gateway then the device should be bound, but if it's not a gateway then it should not be bound.  That difference is manifest (in the `createDeviceEntry` call), but then there are _also_ other differences I don't understand.

udmis/src/main/java/com/google/bos/udmi/service/core/ProvisioningEngine.java
```diff
@@ -112,6 +113,42 @@ private boolean shouldProvision(Date generation, CloudModel cloudModel) {
     return TRUE_OPTION.equals(ifNotNullGet(cloudModel.metadata, m -> m.get(UDMI_PROVISION_ENABLE)));
   }
 
+  private void gatewayDiscovery(DiscoveryEvents discoveryEvent, CloudModel gatewayModel,
+      Envelope envelope, String registryId, String deviceId, Date generation, String expectedId) {
+    Set<String> deviceIds = refreshModelDevices(registryId, deviceId, generation, gatewayModel);
+    if (deviceIds == null) {
```
grafnu, line 119: doesn't `shouldProvision` work here too?  Seems like that should be constant in both cases.

udmis/src/main/java/com/google/bos/udmi/service/core/ProvisioningEngine.java
khyatimahendru: This is something I wanted to check too. 

From what I understood, if discovery node is a gateway, we would check the expected deviceId against the gateway's proxied devices. And then if it does not exist, we create a device entry and bind it to the the gateway. 

If the discovery node is not a gateway, there's no proxied devices. So we should check if the device exists in the registry. If not, create the device entry and do not bind it to any gateway.

In the gateway case, the difference is that we do not explicitly check if the device entry already exists in the registry. It would fail if it does and then be handled in line 68 and then be bound:

```
catchToElse(
        (Supplier<CloudModel>) () -> iotAccess.modelDevice(registryId, expectedId, cloudModel,
            null),
        (Consumer<Exception>) e -> error(
            "Error creating device (exists but not bound?): " + friendlyStackTrace(e)));
if (shouldBindToGateway) {
      bindDeviceToGateway(registryId, expectedId, deviceId);
    }
```

I guess if I remove the explicit check from the non gateway case too, the only difference would be the `bindToGateway` call.

udmis/src/main/java/com/google/bos/udmi/service/core/ProvisioningEngine.java
khyatimahendru: Sure, I will rename these functions

udmis/src/main/java/com/google/bos/udmi/service/core/ProvisioningEngine.java
grafnu: I think it can be generalized to "is the new device in the registry" since all devices, either bound to a gateway or not, are in the registry either way.  The only optimization might be that for the gateway case you don't need to list ALL the devices so might be faster, but functionally it should be the same. You can see this in the current code with the function `private GatewayListOptions getBoundDevicesOptions(String gatewayId)` -- it's basically an option passed to the list function. 

You're not going to call that function directly -- but the point is to realize that the two cases are really the same thing. The only required difference is that in the gateway case it binds the discovered device to the gateway, but in the non-gateway case it doesn't.  Everything else (checking the should-provision setting, creating the device entry, checking if the device already exists) is essentially the same (with maybe some performance optimizations)!

This isn't really a question of how the UDMIS code works, but rather how ClearBlade works.

grafnu (approved):
Sorry, I didn't notice this was sitting here waiting for my review until just now!  Or something. Anyway, looks good!

PR #1112, opened 2025-04-28
Backend workers for sync to/from BAMBI

khyatimahendru:


common/src/main/java/com/google/udmi/util/JsonUtil.java
```diff
@@ -486,4 +489,121 @@ public static void writeFile(Object theThing, File file) {
...
+      if (value instanceof Map) {
+        flatten((Map<String, Object>) value, newKey, flattenedMap, separator);
+      } else {
+        flattenedMap.put(newKey, value);
+      }
+    }
+  }
+
+  public static JsonNode nestFlattenedJson(Map<String, String> flattenedJsonMap) {
+    ObjectMapper mapper = new ObjectMapper();
```
grafnu, line 516: There are pre-existing constants for various ObjectMappers at the top of this file... so likely best to use one of those.

common/src/main/java/com/google/udmi/util/JsonUtil.java
```diff
@@ -486,4 +489,121 @@ public static void writeFile(Object theThing, File file) {
...
+      int index, ObjectMapper mapper) {
+    if (index == parts.length - 1) {
+      if (currentNode instanceof ObjectNode) {
+        ((ObjectNode) currentNode).put(parts[index], value);
+      } else if (currentNode instanceof ArrayNode) {
+        try {
+          int arrayIndex = Integer.parseInt(parts[index]);
+          ((ArrayNode) currentNode).set(arrayIndex, mapper.getNodeFactory().textNode(value));
+        } catch (NumberFormatException e) {
+          System.err.println("Warning: Cannot set array value at non-numeric index: " + parts[index]);
```
grafnu, line 539: shouldn't silently squelch this error here -- likely should catch and then rethrow as a RuntimeException

common/src/main/java/com/google/udmi/util/JsonUtil.java
```diff
@@ -486,4 +489,121 @@ public static void writeFile(Object theThing, File file) {
...
+        }
+      }
+      return;
+    }
+
+    String currentPart = parts[index];
+    JsonNode nextNode = currentNode.get(currentPart);
+
+    boolean isArrayIndex = false;
```
grafnu, line 548: make this final with no assignment, and then set =false in the catch clause

common/src/main/java/com/google/udmi/util/JsonUtil.java
```diff
@@ -486,4 +489,121 @@ public static void writeFile(Object theThing, File file) {
...
+    int arrayIndex = -1;
+    if (index < parts.length - 1) {
+      try {
+        arrayIndex = Integer.parseInt(parts[index + 1]);
+        isArrayIndex = true;
+      } catch (NumberFormatException ignored) {
+      }
+    }
+
```
grafnu, line 557: function getting along to be too long, so consider ways to break it up...

etc/shell_common.sh
```diff
@@ -112,4 +112,18 @@ function stream_to_gsheets {
...
         "$sheet_id" "$tool_name.$timestamp.log"
-}
\ No newline at end of file
+}
+
+function sync_site_model_to_bambi {
+    local spreadsheet_id=$1
+    local local_path_to_site_model=$2
+    java -cp "$UDMI_JAR" "com.google.bos.iot.core.bambi.BambiSync" \
+    "$spreadsheet_id" "$local_path_to_site_model"
```
grafnu, line 121: indent

validator/src/main/java/com/google/bos/iot/core/bambi/BambiSiteModelManager.java
```diff
@@ -0,0 +1,295 @@
...
+   * Write relevant data to system, cloud, gateway, localnet, pointset, and points tabs in the
+   * spreadsheet.
+   *
+   * @param deviceToMetadataMap Mapping of device id to the device metadata
+   */
+  public void writeDevicesMetadata(Map<String, Map<String, String>> deviceToMetadataMap) {
+    Map<BambiSheet, List<List<Object>>> dataToWrite = new HashMap<>();
+
+    List<SheetConfig> simpleTableConfigs = List.of(
+        new SheetConfig(BambiSheet.SYSTEM, bambiSiteModel.getSystemDataHeaders(), "system."),
```
grafnu, line 109: seems like the last argument ("system.") can/should be implicitly calculated from the first (BambiSheet.SYSTEM)?

grafnu (commented):
Not sure if this was ready for review, but I saw it so I reviewed it!  Looks like a good chunk of work... the comments are mostly superficial.

etc/shell_common.sh
```diff
@@ -112,4 +112,18 @@ function stream_to_gsheets {
...
+}
+
+function sync_site_model_to_bambi {
+    local spreadsheet_id=$1
+    local local_path_to_site_model=$2
+    java -cp "$UDMI_JAR" "com.google.bos.iot.core.bambi.BambiSync" \
+        "$spreadsheet_id" "$local_path_to_site_model"
+}
+
+function sync_bambi_site_model_to_disk {
```
grafnu, line 124: is this the inverse of the one above?  If so, then I'd expect it to be `sync_bambi_to_site_model` or the other one would be `sync_disk_to_bambi_site_model` or something -- just parity in the words used (just in a different order)

grafnu (approved):
Looks good!  The thing I'm thinking about now is how to instantiate some kind of integration testing around this -- e.g. setting up a test-sheet for a repo (just like we have a test GCP instance) and making sure it all works as intended... not critical for the initial MVP, but at some point we'll likely need/want something like that!

etc/shell_common.sh
khyatimahendru: Renamed the function

PR #1115, opened 2025-05-04
Adding test for ms in timestamps

grafnu:


pubber/src/test/java/udmi/lib/MqttPublisherTest.java
```diff
@@ -52,7 +52,7 @@ class MockPublisher extends MqttPublisher {
 
     public MockPublisher(EndpointConfiguration configuration,
         Consumer<Exception> onError) {
-      super(configuration, onError, null);
+      super(configuration, onError, null, true);
```
jainrocks, line 55: Do we need to add tests for the case when msTimestamp is false?

jainrocks (approved):

pubber/src/main/java/udmi/lib/base/MqttDevice.java
```diff
@@ -48,12 +50,13 @@ public MqttDevice(String deviceId, MqttDevice target) {
     this.deviceId = deviceId;
     publisher = target.publisher;
     certManager = null;
+    msTimestamp = false;
```
jainrocks, line 53: We are hardcoding to false, if not present, which means by default the timestamp will not be in ms?

grafnu:
This is a very niche case and already covered by the integration tests, so
likely doesn't warrant more testing work at this point.

PR #1117, opened 2025-05-10
Handling spurious log for last start updates when the last start is not changed

jainrocks:
Handling spurious log for last start updates when the last start is not changed

### DIFF
--- a/udmis/src/main/java/com/google/bos/udmi/service/access/IotAccessBase.java
+++ b/udmis/src/main/java/com/google/bos/udmi/service/access/IotAccessBase.java
@@ -233,7 +233,13 @@ public abstract class IotAccessBase extends ContainerBase implements IotAccessPr
           debug("Retrieved config %s/%s #%d", registryId, deviceId, version);
           String updatedConfig = ifNotNullGet(safeMunge(munger, currentConfig),
               updated -> checkedUpdate(envelope, version, updated));
-          debug("Applied config %s/%s #%d", registryId, deviceId, version);
+          /**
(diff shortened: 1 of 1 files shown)


udmis/src/main/java/com/google/bos/udmi/service/access/IotAccessBase.java
```diff
@@ -233,7 +233,13 @@ public String modifyConfig(Envelope envelope, Function<Entry<Long, String>, Stri
           debug("Retrieved config %s/%s #%d", registryId, deviceId, version);
           String updatedConfig = ifNotNullGet(safeMunge(munger, currentConfig),
               updated -> checkedUpdate(envelope, version, updated));
-          debug("Applied config %s/%s #%d", registryId, deviceId, version);
+          /**
+           * Before adding below debug log, check if any configuration was actually updated.
+           * This prevents a spurious "applied config" message when no changes occurred.
```
grafnu, line 238: Nit: I usually try to limit comments to the _why_ of something when the what is indicated by the code. Also prefer // over /* for in-function comments. e.g., _consider_ just having:

`// Prevent a spurious "applied config" message when no changes occurred.`

(But, this is just a _nit_ of mine, so feel free to keep or change.)

grafnu (approved):

udmis/src/main/java/com/google/bos/udmi/service/access/IotAccessBase.java
jainrocks: sure, SGTM

PR #1120, opened 2025-05-20
Move BAMBI to a separate module for independent build and deployment

khyatimahendru:


services/bin/bambi_service
```diff
@@ -0,0 +1,140 @@
+#!/bin/bash -e
```
noursaidi, line 1: General comment - whilst the bash script is fine for initial testing, we should port it into a proper Java consumer

services/bin/bambi_service
```diff
@@ -0,0 +1,140 @@
...
+        gcloud pubsub subscriptions create "${BAMBI_REQUESTS_SUBSCRIPTION}" --topic="${BAMBI_REQUESTS_TOPIC}" --expiration-period=24h
+    fi
+fi
+
+echo Polling for messages...
+
+while true; do
+    PULL_OUTPUT=$(gcloud pubsub subscriptions pull \
+        "${BAMBI_REQUESTS_SUBSCRIPTION}" \
+        --auto-ack \
```
noursaidi, line 73: to consider for the future, what is the failure mechanism if there is an error? auto-ack ackknowledges the message, so if there is an error the request is lost.

grafnu (approved):
One commend about the location of the documentation -- not a super big deal, but just something to consider while it's still all new/current!

services/bin/bambi_service
khyatimahendru: Yes, that's the plan! I am already working on the java application - waiting on the git repository decision.

services/README.md
```diff
@@ -0,0 +1,43 @@
+## BAMBI: BOS Automated Management Building Interface - Backend Service
```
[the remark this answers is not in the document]
khyatimahendru: Moved it to docs/tools

PR #1123, opened 2025-06-04
Create release pipeline for container images and discovery node

noursaidi:
This creates new container images for the core UDMI images (`udmi/validator`, `udmi/udmis`, etc) which are built only on pushes to master.

It also builds the discoverynode binary images when these files have changed on pushes to master.

image

### DIFF
--- a/.github/workflows/release-discovery.yaml
+++ b/.github/workflows/release-discovery.yaml
@@ -6,7 +6,7 @@ on:
       - 'master'
     paths: 
      - misc/discoverynode/**
-     - .github.workflows/release-discovery.yaml
+     - .github/workflows/release-discovery.yaml
 
 jobs:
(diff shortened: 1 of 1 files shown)


jainrocks (approved):

.github/workflows/release-discovery.yaml
```diff
@@ -0,0 +1,84 @@
+name: UDMI Discovery Node Release
+
+on:
+  push:
+    branches:
+      - 'master'
+    paths: 
+     - misc/discoverynode/**
+     - .github.workflows/release-discovery.yaml
```
jainrocks, line 9: .github.workflows/release-discovery.yaml is this correct?
or it should be ".github/workflows/release-discovery.yaml"

PR #1125, opened 2025-06-05
Bambi Backend Service with Git Interface

khyatimahendru:


services/bin/bambi_service
```diff
@@ -15,17 +15,19 @@ LOCAL=
 NO_CHECK=
 [[ ${1:-} == "--no-check" ]] && NO_CHECK=1 && shift
 
+# shellcheck disable=SC2155
+export GCP_PROJECT=$(gcloud config get-value project)
+echo Using GCP_PROJECT "$GCP_PROJECT"
+
 if [[ -n ${UDMI_NAMESPACE:-} ]]; then
     UDMI_PREFIX="${UDMI_NAMESPACE}~"
+    TARGET="//gbos/${GCP_PROJECT}/${UDMI_NAMESPACE}"
```
grafnu, line 24: Hmmm... this might be a problem since a lot of things don't use `//gbos/` and I'm not sure we can/should hard-code it.

services/src/main/java/com/google/bos/iot/core/bambi/BambiService.java
```diff
@@ -0,0 +1,412 @@
...
+    this.udmiNamespace = spec.udmiNamespace;
+    this.baseCloningDir = siteModelBaseDir;
+    this.localOriginDir = localOriginDir;
+    this.pollingExecutor = Executors.newSingleThreadExecutor(
+        r -> new Thread(r, SERVICE_NAME + "-poller"));
+
+    String udmiNamespacePrefix = ofNullable(spec.udmiNamespace).map(ns -> ns + "~").orElse("");
+    String requestsSubscription = udmiNamespacePrefix + "bambi-requests";
+
+    if (MQTT.equals(protocol)) {
```
grafnu, line 91: This is making me twitchy since it's replicating logic that really should be centralized. E.g. BAMBI shouldn't know about the protocol if it's just using a MessagingClient.  The obvious solution would be to push this logic down into the MessagingClient abstraction so it's not in BAMBI, but I'm not sure if it's that simple... ?

services/src/main/java/com/google/bos/iot/core/bambi/BambiService.java
```diff
@@ -0,0 +1,412 @@
...
+    LOGGER.info("Cloning base directory: {}", baseCloningDir);
+    try {
+      Files.createDirectories(Paths.get(this.baseCloningDir));
+    } catch (IOException e) {
+      throw new IllegalStateException(
+          "Could not create site model clone directory: " + this.baseCloningDir, e);
+    }
+  }
+
+  private MessagingClient createMqttClient(String requestsSubscription) {
```
grafnu, line 184: Seems like this, and the next method, should be pushed in the MessagingClient side of the abstraciton.

common/src/main/java/com/google/udmi/util/SheetsAppender.java
```diff
@@ -0,0 +1,36 @@
...
+import ch.qos.logback.core.AppenderBase;
+import ch.qos.logback.core.encoder.Encoder;
+
+public class SheetsAppender extends AppenderBase<ILoggingEvent> {
+
+  private static SheetsOutputStream sheetsOutputStream;
+  private Encoder<ILoggingEvent> encoder;
+
+  @Override
+  protected void append(ILoggingEvent eventObject) {
```
grafnu, line 13: Where is this used? I'm trying to evaluate the wisdom of using the `ch.qos.` package... and I can't quite tell how it's being used!

services/src/main/java/com/google/bos/iot/core/bambi/BambiService.java
```diff
@@ -0,0 +1,412 @@
...
+
+  private void prepareDirectory(String directory) throws IOException {
+    File dirFile = new File(directory);
+    if (dirFile.exists()) {
+      LOGGER.info("Deleting existing directory {}", directory);
+      deleteDirectory(dirFile);
+    }
+  }
+
+  private GenericGitRepository getRepository(String repoName, String repoClonePath)
```
grafnu, line 387: This should be pushed down into the `Repository` abstraction, since it's not really a function of Bambi specifically.

validator/src/main/java/com/google/udmi/util/git/GitRepositoryInterface.java
```diff
@@ -0,0 +1,70 @@
...
+import org.eclipse.jgit.api.Status;
+import org.eclipse.jgit.api.errors.GitAPIException;
+import org.eclipse.jgit.revwalk.RevCommit;
+import org.eclipse.jgit.transport.FetchResult;
+import org.eclipse.jgit.transport.PushResult;
+
+/**
+ * Interface for a Git Repository.
+ */
+public interface GitRepositoryInterface extends AutoCloseable {
```
grafnu, line 14: At this point, I think it would be prudent to introduce one additional layer of abstraction which is just `sourceRepository` that doesn't expose _everything_ git has to offer, and doesn't have any explicit ties to git -- e.g. conceptualize it just as being for any repository.  Should be fairly easy to do -- just have the new interface only implement the actual functions that are used by BAMBI.   It's generally easier to add things later than it is to take things away (which might break something).

grafnu (commented):
Added some comments about the overall structure. There might be some tricky things to work through here, so have a think, we can chat a bit Tuesday, and then have a meet to talk through the nuances. In general (so not just with your PR), my biggest concern often has to do with blurring abstractions and things like that, and dependency management -- which is where most of the comments come from!

services/src/main/java/com/google/bos/iot/core/bambi/BambiService.java
khyatimahendru: Removed references to protocol

services/src/main/java/com/google/bos/iot/core/bambi/BambiService.java
khyatimahendru: Moved this to MessagingClient

services/src/main/java/com/google/bos/iot/core/bambi/auth/IdVerifier.java
```diff
@@ -0,0 +1,32 @@
+package com.google.bos.iot.core.bambi.auth;
+
+/**
+ * Interface for verifying identity for incoming requests.
+ */
+public interface IdVerifier {
+
+  String PUBSUB = "pubsub";
```
grafnu, line 8: is this the same as the project target //pubsub vs. //mqtt -- ?  If so, we should use those already-defined constants... ?

services/src/main/java/com/google/bos/iot/core/bambi/model/BambiSheet.java
```diff
@@ -0,0 +1,25 @@
+package com.google.bos.iot.core.bambi.model;
+
+/**
+ * Enum representing the different tabs/sheets in a BAMBI spreadsheet.
+ */
+public enum BambiSheet {
```
grafnu, line 6: `BambiSheetName`?  Just `BambiSheet` to me would be a thing that represented the entire sheet?

validator/src/main/java/com/google/udmi/util/messaging/MessagingClient.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * An interface for a generic messaging client, abstracting away the specifics
+ * of the underlying technology (e.g., Google Pub/Sub, MQTT).
+ */
+public interface MessagingClient extends Closeable {
+
+  String MQTT = "mqtt";
```
grafnu, line 14: I would take these from IoTProvider.java to have a single point of definition. It's not a 100% fit, but I think better than having extra string constants floating around.

validator/src/main/java/com/google/udmi/util/messaging/MessagingClient.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  /**
+   * Polls for a message from the subscription, waiting up to the specified timeout.
+   *
+   * @param timeout The maximum time to wait.
+   * @param unit The time unit of the timeout argument.
+   * @return The received PubsubMessage, or null if the timeout is reached.
+   */
+  PubsubMessage poll(long timeout, TimeUnit unit);
```
grafnu, line 36: prefer `Duration` over two args -- if that doesn't mess other things up too much.

services/src/main/java/com/google/bos/iot/core/bambi/model/SourceRepository.java
```diff
@@ -0,0 +1,141 @@
+package com.google.bos.iot.core.bambi.model;
```
grafnu, line 1: Is this specific to Bambi?  Seems like it should be down in the utils or common?

grafnu (approved):
Nothing major, so good to submit -- just a few naming things and location of classes etc...

services/src/main/java/com/google/bos/iot/core/bambi/auth/IdVerifier.java
khyatimahendru: Removed these and reused the enum from IotAccess.IotProvider as suggested

validator/src/main/java/com/google/udmi/util/messaging/MessagingClient.java
khyatimahendru: Removed these and reused the enum from IotAccess.IotProvider as suggested

services/src/main/java/com/google/bos/iot/core/bambi/model/BambiSheet.java
khyatimahendru: Renamed to BambiSheetTab

PR #1127, opened 2025-06-05
add tool for e2e registrar testing in situ

noursaidi:
this is a tool which will run regularly (say every 15 minutes) in live environments (e.g. production, staging, testing), which:

- Uses a reference site model (so there  could be several, maybe one small, one proxy device heavy, one normal device heavy)
- Purges the "testing" registry (one per source site model, e.g. ZZ-TST-ONE, ZZ-TST-TWO)
- Validates that the testing registry is empty
- Runs registrar against using the reference site model into the designated testing registry 
- Validates the new registry, that:
  - All auth correct provisioned (MQTT connections can be made, proxy devices can be attached)
  - Redacted (e.g. timestamps) in the cloud matches the local generated_config

The ultimate objective is to:
- Detect breaking changes in registrar or UDMIS with realistic scenarios (see comment on site model references)
- Make sure the cloud service is working

The invocation will be in some Kubernetes cron using the existing validator image (which pulls the bin/ scripts, so I'm hoping works) and log metrics which can detect number of errors and duration registrar takes to run.


bin/test_registrar_e2e
```diff
@@ -0,0 +1,77 @@
...
+# - The registry is empty (because the same registry is reused and not deleted/recreated)
+# - Test the deletion functionality
+$UDMI_ROOT/bin/registrar $SITE_MODEL_DIR $TARGET_PROJECT -x -d || err=y
+if [[ $err != 0 ]]; then
+  echo "Registrar exitted with error code $err"
+  false
+fi
+
+# Run registrar again, to fetch 0 devices to confirm the directory is empty
+$UDMI_ROOT/bin/registrar $SITE_MODEL_DIR $TARGET_PROJECT -x -d | tee $TMP_DIR/deletion.txt || err=y
```
grafnu, line 46: run w/o -x -d for the check step?

grafnu (approved):

PR #1128, opened 2025-06-09
Changes in the mapping agent to map the extra unknown discovered devices to devices in the site model 

jainrocks:
Changes in the mapping agent to map the extra unknown discovered devices to devices in the site model


validator/src/main/java/com/google/daq/mqtt/mapping/MappingAgent.java
```diff
@@ -141,25 +144,66 @@ private FamilyDiscoveryConfig getFamilyDiscoveryConfig(String family) {
...
+    List<Entry<String, Metadata>> mappedDiscoveredEntries = getMappedDiscoveredEntries();
+    Map<String, Metadata> devicesEntriesMap = getDevicesEntries();
+
+    mappedDiscoveredEntries.forEach(entry -> {
+
+      if (devicesEntriesMap.containsKey(entry.getKey())) {
+        System.err.println("Skipping existing device file for family::address = " + entry.getKey());
+        //TODO: update the existing device
+      } else {
+        // UNK denotes unknown device
```
grafnu, line 157: I'd extract the "UNK-" to a named class constant.  (And then you could likely remove the comment too since it would be code-as-commend.)

validator/src/main/java/com/google/daq/mqtt/mapping/MappingAgent.java
```diff
@@ -141,25 +144,66 @@ private FamilyDiscoveryConfig getFamilyDiscoveryConfig(String family) {
...
+        String newDeviceName = "UNK-" + suffixToStart;
+        suffixToStart++;
```
grafnu, line 159: I think you might need something a bit more sophisticated here... e.g. need to cehck the devicesEntryMap to find the next not-used UNK- entry, otherwise this might clobber existing UNK- devices.

validator/src/main/java/com/google/daq/mqtt/mapping/MappingAgent.java
```diff
@@ -141,25 +144,66 @@ private FamilyDiscoveryConfig getFamilyDiscoveryConfig(String family) {
...
-        System.err.println("Skipping existing device file " + metadataFile);
-      } else {
-        System.err.println("Writing device metadata file " + metadataFile);
-        metadataFile.getParentFile().mkdirs();
-        JsonUtil.writeFile(entry.getValue(), metadataFile);
+    return mappedDiscoveredEntries;
+  }
+
+  private Map<String, Metadata> getDevicesEntries() {
+    File devicesDir = siteModel.getDevicesDir();
```
grafnu, line 183: Site model should already have a function I think that gets all the devices?  I think it's `public Set<String> getDeviceIds()`

validator/src/main/java/com/google/daq/mqtt/mapping/MappingAgent.java
```diff
@@ -141,25 +144,66 @@ private FamilyDiscoveryConfig getFamilyDiscoveryConfig(String family) {
...
+      if (devicesEntriesMap.containsKey(entry.getKey())) {
+        System.err.println("Skipping existing device file for family::address = " + entry.getKey());
+        //TODO: update the existing device
+      } else {
+        // UNK denotes unknown device
+        String newDeviceName = "UNK-" + suffixToStart;
+        suffixToStart++;
+        File metadataFile = siteModel.getDeviceFile(newDeviceName, METADATA_JSON);
+        System.err.println("Writing device metadata file " + metadataFile);
+        metadataFile.getParentFile().mkdirs();
```
grafnu, line 162: I think we should push this functionality (create a new device) into SiteModel... ?

validator/src/main/java/com/google/daq/mqtt/mapping/MappingAgent.java
```diff
@@ -141,25 +144,66 @@ private FamilyDiscoveryConfig getFamilyDiscoveryConfig(String family) {
...
+        JsonUtil.writeFile(entry.getValue(), metadataFile);
+      }
+    });
+
+    updateProxyIdsForDiscoveryNode(mappedDiscoveredEntries);
+  }
+
+  private List<Entry<String, Metadata>> getMappedDiscoveredEntries() {
     File extrasDir = siteModel.getExtrasDir();
```
grafnu, line 171: Yah -- so I know you're just copying/extending what's already there, but I think now might be a good time to clean it up! (You know, do as I say not as I did yadda yadda...). YDYT?

validator/src/main/java/com/google/daq/mqtt/mapping/MappingAgent.java
```diff
@@ -185,8 +234,8 @@ private Entry<String, Metadata> convertExtra(File file) {
     metadata.system = new SystemModel();
     metadata.gateway = new GatewayModel();
     metadata.gateway.gateway_id = deviceId;
-    String deviceName = (String) discoveryEvents.system.ancillary.get("device-name");
-    return Map.entry(deviceName, metadata);
+    String familyWithAddress = discoveryEvents.family + "::" + discoveryEvents.addr;
```
grafnu, line 237: extract this to a function so it can be shared with the other place it's used, so it's obvious that they're the same thing!

validator/src/main/java/com/google/daq/mqtt/mapping/MappingAgent.java
jainrocks: sure, Thanks. Will check on this

validator/src/main/java/com/google/daq/mqtt/mapping/MappingAgent.java
jainrocks: that makes sense, will do

validator/src/main/java/com/google/daq/mqtt/mapping/MappingAgent.java
jainrocks: yup :)

PR #1136, opened 2025-06-23
discovery node improvements and fixes

noursaidi:
- fix deployment pipeline using wrong image
- better handling and logging of failed packet capture start
- modular arithemtic calcs for past generation with interval rate
- add ping scan
- minor bug fixes/code improvements (mainly docstrings and types)


misc/discoverynode/src/udmi/discovery/discovery.py
```diff
@@ -323,41 +340,81 @@ def controller(self, config_dict: None | dict[str:Any]):
...
+        scheduled_start = generation_from_config.timestamp()
+  
+      elif seconds_from_now < MAX_THRESHOLD_GENERATION and config.scan_interval_sec:
+        # Generation is in the past, but it has interval
+        # Use modular arithemtic to determine when the start time is
+        cycles_elapsed, seconds_into_cycle = divmod(abs(seconds_from_now), config.scan_interval_sec)
+
+        # determine whether to join this cycle or wait till the next
+        cycle_modifer = 1 if seconds_into_cycle > abs(MAX_THRESHOLD_GENERATION) else 0
+        logging.warning(f"scan interval: {config.scan_interval_sec}")
```
noursaidi, line 369: note to self change to debug log lines

misc/discoverynode/src/udmi/discovery/ether.py
```diff
@@ -0,0 +1,173 @@
...
+    self.nmap_thread.start()
+
+  def nmap_stop_discovery(self):
+    logging.info("stopping")
+    self.cancel_threads.set()
+    self.nmap_thread.join()
+
+  @discovery.catch_exceptions_to_state
+  @discovery.main_task
+  def nmap_runner(self):
```
noursaidi, line 128: note to self: change this to take config.addrs
note to reviewrs: this all used to be in `nmap.py` which was renamed

misc/discoverynode/src/udmi/discovery/passive.py
```diff
@@ -70,9 +71,17 @@ def _get_packet_counter_total(self):
     except FileNotFoundError:
       return
 
+  def scapy_is_go(self): 
+    # Signal that packet capture has started. Callback executed by scapy
+    # library on success. Necesary because exception disapears into the
+    # the aether and we never no.
```
noursaidi, line 77: never know

misc/discoverynode/src/udmi/discovery/ether.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+class EtherDiscovery(discovery.DiscoveryController):
+  """Ether Network Discovery."""
+
+  family = "ether"
+
+  def __init__(self, state, publisher, *, target_ips: list[str] = None):
+    self.cancel_threads = threading.Event()
+    self.target_ips = target_ips
```
khyatimahendru, line 22: `self.target_ips` is assigned twice (line 22 & 26)

misc/discoverynode/src/udmi/discovery/ether.py
```diff
@@ -0,0 +1,173 @@
...
+          break
+
+  def ping_task(self, target_ip: str) -> bool:
+    """Sends a single ping to the target IP. Returns True if successful."""
+    logging.debug("ping task %s", target_ip)
+    try:
+      # -c 1: expect 1 packet
+      # -W 2: wait 2 seconds for a response
+      result = subprocess.run(
+          ["/usr/bin/ping", "-c", "1", "-2", "2", target_ip],
```
khyatimahendru, line 79: The '-2' looks incorrect, should it be '-w'?

misc/discoverynode/src/udmi/discovery/ether.py
```diff
@@ -0,0 +1,173 @@
...
+
+    with subprocess.Popen(
+        [
+            "/usr/bin/nmap",
+            "--script",
+            "banner",
+            "-p-",
+            "-T4",
+            "-A",
+            self.target_ips[0],
```
khyatimahendru, line 139: Do we only want to use the first element from `self.target_ips`? Or is it that the list will contain just one element?

misc/discoverynode/src/udmi/discovery/ether.py
```diff
@@ -0,0 +1,173 @@
...
+    self.nmap_thread.join()
+
+  @discovery.catch_exceptions_to_state
+  @discovery.main_task
+  def nmap_runner(self):
+    OUTPUT_FILE = "nmaplocalhost.xml"
+
+    with subprocess.Popen(
+        [
+            "/usr/bin/nmap",
```
khyatimahendru, line 133: Just a suggestion, you don't need to change it if it works - considering that we have different environments, it might be better to replace the hardcoded path with `shutil.which("nmap")`

misc/discoverynode/src/udmi/discovery/ether.py
noursaidi: removed (legacy)

misc/discoverynode/src/udmi/discovery/ether.py
noursaidi: Thanks, that's a good idea, I didn't think of it. I'll track it as an improvement for a future PR

misc/discoverynode/src/udmi/discovery/discovery.py
```diff
@@ -39,6 +39,7 @@ def _impl(self, *args, **kwargs):
     try:
       return method(self, *args, **kwargs)
     except Exception as err:
+      raise(err)
```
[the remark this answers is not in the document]
noursaidi: No, that was an accidental commit, thanks

PR #1138, opened 2025-06-25
Update num_id with -u flag

grafnu:


validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
```diff
@@ -357,6 +364,29 @@ protected void execute(Runnable modelMunger) {
...
+    siteModel.forEachMetadata((deviceId, metadata) -> {
+      CloudModel registeredDevice = cloudIotManager.getRegisteredDevice(deviceId);
+      if (registeredDevice == null) {
+        return;
+      }
+      Metadata localMetadata = siteModel.getMetadata(deviceId);
+      if (localMetadata.cloud == null) {
+        localMetadata.cloud = new CloudModel();
+      }
+      String localId = localMetadata.cloud.num_id;
```
jainrocks, line 379: if localMetadat.cloud was null and if new CloudModel is assigned in line 377, `localMetadata.cloud.num_id` might be null

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
```diff
@@ -357,6 +364,29 @@ protected void execute(Runnable modelMunger) {
     }
   }
 
+  private void updateDeviceMetadata() {
+    checkNotNull(projectId, "can't update metadata: cloud project not defined");
+    AtomicInteger updatedCount = new AtomicInteger();
```
jainrocks, line 369: I am just thinking, if this should be the class variable instead of local method variable

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
jainrocks: Okay, got it, can handle .forEachMetadata part, removing this comment, removing this comment

jainrocks (approved):
LGTM

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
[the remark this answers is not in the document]
grafnu: But that's ok -- localId is written with this flow -- the read is only for diagnostic (logging) purposes!

PR #1142, opened 2025-06-27
Update model_system.json

ViktorasCes:
Changed model_system.json tags regex pattern to allow tags with underscores


ViktorasCes:
Hey Trevor, if I recall correctly, you've implemented this change via a separate PR because of the issues I've faced with `bin/gencode` on codespaces. I've checked and all changes proposed in here are already included. I'll close this one out.

PR #1147, opened 2025-07-02
Changes for fetching the model and pushing the model to proposal branch post the mapping processs

jainrocks:
Changes for fetching the model and pushing the model to proposal branch post the mapping process

The arguments passed to Mapping Service:
registryId: e.g. ZZ-TRI-FECTA
projectId: GCP project id
baseCloningDirectory: local path for the repository
projectSpec: eg. //mqtt/localhost
mappingFamily: e.g. vendor/bacnet
discoveryNodeDeviceId: discoveryNode deviceId e.g. GAT-123

### DIFF
--- a/services/bin/mapping_service
+++ b/services/bin/mapping_service
@@ -3,7 +3,7 @@
 UDMI_ROOT=$(dirname $0)/../..
 source $UDMI_ROOT/etc/shell_common.sh
 
-if [[ $# -lt 6 ]]; then
+if [[ $# -ne 6 ]]; then
     usage registry_id project_id base_cloning_directory project_spec mapping_family discoveryNodeDeviceId
 fi
 


bin/mapping_service
grafnu, line 1: I think there's a place for the actual service stuff in `services/` -- so should be `services/bin/mapping_service`

validator/src/main/java/com/google/daq/mqtt/mapping/MappingService.java
```diff
@@ -0,0 +1,92 @@
...
+    String exportBranch = String.format("proposal/%s/%s", SERVICE_NAME, timestamp);
+    if (!sourceRepository.checkoutNewBranch(exportBranch)) {
+      throw new RuntimeException("Unable to create and checkout export branch " + exportBranch);
+    }
+
+    String udmiModelPath = Paths.get(sourceRepository.getDirectory(), "udmi").toString();
+    (new Registrar()).processArgs(new ArrayList<>(List.of(udmiModelPath, projectSpec))).execute();
+    MappingAgent mappingAgent = new MappingAgent(new ArrayList<>(List.of(udmiModelPath, projectSpec)));
+    mappingAgent.setDiscoveryNodeDeviceId(this.discoveryNodeDeviceId);
+    mappingAgent.process(new ArrayList<>(List.of("map", mappingFamily)));
```
grafnu, line 83: This seems a bit odd to be "simulating" a command line when it's a direct call on the class.  I'd consider extracting a method inside of MappingAgent that this can call directly, or at the _very_ least, if that's too messy for some reason (can't think why), use a named constant for "map" so it's crystal clear what the string is doing.

grafnu (commented):
A few comments and then all sorts of lint errors!

PR #1149, opened 2025-07-07
Add containers for reconciler and registrar services & related changes to bin/container script

khyatimahendru:
Summary of changes

* Add support to `bin/container` for containerization of submodules. So both type of invocations are supported:
    - `bin/container services.bambi push`
    - `bin/container udmis push`
* Change the `--no-check` flag in the service scripts to `--create`. The services would NOT create the pubsub resources if the `--create` flag is not passed.
* The service scripts now accept inputs for the messaging_spec, registrar_target, cloning_dir, etc. Invocations look like below:
   - `services/bin/bambi_service message_spec cloning_dir [--local] [--create] [local_orig_dir]`
   - `services/bin/reconciler_service message_spec cloning_dir [--local] [--create] [local_orig_dir]`
   - `services/bin/registrar_service message_spec registrar_target cloning_dir [--local] [--create] [local_orig_dir]`
* All the services now write to `/tmp/` instead of `/root/`
* The names of all pubsub topics and subscriptions are standardized to include the prefix `udmi_` and use underscores instead of hyphens.


bin/container
```diff
@@ -10,6 +10,20 @@ fi
 target=$1
 cmd=$2
 shift 2
+target_tag=${target}
+
+if [[ "$target" == *.* ]]; then
+  sub_target=${target#*.}
+  target=${target%%.*}
+  dockerfile="Dockerfile.${target}.${sub_target}"
```
grafnu, line 18: does it still work if `_` is used to join with sub_target?  For some reason the double-`.` notation is making me twitchy... so this would end up something like `Dockerfile.services_registrar`?  Or, does that actively seem less good to you?

services/etc/registrar/k8s_config.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config
+data:
+  BITBOX_DISCOVERY:
+  ETCD_CLUSTER:
+  UDMI_NAMESPACE: @UDMI_NAMESPACE@
+  REGISTRAR_IOT_PROVIDER: gbos
```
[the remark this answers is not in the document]
grafnu: If this is a service that's running in GCP then `gref` would be better... but, wouldn't that break something?  I'm just confused as to how both work equally as well.  But, `gref` should only really be used from the "outside" since it's a different authentication domain.

services/etc/registrar/k8s_config.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config
+data:
+  BITBOX_DISCOVERY:
```
grafnu, line 6: Is `BITBOX_DISCOVERY` actually needed?  I forget how all this is used, but my thinking is that if it's just leftover cruft then we should take it out now... what happens if it's removed from here?

services/etc/registrar/k8s_registrar.yaml
```diff
@@ -0,0 +1,39 @@
...
+            cpu: 500m
+            memory: 500Mi
+        env:
+        - name: UDMI_NAMESPACE
+          valueFrom:
+            configMapKeyRef:
+              name: config
+              key: UDMI_NAMESPACE
+              optional: true
+        - name: REGISTRAR_IOT_PROVIDER
```
grafnu, line 34: `REGISTRAR_` prefix seems redundant -- is that actually needed?

bin/container
khyatimahendru: Underscore works too! I will make the change

services/etc/registrar/k8s_config.yaml
khyatimahendru: Okay, I did not know this is not used. I will test and remove it.

grafnu (approved):

PR #1152, opened 2025-07-09
Add discovery marker events

grafnu:


pubber/src/main/java/udmi/lib/client/manager/DiscoveryManager.java
```diff
@@ -289,30 +289,47 @@ default void startDiscoveryScan(String family, Date scanGeneration) {
...
     familyDiscoveryState.active_count = sendCount.get();
+    sendMarkerDiscoveryEvent(family, familyDiscoveryState, scanGeneration);
     updateState();
     startDiscoveryForFamily(family, scanGeneration, familyDiscoveryState, sendCount);
   }
 
+  private void sendMarkerDiscoveryEvent(String family, FamilyDiscoveryState familyDiscoveryState,
+      Date scanGeneration) {
+    DiscoveryEvents discoveryEvent = new DiscoveryEvents();
+    boolean isActive = familyDiscoveryState.phase == ACTIVE;
```
jainrocks, line 300: Should we check for STOP state to send negative event No?

pubber/src/main/java/udmi/lib/client/manager/DiscoveryManager.java
```diff
@@ -289,30 +289,47 @@ default void startDiscoveryScan(String family, Date scanGeneration) {
...
+    discoveryEvent.event_no = (isActive ? 1 : -1) * familyDiscoveryState.active_count;
```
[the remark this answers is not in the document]
grafnu: The registry ID is implicitly included in the message attributes, e.g. see validator/sequences/scan_single_now/events_discovery.attr

validator/sequences/scan_single_future/sequence.log
```diff
@@ -1,2562 +1,1942 @@
-2024-12-17T05:26:04Z NOTICE  starting test scan_single_future ################################
```
jainrocks, line 1: Do we need log files to be pushed to git?

pubber/src/main/java/udmi/lib/client/manager/DiscoveryManager.java
grafnu: Updated logic doesn't require a state check.

validator/sequences/scan_single_future/sequence.log
grafnu: These files are there as a reference (i.e., automatically generated documentation), just in case somebody wants to look at a sample log of a "known good" run (to compare with a not-working run when developing a device)

khyatimahendru (approved):

jainrocks (approved):
LGTM

PR #1153, opened 2025-07-10
RAD-5926 persistent MQTT subscription configuration On/Off 

pablinchapin:
Adding additionalProperties on DevicePersistent definition, to store the MQTT configuration subscription state

### DIFF
--- a/gencode/java/udmi/schema/DevicePersistent.java
+++ b/gencode/java/udmi/schema/DevicePersistent.java
@@ -1,15 +1,10 @@
 
 package udmi.schema;
 
-import com.fasterxml.jackson.annotation.JsonAnyGetter;
-import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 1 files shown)


grafnu (commented):
Is this a manual edit to that one file?  Everything in the `gencode/` directory is automatically generated and can't be manually edited. If you need to make changes to the device persistence class, you should edit `schema/persistent_device.json` and then run `bin/gencode` to update the `gencode/` files. 

Conceptually I'm not a big fan of a generic bucket of "additional properties" because there's no way to really know what it might be for and what might get clobbered by something else.  Is there a description or motivation for what this is for, exactly?  If at all possible, I'd like to keep it using named properties rather than a generic map.

grafnu:
One additional thought was maybe to just rename the bucket to `vendor_properties` (rather than `additional_properties`) to just clearly indicate that the contents of the bucket are for vendor-side use only, so it won't conflict with anything outside of the vendor scope. Would still have to update the schema and go that route, but then wouldn't require explicitly naming anything.

PR #1159, opened 2025-07-25
Add registrar option -o to process only changed devices

khyatimahendru:


validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
```diff
@@ -377,6 +391,16 @@ protected void execute(Runnable modelMunger) {
     }
   }
 
+  private void loadSiteRegistrationTimestamps() {
+    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss'Z'")
+        .withZone(ZoneOffset.UTC);
+    currentRunTimestamp = formatter.format(Instant.now());
```
grafnu, line 397: already a function for this `isoConvert` somewhere in `Common`

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
```diff
@@ -185,6 +190,9 @@ public class Registrar {
   private boolean doNotUpdate;
   private boolean expandDependencies;
   private boolean updateMetadata;
+  private String currentRunTimestamp;
```
grafnu, line 193: nit: I'd either make both these `Instant` or both `String` (iso) -- seems weird to have them different.  Just weird, not necessarily bad so if it doesn't make sense then don't worry about it.

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
```diff
@@ -653,6 +679,23 @@ private Map<String, LocalDevice> loadAllDevices() {
...
   }
 
+  private Set<String> loadChangedDevices() {
+    if (optimizeRun && lastRunInstant != null) {
+      System.err.println("Collecting devices changed after " + lastRunInstant);
+      return allDevices.entrySet().stream()
+          .filter(entry -> {
+            LocalDevice device = entry.getValue();
+            Metadata metadata = device.getMetadata();
+            if (metadata == null || metadata.timestamp == null) {
```
grafnu, line 689: seems like you could combine this if with the following return since it's all boolean logic?

grafnu (approved):

validator/src/main/java/com/google/daq/mqtt/registrar/Registrar.java
khyatimahendru: Thank you! I have changed this to use `isoConvert`

PR #1160, opened 2025-07-25
Initiate mapping process on consuming discovery complete triggger

jainrocks:
1. Consume the events from udmi_target Pub/Sub topic
2. Initiate the mapping process if event number is negative


services/src/main/java/com/google/daq/mqtt/mapping/MappingService.java
```diff
@@ -1,102 +1,126 @@
...
-    String repoCloneDir = Paths.get(baseCloningDirectory, registryId).toString();
-    sourceRepository = new SourceRepository(registryId, repoCloneDir, null, projectId, null);
+        if (registryId.isEmpty()) {
+          LOGGER.error("Registry Id not found for the message.");
+          return;
+        }
 
-  }
+        String discoveryNodeDeviceId = message.getAttributesOrDefault(DISCOVERY_NODE_DEVICE_ID_FIELD, "");
+        if (discoveryNodeDeviceId.isEmpty()) {
```
khyatimahendru, line 92: Do we want to process further if discovery node is empty? I think the Mapping Agent does some attribute checks later on the discovery node and further operations might not go through. Better to fail fast!

services/bin/mapping_service
```diff
@@ -1,24 +1,81 @@
...
+                is_local=1
+                shift
+                ;;
+            --create)
+                should_create=1
+                shift
+                ;;
+            *)
+                if [[ -z "$local_origin_dir" ]]; then
+                    local_origin_dir="$1"
```
grafnu, line 43: this should also check for something that starts with `-` and generate an error (e.g. unknown command line option)

services/bin/mapping_service
```diff
@@ -1,24 +1,81 @@
...
+        ensure_pubsub_topic_exists "$repo_updates_topic"
+        ensure_pubsub_subscription_exists "$repo_updates_subscription" "$repo_updates_topic" --no-expiration
+    fi
+
+    echo "Starting ${service_name} Service..."
+    java -cp "$SERVICES_JAR" "com.google.daq.mqtt.mapping.MappingService" \
+        "$message_spec" "$project_spec" "$cloning_dir" "$local_origin_dir"
+}
+
+main "$@"
```
grafnu, line 81: what's the utility of wrapping that all in `main`? -- it seems like it's just adding extra indent and then non-linear program flow?

services/src/main/java/com/google/daq/mqtt/mapping/MappingService.java
```diff
@@ -1,102 +1,126 @@
...
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Mapping Service: which currently runs from command line
+ * Mapping Service basically
```
grafnu, line 20: replace `basically` with `that`, I think -- otherwise it reads oddly informal...

services/src/main/java/com/google/daq/mqtt/mapping/MappingService.java
```diff
@@ -1,102 +1,126 @@
...
+      System.err.println(
+          "Usage: MappingService <projectTarget> <registrarTarget> <siteModelCloneDir> "
+              + "[<localOriginDir>]");
+      System.exit(1);
+    }
+
+    String projectTarget = args[0];
+    String projectSpec = args[1];
+    String siteModelCloneDir = args[2];
+    String localOriginDir =  (args.length == 4) && isNotEmpty(args[3]) ? args[3] : null;
```
grafnu, line 50: extra space after `=`

services/src/main/java/com/google/daq/mqtt/mapping/MappingService.java
```diff
@@ -1,102 +1,126 @@
...
-   * projectSpec: eg. //mqtt/localhost
-   * mappingFamily:vendor/bacnet
-   * discoveryNodeDeviceId: discoveryNode deviceId e.g. GAT-123
+   * Main entry point for the application.
    */
   public static void main(String[] args) {
-    MappingService mappingService = new MappingService(args);
-    mappingService.initialize();
+    if (args.length < 3 || args.length > 4) {
+      System.err.println(
```
grafnu, line 41: Use `LOGGER`?

services/src/main/java/com/google/daq/mqtt/mapping/MappingService.java
```diff
@@ -1,102 +1,126 @@
...
+          if (!repository.commitAndPush("Merge changes from source: MappingService")) {
+            throw new RuntimeException("Unable to commit and push changes to branch " + exportBranch);
+          }
+          LOGGER.info("Export operation complete.");
+          repository.delete();
+        } else {
+          LOGGER.error("Could not clone repository! PR message was not published!");
+        }
+      }
+    } catch (IOException e) {
```
grafnu, line 122: try/catch is redundant with code in `AbstractPollingService` so I'd take it out. If there's something specific that's throwing the IOException I'd narrow down the try/catch to just that one thing and add some more meaningful detail? (If it's coming from the SourceRepsitory class I'd push it down into that?)

services/bin/mapping_service
jainrocks: More for code organization and maintainability, and we are using similar scripts for the reconciler, bambi and registrar service.

services/bin/mapping_service
jainrocks: sure

grafnu (approved):

PR #1166, opened 2025-07-31
discovery node - less logging, print device/registry ID with public key, more bacnet properties

noursaidi:


pisuke (approved):
Thanks for adding these new BACnet fields, all good for me.

PR #1177, opened 2025-08-14
Fix registrar error when a file is present in the devices folder

khyatimahendru:


pisuke (approved):

common/src/main/java/com/google/udmi/util/SiteModel.java
```diff
@@ -275,12 +277,13 @@ public static List<String> listDevices(File devicesDir) {
...
     String[] devices = requireNonNull(devicesDir.list());
-    return Arrays.stream(devices).filter(SiteModel::validDeviceDirectory)
+    return Arrays.stream(devices).filter(device -> validDeviceDirectory(devicesDir, device))
         .collect(Collectors.toList());
   }
 
-  private static boolean validDeviceDirectory(String dirName) {
-    return !(dirName.startsWith(".") || dirName.endsWith("~"));
+  private static boolean validDeviceDirectory(File baseDir, String dirName) {
+    return !(dirName.startsWith(".") || dirName.endsWith("~")) &&
```
grafnu, line 285: The bug is correct in asserting this should be a "warning" of some sort, but I'm not sure we currently have a mechanism in place to have site-model level warnings (or maybe we do, I honestly can't remember). But, if not, I think a better intermediate solution would be to just have a more clear error message?

I can't think of any legit reason to have a stray file in that directory.

I actually can't even remember why this logic is here in the first place -- e.g. the `startsWith` and `endsWith` checks -- seems like those should be errors/warnings as well..

grafnu (commented):
Not entirely a bad change, my concern is that it actually moves the needle in the wrong direction with being unnecessarily permissive of stray junk in the site_model...

common/src/main/java/com/google/udmi/util/SiteModel.java
```diff
@@ -104,6 +107,9 @@ public class SiteModel {
   private static final File CONFIG_OUT_DIR = new File("out/");
   private static final String RSA_PRIVATE_KEY = "rsa_private.pkcs8";
   private static final String EC_PRIVATE_KEY = "ec_private.pkcs8";
+  private static final Set<String> ALLOWED_FILES_IN_DEVICES_DIR = new HashSet<>(List.of(
```
grafnu, line 110: ImmutableSet.of()?

grafnu (approved):

PR #1182, opened 2025-08-19
Add validation for family providers ether, ipv4, ipv6, fqdn

khyatimahendru:


grafnu (approved):
Ideally we'd get some tests in place like there is for BACNet, but the actual code here looks good so I'll leave that up to you -- so maybe either add some (unit) tests like BACnet has or create a bug to add tests...

khyatimahendru:
> Ideally we'd get some tests in place like there is for BACNet, but the actual code here looks good so I'll leave that up to you -- so maybe either add some (unit) tests like BACnet has or create a bug to add tests...

Thank you, I added tests for each provider

PR #1186, opened 2025-09-15
Importing sites numIds

jainrocks:


bin/import_site_num_ids
```diff
@@ -0,0 +1,156 @@
...
+
+    DEVICES_PATH="$SITE_PATH/udmi/devices"
+
+    if [ ! -d "$DEVICES_PATH" ]; then
+        echo "  WARNING: Devices directory not found at '$DEVICES_PATH'. Skipping device-level operations."
+        continue
+    fi
+    
+    # Step A: Normalize metadata.json files in the devices directory.
+    echo "Normalizing metadata for $site_model..."
```
khyatimahendru, line 94: Can we have a separate script for normalizing metadata? Normalizing might introduce a lot of changes and the diff might become difficult to review. Normalizing as a separate step will increase the confidence in review when adding the actual num ids

PR #1278, opened 2026-04-02
Harmonize reflector topics between `mqtt` and `gbos` providers

noursaidi:
The MQTT topics for sending `reflect` messages differs between the mqtt provider e.g. `//mqtt/localhost` and the gbos provider e.g. ``//gbos/project`.

There is no reason for the difference apart. I believe the topic structure after the topic prefix should match within UDMI irrespective of the transport.

This PR aligns the two. The driver here is that when setting UDMIS to the `implicit` provider aka mosquitto, reflect clients connecting the current `gbos` provider do not work.

Alternatively, we could just make the clients more flexible to support any authentication mechanism and topic structure, however as per above statement I don't think there should be a difference here.


grafnu (commented):
The reason the topic structure was different was somewhat just so that we knew it _could_ be different in the MQTT space. Not sure if it's better to be dynamic and flexible, or strict and pedantic. Both have their uses. But, bottom line is I'm not sure it matters for now and honestly I'm entirely sure what you're referring to, specifically, since it still seems like there's two fundamentally different topic structures, but that's ok.

grafnu:
What's the status of this PR?  I'd like to close it out or close it down just to keep things moving!

PR #1281, opened 2026-04-07
Sequencer tests for OTA blob updates

khyatimahendru:


docs/specs/sequences/generated.md
```diff
@@ -445,6 +447,27 @@ Check that a gateway proxies state updates for indicated devices
...
+
+1. Update config trigger ota update for pubber_module
+    * Add `blobset` = { "blobs": { "pubber_module": { "phase": `final`, "generation": `blob generation`, "sha256": `blob data hash`, "url": `software data` } } }
+1. Wait for pubber_module phase is APPLY
+1. Wait until system logs level `ERROR` category `blobset.blob.fetch.failure`
+1. Wait for pubber_module phase is FINAL and status is not null
+
+Test passed.
+
+## ota_happy_path (PREVIEW)
```
grafnu: `ota_success`?

docs/specs/categories.md
```diff
@@ -58,6 +58,13 @@ could be used with a **WARNING** level for a warning case during normal operatio
   * _blob_: Conditions specific to an individual blob
     * _receive_: (**DEBUG**) About receiving a blob update
     * _fetch_: (**DEBUG**) Fetching a blob update
+      * _success_: (**DEBUG**) Successfully fetched a blob
```
grafnu, line 61: INFO?

docs/specs/categories.md
```diff
@@ -58,6 +58,13 @@ could be used with a **WARNING** level for a warning case during normal operatio
...
+      * _failure_: (**ERROR**) Blob fetch failed
+    * _verify_: Verifying a blob
+      * _parse_: (**ERROR**) Error parsing blob
```
grafnu, line 64: some kind of INFO level for logging?  These categories aren't just error states, but also logging. Not totally consistent, but something to at least pretend we understand.

schema/options_pubber.json
```diff
@@ -127,6 +127,14 @@
     },
     "softwareFirmwareValue": {
       "type": "string"
+    },
+    "hardwareIncompatible": {
```
grafnu, line 131: would it be easy/possible to encode these states in the blob that sent instead?  Having pubber options just ends up being unwieldy in a number of ways.  Not sure it's worth the extra work, but thought I'd at least ask. E.g., the blob would be successful in all other respects, but pubber would look at the contents for specific values and use that to decide what error to throw?

grafnu (commented):
Still marked as `draft`...?

docs/specs/sequences/generated.md
khyatimahendru: I have fixed this by regenerating the golden files. This file was outdated.

PR #1285, opened 2026-04-08
docs: clarify MQTT password-based authentication credentials

jainrocks:
This PR clarifies the credential requirements for MQTT password-based authentication in the client documentation, explicitly stating the password derivation from the private key and the username formatting.

---
*PR created automatically by Jules for task [1912876625227867514](https://jules.google.com/task/1912876625227867514) started by @jainrocks*


noursaidi:
I don't think we should document this because it's not _ideal_ behavior. 

Typically when client certificates are used (as they are in the current configuration) passwords would not be used, and I believe that mosquitto doesn't validate passwords when client certificates are provided

PR #1299, opened 2026-04-17
feat: support Modbus TCP URL format

grafnu:
This change implements support for the new Modbus TCP URL format as defined in the task, while ensuring backward compatibility with the existing Modbus RTU format. 

Key changes:
- `validator/src/main/java/com/google/daq/mqtt/util/providers/ModbusFamilyProvider.java`: Updated `validateUrl` method to validate against both TCP and RTU regular expression patterns.
- `validator/src/test/java/com/google/daq/mqtt/util/providers/ModbusFamilyProviderTest.java`: Added comprehensive unit tests for valid and invalid references in both TCP and RTU formats.
- `docs/specs/modbus.md`: Documented the new TCP URL structure alongside the existing RTU structure.

---
*PR created automatically by Jules for task [3219882871980131686](https://jules.google.com/task/3219882871980131686) started by @grafnu*


grafnu:
The TCP and RTU alternatives of the URL should be the same, the differentiation will be handled as part of the network identifier. Use a semantic encoding for the register type and any other non-scalar field.  For TCP, use the network identifier instead of host, so the initial part would be `modbus://network/` (the same as for RTU).

grafnu:
Modify the URL to be `modbus://[network@]host[:port]/` --- which is fully compliant with the URL standard.

grafnu:
The interpreatation parameters need to be single words and have concrete values (examples don't make a good specification), and it needs to be easily URL-parsable.

PR #1301, opened 2026-04-17
CI tests for bridgehead

jainrocks:


.github/workflows/testing.yml
```diff
@@ -458,3 +454,26 @@ jobs:
...
+        with:
+          distribution: 'temurin'
+          java-version: '21'
+      - name: Setup Base
+        run: bin/setup_base
+      - name: Run Bridgehead Test
+        run: |
+          bash bridgehead/tests/test_bridgehead.sh
+      - name: Upload logs on failure
+        if: ${{ failure() }}
```
grafnu: Upload in all cases except for cancelled (look for other uses of upload in this file) -- there's no real reason not to, and it makes diagnostics easier because the output is regular (file is always there)

jainrocks: sure, updated.

grafnu (approved):
I tried to leave a comment in the file, but I can't tell if GitHub is broken or their UI is just bad.

nb: normally I'd just make test_bridgehead executable (like other test_scripts floating around) and drop the sh, so it could just be `bridgehead/tests/test_bridgehead` -- because sometimes it might just be a python test runner.  Functionally doesn't matter, but just something to consider.

PR #1308, opened 2026-04-27
Mosquitto acl support for Mosquitto GKE setup

jainrocks:


bin/mosquctl_gateway
```diff
@@ -0,0 +1,45 @@
...
+    # Create role if not exists (ignore error if exists)
+    $MOSQUITTO_CTRL createRole $role_name || true
+    # Add role to gateway client (ignore error if already added)
+    $MOSQUITTO_CTRL addClientRole $gateway_id $role_name || true
+    
+    # Add ACLs
+    $MOSQUITTO_CTRL addRoleACL $role_name subscribePattern "$device_id/config" allow
+    $MOSQUITTO_CTRL addRoleACL $role_name subscribePattern "$device_id/commands" allow
+    $MOSQUITTO_CTRL addRoleACL $role_name subscribePattern "$device_id/errors" allow
+    $MOSQUITTO_CTRL addRoleACL $role_name publishClientSend "$device_id/events" allow
```
noursaidi: this needs to be the wildcard topic `$device_id/events/#`

noursaidi:
We should call mosquitto_ctrl directly and not use shell scripts, but that can be incorporated in a later PR

jainrocks:
> We should call mosquitto_ctrl directly and not use shell scripts, but that can be incorporated in a later PR

sure, will take up in a later PR.

bin/mosquctl_gateway
jainrocks: Thanks, tests worked, discovery events are getting pushed now.

bin/mosquctl_client
```diff
@@ -17,9 +17,23 @@ shift 2
...
+    $MOSQUITTO_CTRL addClientRole $client_user $role_name
+    
+    $MOSQUITTO_CTRL addRoleACL $role_name subscribePattern "$client_id/config" allow
+    $MOSQUITTO_CTRL addRoleACL $role_name subscribePattern "$client_id/commands" allow
+    $MOSQUITTO_CTRL addRoleACL $role_name subscribePattern "$client_id/errors" allow
+    $MOSQUITTO_CTRL addRoleACL $role_name publishClientSend "$client_id/events/#" allow
+    $MOSQUITTO_CTRL addRoleACL $role_name publishClientSend "$client_id/state" allow
+    echo "Device $client_id registered correctly."
+else
+    echo "Device $client_id deleted correctly."
```
grafnu, line 38: Do the ACLs get deleted automatically when the role gets deleted?

grafnu (approved):

PR #1316, opened 2026-05-06
Resolve reflector thread hangs, NPEs, and blank config drops; inject diagnostic metrics

khyatimahendru:


udmis/src/main/java/com/google/bos/udmi/service/core/ProcessorBase.java
```diff
@@ -302,7 +301,9 @@ private String updateConfig(Entry<Long, String> previous, Envelope attributes,
       mungeConfigDebug(attributes, "undefined", (String) extraField);
       return BROKEN_CONFIG_JSON;
     } else if (newLastStart != null) {
-      payload = asMap(ofNullable(previous.getValue()).orElse(EMPTY_JSON));
+      String prevValue = previous.getValue();
```
khyatimahendru: @grafnu I wanted to check if it is intended behavior to catch and throw an exception when `previous` is an empty string? The change I made for handling an empty value is causing an itemized test (`27 TEST system_last_update dupeState`) to fail and I want to confirm if it should be undone.

PR #1317, opened 2026-05-08
Fix sequencer CI failures introduced after Mosquitto ACL enforcement

khyatimahendru:
Sequencer tests were failing when run against `//mqtt/localhost`. Following fixes have been introduced:

- Grant Missing Gateway Attach ACLs
  - **Problem:** Mosquitto rejected gateway attempts to publish to their proxy devices' attach topics, causing MqttException (128) subscription denials.
  - **Solution:** Added the explicit `publishClientSend "$device_id/attach" allow` ACL rule during gateway provisioning.
- Refine Extraneous Subscription Callbacks
  - **Problem:** Sharded proxy-only tests share an underlying gateway connection (GAT-123) that automatically subscribes to the gateway's config updates. Because no handler was registered for the gateway itself, incoming config messages threw fatal runtime exceptions that crashed the Paho MQTT communication thread.
  - **Solution:** Replaced the fatal error logic with a non-fatal warning for unmanaged shared conduits.
- Eliminate Sharding-Induced ACL Desync
  - **Problem:** Because the local etcd database retains cached binding states across CI shards while ephemeral Mosquitto containers frequently lose their dynamic security ACL memory between steps, etcd skipped sending bind operations. This left the broker without the ACLs required to authorize subscriptions.
  - **Solution:** Conditionally forces full proxy binding when executing against local providers (MQTT or IMPLICIT) to guarantee that Mosquitto dynamic security ACLs are fully populated every single time, while preserving the original delta optimization (toBind) for remote cloud registries.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -461,8 +461,8 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 20
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-java@v4
+      - uses: actions/checkout@v6
+      - uses: actions/setup-java@v5
         with:
           distribution: 'temurin'
           java-version: '21'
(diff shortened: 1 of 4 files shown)


grafnu:
For now I'd just revert the breaking change to give us time to properly address and fix!  I don't have enough knowledge to know if these fixes are sane or not.

PR #1325, opened 2026-05-11
Sequencer Tests for Blob Updates

khyatimahendru:
- Add new sequencer tests for blob updates.
- Add utility script `bin/update_blob` to faciliate blob updates to a device.
- Add test script `bin/test_blob_updates` and the corresponding CI test job.
- Add docs for blob updates specification (PREVIEW at https://khyatimahendru.github.io/udmi/docs/specs/blob_updates.html).
- Enable module updates in pubber using a git based module emulator.


docs/specs/categories.md
```diff
@@ -55,10 +55,22 @@ could be used with a **WARNING** level for a warning case during normal operatio
   * _device_: Relating to a specific individual device
     * _apply_: (**INFO**) Stage of applying a device mapping
 * _blobset_: Handling update of device data blobs
-  * _blob_: Conditions specific to an individual blob
+  * _blob_: (**INFO**) Conditions specific to an individual blob
+    * _update_: (**INFO**) Processing a blob update
     * _receive_: (**DEBUG**) About receiving a blob update
     * _fetch_: (**DEBUG**) Fetching a blob update
+      * _oversize_: (**ERROR**) Insufficient storage to download or unpack blob
```
grafnu: I think there might be one too many level of detail here for what a "category" is supposed to be. It wasn't meant to be an exact error code, so much as a description of what general category of error.  E.g. "it had an error fetching the blob" -- but not really all the details.   Trying to enumerate all the potential for the errors might be a bit onerous and therefore fragile, and I'm not sure what value it would add.  What would happen if we lost that last level?

schema/model_testing_update.json
```diff
@@ -0,0 +1,25 @@
+{
+  "$udmi_version": "1.5.2",
+  "title": "Blob Update Testing Model",
+  "type": "object",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "additionalProperties": false,
+  "properties": {
```
grafnu, line 7: hmmm -- I think rather than including all of this inline, we should start to define the "catalog" in the site_model. All this would do then is have the necessary information for the catalog entry to test with. We probably woudln't need to create another model here, rather either just use the existing one.  Just trying to keep the proliferation of new stuff down.

Does that make sense (conceptually)?  Have a think about it and we can find a time to chat soon about it.

A starting point would be to rename this structure something like "catalog_blob_entry" .. the harder part is we'd need to create a new directory strucutre that's not under "devices/" to old the stuff (e.g. "catalog/blobs/blob_name.json") or something.

docs/specs/categories.md
grafnu: I was thinking -- and basically the categories are what various logging messages would use -- so there's basically multiple log messages that would work for any given category, and then the errors are just log messages that have a higher level to them. So, like _fetch_ is a stage that the system goes through and would have multiple log messages and then maybe an error or warnings or something.

I suppose I'm somewhat reactoing to the over-specificaiton of errors and exceptions in Java, where they can get kinda overly specific sometimes but they don't really get used very much.

schema/model_testing_update.json
grafnu: If you want, tho, we could just leave this as is for now since it's working fine, and then when we do the catalog of blobs we can change it to be that structure instead.  Don't really see the need to make you change _this_ stuff now.

docs/specs/categories.md
khyatimahendru: I have simplified by removing the fourth level of categories, PTAL!

schema/model_testing_update.json
khyatimahendru: Yes, that's a great idea! I think it's best we take it up later when there's more clarity on the "catalog of blobs" piece.

grafnu (approved):

PR #1328, opened 2026-05-12
add setting for force password in implicitIoTAccessProvider

noursaidi:
Add a configuration option to force passwords in ImplicitIoTAccessProvider to `unused`, which is used when an external proxy is responsible for authenticating devices, and Mosquitto is still responsible for authorization, but possibly requires a username to make it work.

This is configured by adding force_password=true to the options string of the implicit provider configuration.
e.g. 
```json
"implicit": {
      "provider": "implicit",
      "options": "enabled=${ETCD_CLUSTER},force_password=true"
    }
```

### DIFF
--- a/udmis/src/main/java/com/google/bos/udmi/service/access/ImplicitIotAccessProvider.java
+++ b/udmis/src/main/java/com/google/bos/udmi/service/access/ImplicitIotAccessProvider.java
@@ -67,16 +67,15 @@ import udmi.schema.IotAccess.IotProvider;
  *
  * <p>Supported options:
  * <ul>
- * <li><code>force_password</code>: If true, sets the password for all devices to "unused".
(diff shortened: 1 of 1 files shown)


noursaidi:
https://mosquitto.org/documentation/dynamic-security/

> Password
> The client password maps to the password provided in the CONNECT packet when a device connects. **The password may be unset when a client is created, this will mean that devices will be unable to connect as the corresponding client**.

grafnu (approved):
Would it make sense to add a bit of configurable magic here, so make it something like `force_password=unused` or `use_password=unused` so in some bizarre world maybe there might be use for a different password in different parts of the system?  This is just "design by principle" where in general boolean flags are bad since they offer no room for future tweaks. I'm not sure if it will every make any difference.

noursaidi:
`use_password=unused` is much better! I will update, thanks

PR #1329, opened 2026-05-12
Scale RegistrarService using Redis Distributed Locks and ThreadPool

khyatimahendru:


noursaidi:
No redis please. It adds substantial maintenance/complexity which we are currently not able to support

PR #1332, opened 2026-05-13
Add region property to site and device metadata with automatic propogation

noursaidi:
This change introduces a new `region` field to the UDMI metadata schemas. It is added as a top-level property in `site_metadata.json` and as a property within the `location` object in `model_system.json` (used for device metadata).

The Registrar has been updated to automatically propagate this `region` from the site metadata to any device that does not have its own `region` explicitly set. This ensures consistency across a site while allowing for per-device overrides.

Java classes were regenerated to include the new fields, and unit tests were added to `LocalDeviceTest.java` to verify that:
1. A missing region in device metadata is correctly populated from site metadata.
2. An existing region in device metadata is preserved and not overwritten.

---
*PR created automatically by Jules for task [17728302459280877011](https://jules.google.com/task/17728302459280877011) started by @noursaidi*

### DIFF
--- a/schema/model_system.json
+++ b/schema/model_system.json
@@ -79,6 +79,14 @@
             "style": "bold"
           }
         },
+        "region": {
+          "description": "The region according to the site model in which the device is installed in",
+          "type": "string",
+          "$presentation": {
+            "display": "show",
+            "style": "bold"
+          }
+        },
         "panel": {
           "description": "The reference of the panel where the device is installed in",
           "type": "string",
--- a/gencode/java/udmi/schema/Location.java
+++ b/gencode/java/udmi/schema/Location.java
@@ -14,6 +14,7 @@ import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonPropertyOrder({
     "site",
+    "region",
     "panel",
     "section",
     "room",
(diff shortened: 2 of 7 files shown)


schema/model_system.json
```diff
@@ -79,6 +79,14 @@
             "style": "bold"
           }
         },
+        "region": {
+          "description": "The region according to the site model in which the device is installed in",
```
grafnu, line 83: why is this needed in the device model?  When is a device ever not in the region for the entire site?

grafnu (commented):
I don't understand the reason for this.... ?!?!?

schema/model_system.json
noursaidi: It will shouldn't ever be different, but it needs to be in the device metadata it's describing the device, and we know which region the device is in (the site metadata describes the site). The use cases would be filtering both messages in pubsub where we can add this as an attribute, and also device, where we filter metadata for devices). We can do elaborate data relationship/joins, but it's better and preferable for all the data to be in the device metadata

grafnu: Ok, I think I understand now. My concern is that it ends up in the pre-normalized metadata too, which then makes it redundant info. Can you add a simple check (should just be a line or two) in registrar that generates a warning if it's in the pre-normalized metadata? 

Also, should have a test for this in `tests/sites` somewhre -- doesnt' need to be new, but just add a region into one of the existing sites, so the effects of this are obvious.  (Also then can add a field into a metadata.json file somewhere and see the error).  Unit tests for things like this aren't always that useful since they're way too spot focused, which is why I'm looking for the site-level tests.

grafnu:
Can we close this out or close it down?

noursaidi:
Closing because not required currently, but remains desired.

PR #1333, opened 2026-05-14
Refactor Mosquitto security integration

noursaidi:
This branch replaces the slow, shell-scripted mosquctl commands with a native Java integration that talks directly to Mosquitto's dynamic security plugin using its MQTT v5 JSON API. The primary reason for this change is that Mosquitto writes its entire security configuration to disk synchronously on every update, which slows down the broker during frequent changes. To protect broker performance and prevent disk bottlenecks, the new system uses a thread-safe queue to collect commands and batches them together, enforcing a minimum five-second delay between updates so the broker writes to disk at most once per batch.

Additionally, the branch introduces an optimistic fallback mechanism where failed client registrations automatically convert into non-blocking modifications without stalling the execution thread. The branch also optimizes the ImplicitIotAccessProvider by batching database transactions and enabling parallel device fetching to speed up gateway-device binding. Idempotent commands are protected by ignoring common benign errors, such as trying to add a group that already exists, and the entire flow is validated by a new integration test suite.


bin/test_registrar_integration.sh
grafnu, line 1: remove `.sh` -- it doesn't matter externally what it is internally

schema/configuration_endpoint.json
```diff
@@ -224,6 +224,30 @@
         }
       }
     },
+    "ca_file": {
+      "description": "Full path to the CA certificate file",
```
grafnu: is it really "full path"?  Usually it would just be a path since 90% of the time it's relative... ?

grafnu (approved):

schema/configuration_endpoint.json
noursaidi: No, a normal path would work. Updated.

PR #1336, opened 2026-05-21
fix: Resolve sequencer last_start regression and gateway proxy concurrency crash

khyatimahendru:
- Generalize sequencer last_start regression guard in SequenceBase.java to prevent backwards timestamps updates on out-of-order delivery.
- Use ConcurrentHashMap to shield against ConcurrentModificationException during Jackson serialization.
- Patched CatchingScheduledThreadPoolExecutor to unpack and log exceptions thrown by Scheduled Futures.


grafnu:
Is there a better way to implement the ThreadSafeHashMap using a standard library class instead?

Yes, there absolutely is! In fact, implementing a thread-safe map by extending HashMap and synchronizing every method is a classic "reinventing the wheel" anti-pattern in modern Java.

While your current implementation solves the ConcurrentModificationException issue by returning snapshot copies in entrySet(), keySet(), and values(), it suffers from major performance bottlenecks because it synchronizes entire read and write operations on a single lock.

grafnu (commented):
No need for a complete new class!

khyatimahendru:
@grafnu I originally intended to use the standard `ConcurrentHashMap` but in the auto-generated classes PointsetState and PointsetEvents, the `points` collection is declared strictly as a concrete `HashMap` rather than the `Map` interface:
```
    @JsonProperty("points")
    @JsonPropertyDescription("Collection of point names, defining the representative point set for this device.")
     public HashMap<String, PointPointsetState> points;

    @JsonProperty("points")
    @JsonPropertyDescription("Collection of point names, defining the representative point set for this device.")
    public HashMap<String, PointPointsetEvents> points;
 ```

Since `ConcurrentHashMap` is a subclass of `Map` and not `HashMap`, I had to add a new class.

grafnu (approved):

khyatimahendru:
As discussed, I have changed the object type in the schema itself and removed the custom class

PR #1342, opened 2026-05-29
Modbus Specs and Test

grafnu:


validator/src/main/java/com/google/daq/mqtt/util/providers/ModbusFamilyProvider.java
noursaidi, line 1: Not specifically related to modbus, but is there a way to not hard code the specification into JAVA code? It's been challenging before when bacnet enforcement came into effect and having to search for Java to understand exactly what is required.

If we've standardised on query strings should be possible to have something text based?

noursaidi (approved):

PR #1351, opened 2026-06-05
2026-06 device testing update

pisuke:
2026-06 device testing update


docs/device_testing/gateways_and_direct_devices.html
```diff
@@ -2,7 +2,7 @@
     <!DOCTYPE html>
     <html>
       <head>
-        <title>UDMI test results for gateways and direct devices</title>
+        <title>UDMI test results for wip gateways and direct devices</title>
```
grafnu: Is this change intentional? I don't have a problem with it, just flagging to make sure it's intended on your part.

pisuke: Thanks for spotting it, I'll correct

PR #1354, opened 2026-06-08
Initial UUFI Implementation

grafnu:


khyatimahendru (approved):

misc/discoverynode/testing/e2e/test_local.py
```diff
@@ -278,7 +278,7 @@ def test_discovered_devices_are_created(
       devices=range(0),
       devices_with_localnet_block=range(0),
       discovery_node_id="AHU-1",
-      discovery_node_is_gateway=False,
+      discovery_node_is_gateway=True,
```
khyatimahendru, line 281: FYI @grafnu @noursaidi 
I got gemini to analyze this test and identify why changing this value does not change the test result. It complains of a bug (now fixed) in the test setup and that having a non-gateway discovery node is invalid according to MappingAgent. I will investigate more on this tomorrow.

---

The naming distinction between the two tests:
1. `test_discovered_proxied_devices_are_created`
2. `test_discovered_devices_are_created`

initially suggested they were meant to test different scenarios (one with a gateway and one without). 

However, looking at the UDMI model and the code of the mapping agent, **a non-gateway device performing discovery of other devices is an invalid configuration**. 

Here is why:

### 1. How UDMI Models Discovered Devices
In [MappingAgent.java](file:///usr/local/google/home/heykhyati/Projects/udmi_clone/udmi/validator/src/main/java/com/google/daq/mqtt/mapping/MappingAgent.java#L365-L368), when a discovery event is processed and a new device metadata file is created, the mapper **always** injects the discovering node as the device's gateway:
```java
metadata.gateway = new GatewayModel();
...
metadata.gateway.gateway_id = deviceId; // The discovering node (e.g. AHU-1)
```
In the UDMI schema, this explicitly defines the discovered device as being "proxied" by the discovering node.

### 2. The Model Violation
Because the discovered devices are registered as proxied devices under `AHU-1`, the discovering node `AHU-1` **must** be configured as a gateway. A standard (non-gateway) device cannot have proxied devices associated with it.

If `AHU-1` is configured with `discovery_node_is_gateway=False`:
* It has no `gateway` block in its metadata.
* The mapping agent tries to update the gateway's proxy list (`metadata.gateway.proxy_ids`) and crashes with a `NullPointerException` because `metadata.gateway` is null.

### Conclusion
The old test configuration (`discovery_node_is_gateway=False` for `AHU-1`) was actually a **bug in the test setup itself**. It created an invalid model state that violated UDMI architecture (a non-gateway performing device proxying). 

By changing it to `True`, the PR corrected this setup bug to align with the schema requirements.

grafnu:
So, it's not entirely invalid for a non-gateway (e.g. spotter) to discover
devices... because the discovery results can be fed into something else to
actually proxy for them. I think the Delta systems might be like this...
for whatever reason they can't do discovery, but they can be a gateway.
So, spotter discovers the device, then puts the BACnet info into the model
for the gateway device, and then the gateway proxies the data. There's
nothing *conceptually *wrong with that path, I just don't know what a test
of that would mean w/o a Testing Cadre setup!

PR #1356, opened 2026-06-09
Move python library files inside clientlib/python

khyatimahendru:


khyatimahendru:
@grafnu do you see an issue with generating the python gencode inside clientlib/python and adding a symlink in the gencode directory? I had to move those inside because poetry could not find the schema files otherwise.

grafnu:
> @grafnu do you see an issue with generating the python gencode inside clientlib/python and adding a symlink in the gencode directory? I had to move those inside because poetry could not find the schema files otherwise.

I suppose it doesn't work the other way around? Generate in gencode/ and then link from clientlib/python?  Only reason is it's generally nice to have all the gencode stuff in one place so it's very clear it's not part of the core code, etc...

khyatimahendru:
> > @grafnu do you see an issue with generating the python gencode inside clientlib/python and adding a symlink in the gencode directory? I had to move those inside because poetry could not find the schema files otherwise.
> 
> I suppose it doesn't work the other way around? Generate in gencode/ and then link from clientlib/python? Only reason is it's generally nice to have all the gencode stuff in one place so it's very clear it's not part of the core code, etc...

Yes, that is how it was originally which did not work when I moved the poetry files inside clientlib. I ended up reverting that and added a custom build script instead.

PR #1363, opened 2026-06-15
Enable no-sudo local setups distinguished by port

grafnu:
Makes it possible to run the setup with a project spec like `//mqtt/localhost:82321` with a specific port number so that multiple instances can be run side-by-side. Furthermore, this removes the need to run with `sudu` as a system service. If there is no port number specified then it has the same previous behavior (runs as a system service requiring `sudo`.


bin/site_trigger
```diff
@@ -0,0 +1,127 @@
...
+
+# Build model update payload
+payload = {
+    "projectId": "vibrant",
+    "transactionId": f"TXN-{os.urandom(4).hex()}",
+    "publishTime": "2026-06-12T16:00:00Z",
+    "source": "site_trigger",
+    "payload": {
+        "version": "1.5.2",
+        "timestamp": "2026-06-12T16:00:00Z",
```
jainrocks: Let's use dynamic UTC timestamps here instead of the hardcoded `2026-06-12T16:00:00Z`?

bin/pull_mqtt
```diff
@@ -1,16 +1,22 @@
 #!/bin/bash -e
 
 UDMI_ROOT=$(dirname $0)/..
-cd $UDMI_ROOT
+# cd $UDMI_ROOT
```
jainrocks: if not needed, we can remove this.

bin/pull_mqtt
```diff
@@ -19,11 +25,16 @@ source $UDMI_ROOT/etc/mosquitto_ctrl.sh
 
 topic_filter_udmi="/r/${registry_id}/d/+/#"
 topic_filter_uufi="/uufi/#"
-OUT_BASE=$UDMI_ROOT/out/registries
```
jainrocks, line 22: Why are we removing this?

jainrocks (approved):
LGTM

bin/pull_mqtt
grafnu: Prefixing with $UDMI_ROOT means that it's in the (global) UDMI directory, rather than in a local run directory. The goal is to allow multiple runs using the same install...

bin/pull_mqtt
grafnu: done

bin/site_trigger
grafnu: done

PR #1369, opened 2026-06-16
Implement Mantis: AI triage agent 

khyatimahendru:


tools/mantis/src/app/config/playbook.yaml.bak
grafnu, line 1: are `.bak` files intended or stray?

tools/mantis/src/app/skills/best-effort-triage/SKILL.md
grafnu, line 1: Just curious as to the naming convention here. Is it like this so that it's easy to say "find all SKILL.md" files and then use the dir-name to actually name what they do?  Rather than have a directory of "SKILLs" where each .md file is the skill name?  Would there ever be anything in here that's not a "SKILL.md" file (i.e. any other kind of .md file?)

util/mantis/src/app/agent.py
grafnu, line 1: Seems odd that `skills` is on the same level as the `.py` files?   Right now it's `src/app/config` would it make sense (or be possible?) for it to be `app/config` and `app/skills` and `app/src`?

grafnu (commented):
Just my first pass at trying to understand the directory structure -- I haven't actually looked at any files yet!

tools/mantis/src/app/config/playbook.yaml.bak
khyatimahendru: I removed the .bak file

tools/mantis/src/app/skills/best-effort-triage/SKILL.md
khyatimahendru: The directory-based structure (`<skill-name>/SKILL.md`) is required by the `agentskills_fs` provider we're using. Each directory is a 'skill package' where we can add other skill-specific assets like few-shot examples.json, schema.json, or custom python scripts.

util/mantis/src/app/agent.py
khyatimahendru: Yes, good idea. I will move the skills and config out.

PR #1370, opened 2026-06-23
Skip etcd prerun/postrun checks for cloud targets

khyatimahendru:
Modified both `bin/test_etcd_prerun` and `bin/test_etcd_postrun` to check the TARGET_PROJECT environment variable. If the target project provider is not 'mqtt', the scripts will now exit early and successfully, preventing failures in cloud test runs where etcd is not started.

### DIFF
--- a/bin/test_etcd_postrun
+++ b/bin/test_etcd_postrun
@@ -6,6 +6,17 @@ ROOT=$(dirname $0)/..
 
 source $ROOT/etc/shell_common.sh
 
+# Skip if not a mqtt run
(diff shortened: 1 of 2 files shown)


grafnu:
I'm confused tho, these are only used in the "test_local" setup that should ALWAYS be a local setup... but that doesn't seem to be right either. But then the `mosquitto` test is also run always?  The original intent of that setup was to always run locally to test that stuff even if some other target was provided...  Let's have a quick chat about this in our end meeting today because I think there's some underlying weirdness going on here.

PR #1372, opened 2026-06-25
mqtt to pubsub bridge overhaul

noursaidi:
- use MQTTv5 with support for shared subscriptions
- use a queue for message processing
- add client id command line option, and publish this (along with receieve time) to PubSub
- make missing device num id a non-error
- panic and exit if the queue remains full

### DIFF
--- a/udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
+++ b/udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
@@ -92,6 +92,9 @@ public class MqttToPubSubBridge {
   private final ThreadPoolExecutor executor;
   private volatile boolean tripped = false;
 
+  /**
+   * Initializes a new instance of the bridge, configuring the underlying executor.
+   */
   public MqttToPubSubBridge() {
     this.executor = new ThreadPoolExecutor(
         NUM_THREADS, NUM_THREADS,
--- a/udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
+++ b/udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
@@ -88,7 +88,7 @@ public class MqttToPubSubBridge {
 
   private static final int MAX_QUEUE_SIZE = 99;
   private static final int NUM_THREADS = Runtime.getRuntime().availableProcessors() * 2;
-  
+
   private final ThreadPoolExecutor executor;
   private volatile boolean tripped = false;
 
@@ -101,6 +101,11 @@ public class MqttToPubSubBridge {
         new ThreadPoolExecutor.CallerRunsPolicy());
   }
 
+  /**
+   * Checks if the circuit breaker has been tripped.
+   *
+   * @return true if tripped, false otherwise
+   */
   public boolean isTripped() {
     return tripped;
   }
@@ -177,7 +182,7 @@ public class MqttToPubSubBridge {
       logger.error("gcp_project_id, pubsub_topic_id, and mqtt_client_id are required.");
       System.exit(1);
     }
-    
+
     Long sessionExpiryInterval = 0xFFFFFFFFL;
     if (mqttSessionExpiryInterval != null) {
       sessionExpiryInterval = Long.parseLong(mqttSessionExpiryInterval);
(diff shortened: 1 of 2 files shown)


grafnu (approved):

udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
```diff
@@ -126,32 +174,42 @@ public static void main(String[] args) {
...
+      }
 
-      // Initialize MQTT Client
       mqttClient =
-          new MqttClient(mqttBrokerUrl, MqttClient.generateClientId(), new MemoryPersistence());
-      MqttConnectOptions connOpts = new MqttConnectOptions();
-      connOpts.setCleanSession(true);
+          new MqttClient(
+              mqttBrokerUrl, mqttClientId, new MemoryPersistence());
+      MqttConnectionOptions connOpts = new MqttConnectionOptions();
```
jainrocks, line 206: There can be a risk of leaking sessions on the MQTT broker if the bridge is run without a stable `MQTT_CLIENT_ID` (the default fallback to random UUID).

I think we should use persistent sessions (`CleanStart(false)` and long expiry) when we have a stable, persistent Client ID. If the ID is random, we should clean up immediately. 
Please update the connection options to be conditional:
```java
String mqttClientId = System.getenv("MQTT_CLIENT_ID");
boolean isStableClientId = (mqttClientId != null && !mqttClientId.isEmpty());
if (!isStableClientId) {
  mqttClientId = java.util.UUID.randomUUID().toString();
}
MqttConnectionOptions connOpts = new MqttConnectionOptions();
connOpts.setAutomaticReconnect(true);
connOpts.setReceiveMaximum(100);
if (isStableClientId) {
  connOpts.setCleanStart(false);
  connOpts.setSessionExpiryInterval(0xFFFFFFFFL); 
} else {
  connOpts.setCleanStart(true);
  connOpts.setSessionExpiryInterval(0L); // Expire immediately on disconnect
}

File: udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java [hunk 2]
```diff
@@ -252,101 +311,164 @@ public static void setupBridge(IMqttClient mqttClient, Publisher publisher,
...
-                attributes.put("deviceNumId", numId);
-              }
-
-              if (topicSuffix != null && topicSuffix.startsWith("events/")) {
-                List<String> parts = Splitter.on('/').splitToList(topicSuffix);
-                if (parts.size() >= 2) {
-                  attributes.put("subFolder", parts.get(1));
-                }
-              }
+              final String recieveTime = java.time.Instant.now().toString();
```
jainrocks: Typo: recieveTime, please change to receiveTime.

udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
```diff
@@ -252,101 +311,164 @@ public static void setupBridge(IMqttClient mqttClient, Publisher publisher,
...
+                  // Publish to Pub/Sub with local retries
+                  long backoff = 100;
+                  while (!tripped) {
                     try {
-                      logger.info(
-                          "Published to Pub/Sub with message ID: {}", messageIdFuture.get());
-                    } catch (InterruptedException | ExecutionException e) {
-                      logger.warn("Error publishing to Pub/Sub", e);
+                      ApiFuture<String> messageIdFuture = publisher.publish(pubsubMessage);
+                      String msgId = messageIdFuture.get(); // blocks until complete
```
jainrocks: In the retry loop, if `publisher.publish` succeeds but `mqttClient.messageArrivedComplete` (the MQTT ACK) throws an exception (e.g., due to a transient connection drop), the entire block is retried. This will cause us to publish the message to Pub/Sub again, leading to avoidable duplicates.
Once we have successfully published to Pub/Sub, we should not retry the publish even if the MQTT ACK fails.

Please Separate the Pub/Sub publish from the MQTT ACK in the try-catch block, or track if the publish succeeded so we only retry the ACK.
```java
// Conceptual fix:
String msgId = null;
while (!tripped) {
  try {
    if (msgId == null) {
      ApiFuture<String> messageIdFuture = publisher.publish(pubsubMessage);
      msgId = messageIdFuture.get();
      logger.debug("Published to Pub/Sub with message ID: {}", msgId);
    }
    // Only attempt ACK if publish succeeded
    mqttClient.messageArrivedComplete(message.getId(), message.getQos());
    break; // Success (both publish and ACK complete)
  } catch (Exception e) {
    if (msgId != null) {
      // Pub/Sub succeeded, but MQTT ACK failed. 
      // We should probably break here and rely on broker redelivery, 
      // rather than looping and duplicating in Pub/Sub.
      logger.warn("Pub/Sub published (ID: {}), but MQTT ACK failed. Terminating retry loop to avoid duplicates.", msgId, e);
      break;
    }
    // Handle Pub/Sub failure retry...
    ...
  }
}

udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
noursaidi: We can't expire sessions immediately, because even though the client Id is random it is stable within a run, temporary connections can occur for a variety of reasons.

I have made Client ID required as an option (because we will not use randomly generated client IDs), and set the session expiry to be passed as command line flag.

udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
noursaidi: Thanks, changed

udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
noursaidi: You're correct, thanks. I will update

jainrocks (approved):
LGTM

PR #1383, opened 2026-07-01
Support local workspace builds in bridgehead

khyatimahendru:


bin/start_bridgehead
```diff
@@ -0,0 +1,142 @@
...
+INFLUXDB_TOKEN=$(openssl rand -hex 32)
+INFLUX_USER=bridgehead
+INFLUX_PASSWORD=password
+GRAFANA_USER=bridgehead
+GRAFANA_PASSWORD=password
+EOF
+    echo ".env created successfully."
+fi
+
+# 4. Build Management UI WAR file (udmis/bridgeheadManager.war)
```
jainrocks, line 125: Please add a check if 'javac' and 'jar' are installed before running them, and fail with a helpful message if missing.

.github/workflows/testing.yml
```diff
@@ -556,28 +556,39 @@ jobs:
...
+        run: bin/setup_base
+      - name: Run Bridgehead Test
+        run: |
+          bridgehead/tests/test_bridgehead
+      - name: udmis.log
+        if: ${{ !cancelled() }}
+        run: more out/bridgehead_udmis.log || true
+      - name: validator.log
+        if: ${{ !cancelled() }}
+        run: more out/bridgehead_validator.log || true
```
jainrocks: please use `cat` instead of `more`

jainrocks (approved):
LGTM

PR #1384, opened 2026-07-03
local pubsub emulator and gref testing

jainrocks:


noursaidi (approved):

bin/test_openbridge
```diff
@@ -0,0 +1,631 @@
...
+wait_for_bridge "State Bridge" "$BRIDGE_STATE_PID" "out/bridge_state.log"
+wait_for_bridge "Reflect Bridge" "$BRIDGE_REFLECT_PID" "out/bridge_reflect.log"
+echo "Bridges are running and connected."
+
+echo "Registering devices via All-Pub/Sub Pipeline..."
+bin/registrar "$SITE_MODEL" //gref/udmis
+
+if ! wait_for_device_config "$DEVICE_ID" "$REGISTRY_ID" "$SITE_MODEL"; then
+  echo "Pre-verification failed. Dumping UDMIS logs..." >&2
+  cat out/udmis_host.log >&2
```
grafnu: best to just fail here, and the have the logs output as a separate workflow step. It cleanly separates the logs and doesn't make the test runner output messy.

grafnu (commented):
Is it possible to test openbridge without docker?  My concern is that this is mixing dependencies. openbridge != bridgehead... and pulling in docker makes other kinds of testing harder since it's more opaque. What would be the fundamental reason that docker is needed to test openbridge?   What are the key bits that are being tested here, isn't it just mosquitto + UDMIS with some specific configurations?

bin/test_openbridge
```diff
@@ -0,0 +1,631 @@
...
+  fi
+done
+
+# 5. Site Model Setup
+echo "Setting up site model link..."
+rm -rf bridgehead/udmi_site_model
+ln -s ../sites/udmi_site_model bridgehead/udmi_site_model
+CREATED_SYMLINK=true
+
+# 7. Build Udmis and Validator
```
grafnu: It's 'UDMIS' or "udmis" not "Udmis"

bin/test_openbridge
jainrocks: Done.

bin/test_openbridge
jainrocks: Sure, updated all log messages, comments, and string literals throughout the script to use canonical `"UDMIS"` (or lowercase `"udmis"` where referencing binary directories/packages).

jainrocks:
> Is it possible to test openbridge without docker? My concern is that this is mixing dependencies. openbridge != bridgehead... and pulling in docker makes other kinds of testing harder since it's more opaque. What would be the fundamental reason that docker is needed to test openbridge? What are the key bits that are being tested here, isn't it just mosquitto + UDMIS with some specific configurations?

You are completely right—OpenBridge shouldn't be coupled to Bridgehead or require Docker Compose.

I have entirely removed all Docker, Compose, and Bridgehead dependencies from bin/test_openbridge. The test now runs purely on bare-metal host processes (native Mosquitto with Dynamic Security + TLS, local Etcd, standalone MqttToPubSubBridge JVM instances, and UDMIS host pods).

bin/test_openbridge
```diff
@@ -0,0 +1,560 @@
...
+
+# 8. Start Native Infrastructure Services (Etcd, Mosquitto, Pub/Sub Emulator)
+export AUTH_USER=scrumptious
+export AUTH_PASS=aardvark
+export SERV_USER=rocket
+export SERV_PASS=monkey
+export registry_id="$REGISTRY_ID"
+
+echo "Starting Etcd natively..."
+bin/start_etcd
```
grafnu: use 'bin/start_local' outside the test itself (see other tests)

bin/test_openbridge
```diff
@@ -0,0 +1,560 @@
...
+  fi
+fi
+
+export UDMI_NO_SUDO=true
+export HOST_IP="127.0.0.1"
+export MOSQUITTO_LOG_PATH="$(pwd)/var/mosquitto/log/mosquitto.log"
+
+if [ ! -d "sites/udmi_site_model" ]; then
+  echo "Site model directory missing. Cloning..."
+  bin/clone_model
```
grafnu: see other tests -- all the base setup should be done outside the test itself (i.e. the test itself should be able to run multiple times with the same setup)

bin/test_openbridge
jainrocks: Sure, Done.

bin/test_openbridge
jainrocks: Sure, Done.

grafnu: Not resolved. It applies to all the base setup not just this one thing.

grafnu (commented):
Issue not resolved.

bin/test_openbridge
jainrocks: I have moved this and the remaining one's out of this test.

jainrocks:
> Issue not resolved.

Sure, I have updated `bin/test_openbridge` so that **all base setup is completely externalized**:
1. **Removed internal base setup from `bin/test_openbridge`**: Removed `bin/start_etcd`, `bin/setup_ca`, `bin/start_mosquitto`, and `mosquctl_site` from inside `bin/test_openbridge`.
2. **Explicit Precondition Checks**: `bin/test_openbridge` now checks that base services (`etcd` on 2379 and `mosquitto` on 8883) are active and fails fast with an explicit error (`Please run 'bin/start_local sites/udmi_site_model //mqtt/localhost' first`) if missing.
3. **CI Workflow Alignment**: Updated `.github/workflows/testing.yml` to execute `bin/clone_model` and `bin/start_local` in the setup step prior to calling `bin/test_openbridge`.

grafnu (approved):
Next time figure how to reign in the AI a bit -- it's still generating a lot of crap that's not necessary... we can talk as a group about it, but... it's leaking through...

PR #1385, opened 2026-07-05
Validator checks for bad timestamp and present_value 

jainrocks:


validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
```diff
@@ -690,10 +694,18 @@ private Instant getInstant(Object msgObject, Map<String, String> attributes) {
       Map<String, Object> mapped = mapCast(msgObject);
       String timestamp = (String) mapped.get(TIMESTAMP_KEY);
       if (timestamp != null) {
-        return JsonUtil.getInstant(timestamp);
+        try {
+          return JsonUtil.getInstant(timestamp);
+        } catch (Exception e) {
+          // Ignore and fallback to publish time.
```
grafnu: mmmm -- not sure we should fallback. this is validating the messages and if it doens't have a timestamp I think that means an error?  There are cases where there might be a valid missing timestamp... not sure if this is one of them?

validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
```diff
@@ -690,10 +694,18 @@ private Instant getInstant(Object msgObject, Map<String, String> attributes) {
...
+        }
       }
     }
-    return JsonUtil.getInstant(attributes.get(PUBLISH_TIME_KEY));
+    try {
+      return JsonUtil.getInstant(attributes.get(PUBLISH_TIME_KEY));
```
grafnu: there's a function catchToNull() that can be used to streamline some of this

grafnu (approved):

validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
jainrocks: `getInstant()` is an internal helper for clock sync (mockNow),  message capturing (writeMessageCapture) and rate limiting (shouldProcessMessage) rather than the message validator itself. Payload timestamp validation is enforced separately by validateTimestamp(), which records errors under Category.VALIDATION_DEVICE_CONTENT. Falling back to publishTime here allows processing to continue so validateTimestamp() can record the violation report.

validator/src/main/java/com/google/daq/mqtt/validator/Validator.java
jainrocks: Sure, Done.

grafnu (approved):

PR #1388, opened 2026-07-13
Allow for TLS with etcd,  UDMIS parallelization bug fixes

noursaidi:
- enable tls with client auth for comms to etcd
- make config updates atomic
- use correlation in mosquitto dynamic security comms to remove crosstalk

### DIFF
--- a/bridgehead/tests/test_bridgehead
+++ b/bridgehead/tests/test_bridgehead
@@ -1,5 +1,5 @@
 #!/bin/bash -e
-
+set -x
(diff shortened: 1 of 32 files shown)


bridgehead/tests/test_bridgehead
```diff
@@ -1,111 +1,103 @@
 #!/bin/bash -e
-
+set -x
```
grafnu: maybe?  If real move to first line with -e

grafnu (approved):

PR #1392, opened 2026-07-16
Allow specific targetting of iot provider in registrar

noursaidi:
When UDMIS is configured with multiple providers, allow registrar to target provisioning commands to a specific provider, e.g.

`bin/registrar sites/udmi_site_model //gref/project -P implicit` (implicit meaning the mosquitto backend)

Use case is in a multi-provider setup, there is a chicken and egg problem because UDMIS would use the _registry_ affinity which is based on where devices were already connected to, but they cannot connect to the new thing without being provisioned there.

Historically, this was not a problem because reflector always connected to the targetted endpoint.

Alternatives could include pinning (e.g. gref always goes to an `implicit` iot access provider backend)


bin/test_registrar_dynamic
```diff
@@ -0,0 +1,71 @@
...
+bin/start_local ${TEST_SITE} $PROJECT_SPEC
+
+# Stop udmis (but keep mosquitto)
+OLD_PID=$(cat var/udmis.pid)
+sudo kill $OLD_PID || true
+sleep 2
+
+# Start UDMIS with our custom pod config
+export UDMIS_CONFIG=var/dynamic_pod.json
+bin/start_udmis
```
grafnu: `start_udmis` is already in `start_local` -- does this need to be done again?  Why can't you set the env variable when `start_local` is first called? Is there any _fundamental_ reason this can't work without the ugly kill-and-restart cycle?

bin/test_registrar_dynamic
```diff
@@ -0,0 +1,71 @@
...
+    },
+    "temporary": {
+      "provider": "temporary"
+    }
+  }
+}
+EOF
+
+# Start standard local setup (it will create var/local_pod.json and start UDMIS with it)
+bin/start_local ${TEST_SITE} $PROJECT_SPEC
```
grafnu: generally I've been trying to keep the setup (`start_local`) outside of the testing script itself... e.g. it would be better to be able to run `test_registrar_dynamic` multiple times with only one setup rather than having to do everything over and over again... does it work to move this to the testing wrapper?  Same with `clone_model` and all that stuff.

bin/test_registrar_dynamic
noursaidi: No - I didn't pay attention to what was already in the github CI test. Updated

bin/test_registrar_dynamic
noursaidi: Done

bin/start_udmis
```diff
@@ -80,28 +80,40 @@ else
...
-        .flow_defaults.port = '"${MQTT_PORT:-8883}"' |
-        .iot_access.implicit.project_id = "'"${ETCD_SCHEME}"'://${ETCD_CLUSTER}:'"${ETCD_PORT}"'" |
-        .iot_access.implicit.endpoint.port = '"${MQTT_PORT:-8883}"' |
-        .iot_access.implicit.endpoint.ca_file = "'"$ABS_CERT_DIR/ca.crt"'" |
-        .iot_access.implicit.endpoint.cert_file = "'"$ABS_CERT_DIR/rsa_private.crt"'" |
-        .iot_access.implicit.endpoint.key_file = "'"$ABS_CERT_DIR/rsa_private.pem"'" |
-        .iot_data.database.project_id = "'"${ETCD_SCHEME}"'://${ETCD_CLUSTER}:'"${ETCD_PORT}"'"' \
-        $SOURCE_POD_CONFIG | sponge var/local_pod.json
+mkdir -p var
+if [[ -f /root/var/local_pod.json ]]; then
```
grafnu: This seems backwards. If UDMIS_DIR is defined then it should use that (or error if it's not valid).  Note that it should also be picking up files from $UDMIS_DIR/var/local_pod.json -- not really directly using the files in etc/ -- they are movied to var/ when the thing is built.

bin/start_udmis
```diff
@@ -80,28 +80,40 @@ else
...
+    .iot_access.implicit.endpoint.port = '"${MQTT_PORT:-8883}"' |
+    .iot_access.implicit.endpoint.ca_file = "'"$ABS_CERT_DIR/ca.crt"'" |
+    .iot_access.implicit.endpoint.cert_file = "'"$ABS_CERT_DIR/rsa_private.crt"'" |
+    .iot_access.implicit.endpoint.key_file = "'"$ABS_CERT_DIR/rsa_private.pem"'" |
+    .iot_data.database.project_id = "'"${ETCD_SCHEME}"'://${ETCD_CLUSTER}:'"${ETCD_PORT}"'"' \
+    $SOURCE_POD_CONFIG | sponge var/local_pod.json
+
+if [[ -f /root/var/dynamic_pod.json ]]; then
+    SOURCE_DYNAMIC_CONFIG=/root/var/dynamic_pod.json
+else
```
grafnu: same issue here

bin/start_udmis
```diff
@@ -80,28 +80,40 @@ else
...
+    SOURCE_DYNAMIC_CONFIG=$UDMIS_DIR/etc/dynamic_pod.json
+fi
+if [[ -f $SOURCE_DYNAMIC_CONFIG ]]; then
+    jq '.include = "local_pod.json"' $SOURCE_DYNAMIC_CONFIG | sponge var/dynamic_pod.json
+fi
+
+if [[ ${UDMIS_CONFIG:-} == dynamic || ${UDMIS_CONFIG:-} == *dynamic_pod.json ]]; then
```
grafnu: this seems too hacky. If UDMIS_CONFIG is defined then it should use it, otherwise not -- why special case "dynamic"?

bin/start_udmis
noursaidi: Agree

bin/start_udmis
noursaidi: No good reason - agree

grafnu (approved):

udmis/src/main/java/com/google/bos/udmi/service/access/TemporaryIotAccessProvider.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  @Override
+  public CloudModel listDevices(String registryId, Consumer<String> progress) {
+    CloudModel cloudModel = new CloudModel();
+    cloudModel.device_ids = new ConcurrentHashMap<>();
+    Map<String, CloudModel> registryStore =
+        store.getOrDefault(registryId, new ConcurrentHashMap<>());
+    for (String deviceId : registryStore.keySet()) {
+      cloudModel.device_ids.put(deviceId, registryStore.get(deviceId));
```
jainrocks, line 42: `registryStore.get(deviceId)` can return `null` if a device is deleted by another thread between `keySet()` snapshot and `get()`. Because `cloudModel.device_ids` is initialized as a `ConcurrentHashMap`, passing `null` as a value to `.put(key, null)` will throw a `NullPointerException`.
Please iterating over `entrySet()` and adding a null check:
```java
for (Map.Entry<String, CloudModel> entry : registryStore.entrySet()) {
  if (entry.getValue() != null) {
    cloudModel.device_ids.put(entry.getKey(), entry.getValue());
    if (progress != null) {
      progress.accept(entry.getKey());
    }
  }
}

PR #1396, opened 2026-07-21
fix createDevice missing initial config in implicit iot access provider

noursaidi:
The addition of compare and swap into config updates introuced this error which was only caught when ETCD is wiped and registrar run.

TODO: Update tests because I would have expected the CI test (which starts with a fresh database) to have caught this if that was the real cause.


noursaidi:
Closing because as convincing as this sounds it was not the cause of the issue - it was a registry affinity thing in PR #1392

PR #1402, opened 2026-07-25
hide state only devices from etcd explorer

noursaidi:
If a device publishes state but is not registered in etcd/mosquitto (because UDMIS is in a multiple backend perspective, and current behaviour is to persist all state messages into ETCD), then they are hidden from the explorer interface

### DIFF
--- a/udmis/src/main/java/com/google/bos/udmi/service/support/EtcdExplorerServer.java
+++ b/udmis/src/main/java/com/google/bos/udmi/service/support/EtcdExplorerServer.java
@@ -79,6 +79,7 @@ public class EtcdExplorerServer {
     return Integer.compare(s1.length(), s2.length());
   };
 
+  private static final String LAST_STATE_KEY_SUFFIX = ":last_state";
   private static final Pattern DEVICES_PATTERN =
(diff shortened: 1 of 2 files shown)


udmis/src/main/java/com/google/bos/udmi/service/support/EtcdExplorerServer.java
```diff
@@ -147,7 +148,7 @@ private void handleGetRegistries(HttpExchange exchange) throws IOException {
     Set<String> uniqueRegistries = new TreeSet<>(NATURAL_COMPARATOR);
     Set<String> uniqueDevices = new TreeSet<>(NATURAL_COMPARATOR);
     for (String key : keys) {
-      if (key.startsWith("/r/")) {
+      if (key.startsWith("/r/") && !key.endsWith(LAST_STATE_KEY_SUFFIX)) {
```
grafnu, line 151: NB: Ideally extract to function since it's the same logic in both places.

grafnu (approved):

jainrocks (approved):

PR #1403, opened 2026-07-26
unset unused attributes in mqtt -> pubsub bridge

noursaidi:


udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
```diff
@@ -489,6 +489,10 @@ public void messageArrived(String topic, MqttMessage message) {
                     }
                   }
 
+                  // Ensure there is no gatewayId pub/sub attribute, and do not set subType at all
```
grafnu, line 492: er, so why is this necessary? It's a new map and these keys were never added... why do they need to be removed?

grafnu (commented):
This doesn't seem right at all. You're removing keys from a new map... either I'm missing something or this really isn't what you intended or need to do. If there is a reason then this is definitely a time to add a comment that explains why.

udmis/src/test/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridgeTest.java
```diff
@@ -419,5 +423,71 @@ void testGetEtcdOptionsPartialSsl() throws Exception {
...
+    org.junit.jupiter.api.Assertions.assertFalse(attributes.containsKey("gatewayId"),
+        "pub/sub attributes should not contain gatewayId");
+    org.junit.jupiter.api.Assertions.assertFalse(attributes.containsKey("subType"),
+        "pub/sub attributes should not contain subType");
+  }
+
+  @Test
+  void testNoGatewayIdOrSubTypeForStateAndReflectTopics() throws Exception {
+    IMqttClient mockMqttClient = mock(IMqttClient.class);
+    when(mockMqttClient.getClientId()).thenReturn("test-client");
```
jainrocks, line 463: These two new test methods duplicate ~30+ lines of setup code already present in `testSetupBridge`. 
Also note that because `attributes` never contained `gatewayId` or `subType` in `MqttToPubSubBridge`, these tests pass even without the `remove(...)` calls in code.
Recommendation: We can safely remove these duplicate methods and rely on the assertions added to `testSetupBridge` and `testSetupBridgeWithSubFolder` to guard against future regressions.

noursaidi:
Closed because this PR is meaningless

PR #1404, opened 2026-07-26
add no broker auth option for implicitiotaccess provider

noursaidi:


udmis/src/test/java/com/google/bos/udmi/service/support/MosquittoBrokerTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+class MosquittoBrokerTest {
+
+  @Test
+  void testBrokerAuthFalseNoDynamicSecurityRequests() {
+    EndpointConfiguration endpoint = new EndpointConfiguration();
+    endpoint.hostname = "invalid.localhost.nonexistent";
+    endpoint.port = 1883;
+
+    // With brokerAuth=false, no connection to invalid broker should be attempted
```
grafnu, line 16: What does auth have to do with the connection attempt?

noursaidi: If broker auth (bad name in hindsight - basically updating the Mosquitto ACL list) is disabled, then it shouldn't even connect onto the Mosquitto contorl plane

grafnu:
Can you add a comment describing this? I understand now. I'm also realizing
that our new robotic overlords (Gemini) actually might find the comments
useful.
It's not auth to the broker, it's that the broker is doing auth on the
connections, right? Both forms of "broker auth" but would need more words
to disambiguate!

noursaidi:
Comment added

udmis/src/main/java/com/google/bos/udmi/service/access/DynamicIotAccessProvider.java
```diff
@@ -58,14 +58,18 @@ private String determineProvider(String registryId) {
...
+      if (!(provider instanceof PubSubIotAccessProvider)) {
+        debug("Registry affinity mapping for %s inherited from reflector %s: %s",
+            registryId, reflectorKey, reflectorAffinity);
+        return reflectorAffinity;
+      }
     }
     getProviders();
     TreeMap<String, String> sortedMap = getProviders().entrySet().stream()
         .filter(access -> access.getValue().isEnabled())
+        .filter(access -> !(access.getValue() instanceof PubSubIotAccessProvider))
```
grafnu: oof, what's going on here?  This kind of expliclit stuff is generaly bad. Ad the very very very least extract it to a function so the horrible hard-coding is sequestered, and maybe add a comment or TODO about making this more generalized in the guture.

noursaidi: I've re-factored the hardcoded pubsub (which also occurred in three other places in this file which had snuck in previously) into what I think is a correct solution:

https://github.com/faucetsdn/udmi/pull/1404/commits/ad17ce01a7f9cfe17f3c8e78001a450f2dc2f23f

udmis/src/main/java/com/google/bos/udmi/service/access/ImplicitIotAccessProvider.java
```diff
@@ -154,6 +158,18 @@ public ImplicitIotAccessProvider(IotAccess iotAccess) {
 
     enabled = isNullOrNotEmpty(options.get(ENABLED_KEY));
     usePassword = options.get(USE_PASSWORD_KEY);
+    // The BrokerAuth option defines whether the broker is intended to enforce ACLs,
+    // When false, the device registeration process will not register ACL's into Mosquitto
```
jainrocks: typo registeration, please change to `registration`

udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
```diff
@@ -413,13 +418,15 @@ public void setupBridge(IMqttClient mqttClient, Publisher publisher,
...
-                logger.error("Failed to re-subscribe to topic {} after auto-reconnect",
-                    mqttSubscriptionTopic, e);
-              }
+              executor.submit(() -> {
+                try {
+                  mqttClient.subscribe(actualSubscriptionTopic, 1);
+                  logger.debug("Successfully re-subscribed to topic: {}", actualSubscriptionTopic);
+                } catch (MqttException e) {
+                  logger.error("Failed to re-subscribe to topic {} after auto-reconnect",
+                      mqttSubscriptionTopic, e);
```
jainrocks: Please change this `actualSubscriptionTopic` in the looger.error for the consistency.

udmis/src/main/java/com/google/bos/udmi/service/access/DynamicIotAccessProvider.java
```diff
@@ -92,9 +92,8 @@ private IotAccessProvider getRegistryProvider(Envelope envelope) {
 
   private IotAccessProvider getRegistryProvider(String registryId, String deviceId) {
     IotAccessProvider provider = getProviderFor(registryId, deviceId);
-    if (provider instanceof PubSubIotAccessProvider) {
-      IotAccessProvider fallback = getProviders().values().stream()
-          .filter(p -> !(p instanceof PubSubIotAccessProvider))
+    if (!getOperatingProviders().containsValue(provider)) {
```
jainrocks: **nit (non-blocking)**:

The `supportsRegistryOperations()` extraction looks great!

In `getRegistryProvider`:
```java
IotAccessProvider provider = getProviderFor(registryId, deviceId);
if (!getOperatingProviders().containsValue(provider)) {
  ...
```

Since `provider` is already retrieved from `getProviders()`, checking `if (provider == null || !provider.supportsRegistryOperations())` directly is a bit more better and avoids allocating a new `Map` and performing a `containsValue()` scan on each invocation:

```java
private IotAccessProvider getRegistryProvider(String registryId, String deviceId) {
  IotAccessProvider provider = getProviderFor(registryId, deviceId);
  if (provider == null || !provider.supportsRegistryOperations()) {
    IotAccessProvider fallback = getOperatingProviders().values().stream()
        .findFirst()
        .orElse(null);
    if (fallback != null) {
      return fallback;
    }
  }
  return provider;
}
```

jainrocks (approved):
LGTM - few comments

grafnu (approved):

PR #1406, opened 2026-07-27
Add message capture integration with postgres database

grafnu:


bin/pull_capture
```diff
@@ -0,0 +1,205 @@
...
+
+
+def escape_sql_string(val):
+  """Safely escape strings for raw SQL queries."""
+  if val is None:
+    return "NULL"
+  return "'" + str(val).replace("'", "''") + "'"
+
+
+def save_to_postgres(envelope, payload):
```
noursaidi: It should use prepared statements and not a home rolled escaping library.

bin/pull_capture
```diff
@@ -0,0 +1,205 @@
...
+  """
+  cmd = [
+      "psql",
+      "-h", "127.0.0.1",
+      "-p", POSTGRES_PORT,
+      "-U", POSTGRES_USER,
+      "-d", POSTGRES_DB,
+      "-c", sql
+  ]
+  subprocess.run(
```
noursaidi: And if they fail?

bin/pull_capture
```diff
@@ -0,0 +1,205 @@
...
+  INSERT INTO udmi_messages (
+      project_id, registry_id, device_id, sub_folder, sub_type,
+      publish_time, payload
+  )
+  VALUES (
+      {project_id}, {registry_id}, {device_id}, {sub_folder}, {sub_type},
+      {publish_time}, {payload_str}
+  );
+  """
+  cmd = [
```
noursaidi: Preferable to use a python library , but main concern is that initiating a concern per insert will not scale

noursaidi (approved):

schema/envelope.json
```diff
@@ -48,6 +48,9 @@
     "rawFolder": {
       "type": "string"
     },
+    "msgSource": {
```
jainrocks: What is the difference between `msgSource` and `source`? Having both might create confusion.

schema/envelope.json
```diff
@@ -3,7 +3,7 @@
   "title": "Envelope",
   "description": "The UDMI `envelope` is not a message itself, per se, but the attributes and other information that is delivered along with a message. [Message Envelope Documentation](../docs/messages/envelope.md)",
   "$section": "Messages",
-  "additionalProperties": true,
+  "additionalProperties": false,
```
jainrocks: Why are we changing this?

bin/pull_capture
grafnu: done

bin/pull_capture
grafnu: done. right now this is just the preliminary script that I need for discovery -- the scaling bit can be worked out later (optimizations)

schema/envelope.json
grafnu: This was an unrelated change, so I took it out. Will do in another PR as needed.

schema/envelope.json
grafnu: Same, part of a different change.

bin/pull_capture
grafnu: converted to library -- this is just an initial pass at the capture scripts for debugging and diagnostics, so not super worried about 100% reliability for now

PR #1407, opened 2026-07-28
Add Spotter

khyatimahendru:


edge/spotter/src/agent.py
```diff
@@ -0,0 +1,221 @@
...
+            
+            b64_data = base64.b64encode(chunk_data).decode()
+            
+            chunk_event = PcapChunkEvent(
+                session_id=session_id,
+                chunk_index=idx,
+                total_chunks=total_chunks,
+                data=b64_data
+            )
+            device.dispatcher.publish_event("events/pcap", chunk_event)
```
khyatimahendru: @grafnu for now I got spotter to publish to `events/pcap`. Do you think we can reuse any of the existing - events/system, events/pointset, events/discovery, events/alarmset, events/mapping, and events/validation? Or should we add a new event construct similar to blobs?

grafnu:
It would either be something new (e.g.) stream, or blobset.  I think it
kinda makse sense for blob-set since then it's the way that blobs are moved
around. So what would happen in the packet capture case, it would create a
new blobset entry and that would be used for sending the data.  How is this
currently controlled?

PR #1410, opened 2026-07-30
Locking in UDMI version 1.5.5 in the tools

grafnu:


khyatimahendru:
These changes look good to me but there is a reference to `VERSION_1_5_2` in bambi (`mergeDeviceMetadata()` in `LocalSiteModelManager.java`) which I believe should be upgraded too. Is there a variable for the latest version we can use there instead?

khyatimahendru (approved):

noursaidi (approved):

grafnu:
> These changes look good to me but there is a reference to `VERSION_1_5_2` in bambi (`mergeDeviceMetadata()` in `LocalSiteModelManager.java`) which I believe should be upgraded too. Is there a variable for the latest version we can use there instead?

Updated with an updated BAMBI version.

PR #1412, opened 2026-07-31
Handle scan_addr validation errors without aborting device config generation

khyatimahendru:
- Catch validation exceptions in ConfigManager for `scan_addr` and target references and record them into schemaViolationsMap.
- Ensure `generated_config.json` is still created while reporting schema violations.
- Add `scanAddrMismatchHandlingTest` unit test in `RegistrarTest`.

### DIFF
--- a/validator/src/main/java/com/google/daq/mqtt/util/ConfigManager.java
+++ b/validator/src/main/java/com/google/daq/mqtt/util/ConfigManager.java
@@ -176,7 +176,7 @@ public class ConfigManager {
       String family = target.family;
       String localAddr = ifNotNullGet(family, this::getLocalnetAddr);
       String gatewayAddr = target.addr;
-      catchIfSchemaViolation(
+      captureSchemaViolation(
               format("%s gateway target", family),
           () -> checkState(localAddr == null || gatewayAddr == null,
               format("both gateway.target.addr and localnet.families.%s.addr should not be defined",
@@ -194,7 +194,7 @@ public class ConfigManager {
     throw new RuntimeException("Unknown protocol family: " + family);
   }
 
-  private void catchIfSchemaViolation(String description, Runnable action) {
+  private void captureSchemaViolation(String description, Runnable action) {
     try {
       action.run();
     } catch (Exception e) {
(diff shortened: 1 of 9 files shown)


validator/src/main/java/com/google/daq/mqtt/util/ConfigManager.java
```diff
@@ -175,23 +176,43 @@ private GatewayConfig getGatewayConfig() {
...
   }
 
+  private FamilyProvider getFamilyProvider(String family) {
+    if (NAMED_FAMILIES.containsKey(family)) {
+      return NAMED_FAMILIES.get(family);
+    }
+    throw new RuntimeException("Unknown protocol family: " + family);
+  }
+
+  private void catchIfSchemaViolation(String description, Runnable action) {
```
grafnu: maybe name it something more about what it results in doing, i.e. its not really just catching the violation, it's adding it to the map. so, maybe "captureSchemaViolation"?

validator/src/main/java/com/google/daq/mqtt/util/ConfigManager.java
```diff
@@ -270,23 +293,18 @@ private String pointConfigRef(PointPointsetModel model) {
...
-        if (NAMED_FAMILIES.containsKey(targetFamily)) {
-          NAMED_FAMILIES.get(targetFamily).validateUrl(fullRef);
-        } else {
-          throw new RuntimeException("Unknown protocol family in URL: " + targetFamily);
-        }
-      } catch (Exception e) {
-        schemaViolationsMap.put(String.format("%s %s: %s", family, pointRef, getCurrentContext()),
-            wrapExceptionWithContext(e, false));
-      }
+      catchIfSchemaViolation(
```
grafnu: same comment

grafnu (approved):

validator/src/main/java/com/google/daq/mqtt/util/ConfigManager.java
khyatimahendru: Thank you, I have renamed the method

PR #1413, opened 2026-07-31
Changing bacnet to / from :

grafnu:


noursaidi (approved):

noursaidi:
Just a caution with this one, although the message downgrader has test cases to show it work, the legacy format is extensively used, so I'd be wondering if tools such as BAMBI could accidentally overwrite the the version field generating new metadata resulting in broken configs, etc.

grafnu:
I don't think it actually makes it into configs anywhere, I don't think
anybody is implementing the bacnet protocol, specifically?  Not sure.  But,
I have no idea how to test it, as such. Everything will be in source
control so it won't be totally lost/broken. What we'd want to do, I think,
it just have a managed roll-out process so that we can tell when something
breaks and limit the blast radius.

PR #1414, opened 2026-07-31
Fix PROXIED resource_type persistence in ImplicitIoTAccessProvider

jainrocks:


udmis/src/main/java/com/google/bos/udmi/service/access/ImplicitIotAccessProvider.java
```diff
@@ -650,6 +657,11 @@ public CloudModel fetchDevice(String registryId, String deviceId) {
       cloudModel.gateway = new GatewayModel();
       cloudModel.gateway.proxy_ids =
           listBoundDevices(registryId, deviceId).keySet().stream().toList();
+    } else if (properties.get(BOUND_TO_KEY) != null) {
+      cloudModel.resource_type = PROXIED;
+      if (!PROXIED.toString().equals(properties.get(RESOURCE_TYPE_PROPERTY))) {
```
grafnu, line 662: Is this stuff basically a hack to clean up registries that have the bad value?  If so, it's likely the easiest way to do it but not the cleanest since it's going to leave hacky code lying around that ends up being more confusing in the long run. Ideally there'd be some out-of-band cleanup of things, but that's also likely a lot of work. If you want to keep the auto-healing part, please add a very clear `TODO` for all the spots that are auto-healing related and then also create a bug to go back and clean them up after things have settled?  This shouldn't be functionality we need in the system long term once things are set to the correct device type at the get-go!

noursaidi: Why is a computation necesary here and not just always `put` the correct value for all properties at every registrar run? In this case `resource type` is/should just be (if it's not already) some property in the normalized Metadata calculated by  Registrar?

I'm also questioning why this particular value needs to even be in the database, but I'll be starting another thread for that discussion.

jainrocks: Thanks @grafnu and @noursaidi for the feedback. 
1. **Eliminated Read Side-Effects**: Agreed on avoiding database mutations inside reads. I've updated `fetchDevice`/`fetchDevicePartial` to purely project `PROXIED`  without performing any `put` to ETCD. This removes the auto-healing writes and leaves no migration debt or cleanup TODOs.
2. **Why Backend Handling is Needed**: Not all devices come from Registrar runs. Devices auto-discovered in the field by gateways are created and bound directly by UDMIS backend in real time, so UDMIS must manage the `PROXIED` status.

grafnu (approved):

PR #1417, opened 2026-08-04
Make private key file optional for devices

khyatimahendru:
- Update `LocalDevice.java` to move private key files into `OPTIONAL_FILES` and remove them from required expected keyFiles.
- Record a `ValidationWarning` when private key files are missing in device folders.


validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
```diff
@@ -380,6 +383,15 @@ public void validateExpectedFiles() {
       }
     }
 
+    if (isGateway() || isDirect()) {
```
grafnu, line 403: I can't remember specifically, but for even for Gateway and Direct devices the private key isn't strictly required. I thought . Jetski says:

Analyzing Key Source
  Yes, there is exactly a setting for this!

  In the device's metadata.json, under the cloud block, there is a boolean field called device_key.

  When you set "device_key": true, it tells the UDMI tooling (like registrar and validator) that the device itself is the secure source of the private key
  (e.g., it generates its own key or was pre-provisioned).

  As a result, the tools operate in a public-key-only mode for that device: they will no longer expect or look for a private key file (like
  rsa_private.pkcs8) in the site model repository, but they will still require the public key (e.g., rsa_public.pem) to be present so it can be registered
  with the cloud endpoint.

noursaidi:
Can we change the warning into a "NOTICE" or even flip it and warn if private keys ARE present in the directory? Context is that we will imminently be banning and purging private keys from site models, but this can be a change for another day.

validator/src/main/java/com/google/daq/mqtt/registrar/LocalDevice.java
khyatimahendru: Thank you, I found this. It is configured in the device metadata. If `isDeviceKeySource()` evaluates to true, `getPrivateKeyFiles()` simply returns an empty set. Now the logic has been updated to not evaluate this function at all when checking for expected files.

khyatimahendru:
> Can we change the warning into a "NOTICE" or even flip it and warn if private keys ARE present in the directory? Context is that we will imminently be banning and purging private keys from site models, but this can be a change for another day.

@noursaidi I've updated the logs, could you please review again?

grafnu (approved):

PR #1418, opened 2026-08-04
fix mqtt->pubsub bridge failure conditions

noursaidi:
The logic arround unacked messages/termination is flawed, this fixes those assumptions whilst keeping with the strategy:
1 - Circuit breaker considers "abandoned/unacked" messages as well as oldest message being "process" 
2 - add a reconnect so unacked messages can get redelivered
3 - in thread retries of Pub/Sub message delivery
4 - graceful termination

### DIFF
--- a/udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
+++ b/udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
@@ -102,6 +102,9 @@ public class MqttToPubSubBridge {
       .maximumSize(10000)
       .build();
 
+  /**
+   * Represents the state of an unacknowledged MQTT message.
+   */
   public enum UnackedState {
     IN_PROCESS,
     ABANDONED
@@ -161,7 +164,8 @@ public class MqttToPubSubBridge {
               fullStartTime = System.currentTimeMillis();
             } else if (System.currentTimeMillis() - fullStartTime >= 300000) {
               tripped = true;
-              logger.error("Unacked message threshold exceeded for 300 seconds! Tripping circuit breaker.");
+              logger.error(
+                  "Unacked message threshold exceeded for 300 seconds! Tripping circuit breaker.");
               try {
                 mqttClient.disconnect();
               } catch (Exception e) {
@@ -190,10 +194,13 @@ public class MqttToPubSubBridge {
         try {
           // Jitter: +/- 50% random delay around baseIntervalSec
           int jitterSec = (int) (baseIntervalSec * 0.5);
-          int delaySec = baseIntervalSec + (jitterSec > 0 ? (random.nextInt(jitterSec * 2 + 1) - jitterSec) : 0);
+          int delaySec = baseIntervalSec
+              + (jitterSec > 0 ? (random.nextInt(jitterSec * 2 + 1) - jitterSec) : 0);
           Thread.sleep(Math.max(1, delaySec) * 1000L);
           if (mqttClient.isConnected() && !tripped) {
-            logger.info("Executing scheduled reconnect (interval: {}s) to rebalance shared subscription...", delaySec);
+            logger.info(
+                "Executing scheduled reconnect (interval: {}s) to rebalance shared subscription...",
+                delaySec);
             try {
               mqttClient.disconnect();
             } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


grafnu (approved):

udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
```diff
@@ -112,7 +208,24 @@ public MqttToPubSubBridge() {
...
         new ArrayBlockingQueue<>(MAX_QUEUE_SIZE),
         new ThreadFactoryBuilder().setNameFormat("mqtt-bridge-%d").setDaemon(true).build(),
-        new ThreadPoolExecutor.AbortPolicy());
+        (runnable, exec) -> {
+          if (exec.isShutdown()) {
+            throw new RejectedExecutionException("Executor is shut down");
+          }
+          try {
+            // Block up to 2 seconds to absorb transient bursts and apply TCP backpressure
+            boolean added = exec.getQueue().offer(runnable, 2, TimeUnit.SECONDS);
```
jainrocks, line 217: Waiting for 2 seconds blocks Paho's internal network thread when the task queue is full. This prevents from reading network heartbeats (`PINGRESP`), causing unexpected MQTT disconnects. Can we please make this non-blocking?
```
(runnable, exec) -> {
  if (exec.isShutdown() || !exec.getQueue().offer(runnable)) {
    throw new RejectedExecutionException("Executor queue full, marking as ABANDONED");
  }
}
```

udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
```diff
@@ -549,6 +806,75 @@ public void authPacketArrived(int reasonCode, MqttProperties properties) {}
...
+            topic, mqttClient, attempt + 1);
+      }, backoffMs, TimeUnit.MILLISECONDS);
+    } else {
+      logger.error(
+          "Failed to publish to Pub/Sub after 5 attempts."
+              + " Leaving message ID {} unacked (marked as ABANDONED).",
+          mqttMessage.getId(), t);
+      unackedMessages.compute(messageKey, (key, existingRecord) -> {
+        if (existingRecord != null) {
+          existingRecord.setState(MessageState.ABANDONED);
```
jainrocks, line 870: When a message fails all 5 retries, it is marked as ABANDONED in the unackedMessages map. If the pod does not trip immediately, these failed messages are never removed from memory, causing the map to grow continuously over time.
Let's add an expiration TTL or use a cache so that old ABANDONED entries are automatically cleaned up after their timeout of say 10 minutes?

udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java
```diff
@@ -708,4 +1034,93 @@ private static String getDeviceNumId(EtcdDataProvider etcdProvider, String regis
...
+        logger.debug("Pub/Sub Publisher shut down.");
+      } catch (Exception e) {
+        logger.warn("Error shutting down Pub/Sub publisher", e);
+      }
+    }
+
+    // 4. Disconnect MQTT client after in-flight message ACKs have completed
+    if (mqttClient != null && mqttClient.isConnected()) {
+      try {
+        mqttClient.disconnect();
```
jainrocks, line 1073: `mqttClient.disconnect()` without arguments can block for up to 30 seconds if the network is sluggish, which may exceed Kubernetes container termination deadlines.
Please pass an explicit 1-second timeout:
``` mqttClient.disconnect(1000L);```

File: udmis/src/main/java/com/google/bos/udmi/service/bridge/MqttToPubSubBridge.java [hunk 4]
```diff
@@ -184,6 +376,15 @@ public static void main(String[] args) {
     String mqttKeepAliveInterval = commandLine.getOptionValue("mqtt_keep_alive_interval", "15");
     int keepAliveInterval = Integer.parseInt(mqttKeepAliveInterval);
     boolean mqttTls = commandLine.hasOption("mqtt_tls");
+    int reconnectIntervalSec = Integer.parseInt(
```
jainrocks, line 379: Please document the newly added CLI parameters in the README.md:

```
--mqtt_clean_start
--mqtt_reconnect_interval_sec
--circuit_breaker_unacked_threshold
--circuit_breaker_timeout_sec```

PR #1422, opened 2026-08-14
Updating mapping flow to new UUFI python utility

grafnu:


.github/workflows/testing.yml
```diff
@@ -79,7 +79,7 @@ jobs:
     timeout-minutes: 10
     if: ${{ !cancelled() }}
     env:
-      TARGET_PROJECT: ${{ vars.TARGET_PROJECT || '//mqtt/localhost:9001' }}
+      TARGET_PROJECT: //mqtt/localhost:9001
```
khyatimahendru, line 82: Is it intended to run all these tests {Automapping, Sequencer, Blob Updates, etc} against the local setup only?

grafnu:
Yes -- and I realize now we never really talked about this (my bad). Let's
cover it in our next meeting!

butler/src/connection.py
```diff
@@ -0,0 +1,335 @@
...
+import os
+import ssl
+import sys
+import time
+import uuid
+from cryptography.hazmat.primitives import serialization
+import jwt
+import paho.mqtt.client as mqtt
+
+DEFAULT_GBOS_HOSTNAME = "mqtt.bos.goog"
```
noursaidi, line 14: Prefer to not specify mqtt.bos.goog anywhere. Would prefer an "authentication" mechanism option (jwt or client certificiate and/or password) and a hostname

noursaidi (approved):

khyatimahendru (approved):

butler/src/connection.py
grafnu: Right now this is just replicating the `//gbos/` schema that's used by the tools. Not sure what you mean by "not specify mqtt.bos.goog" -- since that's what it's _defined_ to be...?!?  Either, all just tweaky glue layer stuff.

PR #1425, opened 2026-08-22
Migrate database capture code to butler, in python

grafnu:


khyatimahendru (approved):
Looks good to me! I think I'll reuse your test mcp server and session management for workbench too - thank you for adding that!

PR #1426, opened 2026-08-24
Initial GUMMI interface

grafnu:


butler/src/connection.py
```diff
@@ -274,6 +274,23 @@ def get_model_topics(self, device_id):
...
+        if self.provider == "jwt":
+            return [
+                f"/uufi/r/{self.actual_registry}/d/{device_id}/c/propose/{sub_folder}",
+                f"/devices/{device_id}/propose",
+            ]
+        uufi_topic = f"/uufi/r/{self.actual_registry}/d/{device_id}/c/propose/{sub_folder}"
+        if self.prefix:
+            uufi_topic = f"/{self.prefix}{uufi_topic}"
+        return [uufi_topic]
+
```
khyatimahendru, line 293: Do you see proposals from the BAMBI** interface also coming in through uufi?

khyatimahendru (approved):
Looks good to me.

butler/src/connection.py
grafnu: yes, I think it would form a more generic proposal mechanism. UUFI isn't all that different so I'm hoping it doesn't change much. Might need to refactor some of the pipeline but that should be relatively easy. When piper is ready we can figure out how to have that be the review tool, but I wanted something self-contained too.

PR #1427, opened 2026-08-28
Updates to the tmux process control model

grafnu:


bin/udmi
```diff
@@ -0,0 +1,310 @@
...
+            sess_name="$session"
+            [[ "$sess_name" =~ ^udmi_ ]] || sess_name="udmi_${sess_name}"
+            tmux_capture_logs "$sess_name" "" "$lines"
+        fi
+    else
+        for s in udmi_barbican udmi_butler udmi_bridgehead udmi_base udmi_mcp; do
+            if tmux_session_exists "$s"; then
+                tmux_capture_logs "$s" "" "$lines"
+            fi
+        done
```
khyatimahendru: I guess we're missing logs for the namespaced sessions here

khyatimahendru (approved):

bin/udmi
grafnu: fixed!

PR #1429, opened 2026-08-28
registry discovery fixes

jainrocks:


grafnu:
Not a good idea to put "scan all the registries" and self-healing code into the lower levels of the stack.  The index should be maintained properly, used for immediate results, and then anything self-healing should be done out of the kernel/core operations.

PR #1433, opened 2026-09-02
Init Spotter with discovery and pcap capabilities

khyatimahendru:


.github/workflows/testing.yml
```diff
@@ -729,3 +729,27 @@ jobs:
...
+          java-version: '21'
+          cache: 'gradle'
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.13'
+      - name: Base setup
+        run: |
+          sudo bin/setup_base
+          bin/clone_model
```
grafnu: double check the need to explicitly have bin/clone_model here -- check the other tests... I think that recently just got folded into one of the other setup phases... but you also may not use that phase so I don't know. Just check!

edge/spotter/GEMINI.md
```diff
@@ -0,0 +1,184 @@
...
+
+### 1.4 UDMI Infrastructure Orchestration & Connection Resilience
+Spotter connects to local or cloud-hosted UDMI infrastructure as an edge client node.
+- **Local Infrastructure Pairing**: When testing against local services, start the core stack using the canonical orchestrator [bin/udmi](../../bin/udmi):
+  ```bash
+  # Start Barbican and Butler services
+  bin/udmi start sites/udmi_site_model //mqtt/localhost:46432
+
+  # Launch Spotter targeting the local broker
+  ./bin/spotter sites/udmi_site_model //mqtt/localhost:46432 AHU-1
```
grafnu, line 76: nb, as mentioned elsewhere I think this could (also) be part of the tmux barbican session -- but maybe not as this is testing spotter proper.  At least let's make sure we understand the plan for that.

edge/spotter/GEMINI.md
```diff
@@ -0,0 +1,184 @@
...
+
+---
+
+## 2. Test Execution Environments & Conditions
+
+Tests must be executed under specific conditions to guarantee environment hygiene and avoid false positives/negatives:
+
+### 2.1 Network & Port Requirements
+- **BACnet Port 47808 Isolation**: BACnet discovery broadcasts to UDP port `47808`. If running tests locally, no other application (e.g., a bare-metal BACnet scanner or another running container) must be bound to port `47808` on the host.
+- **Custom Docker Bridge Subnet**: Parity tests utilize an isolated docker bridge network named `spotter-parity-net` on subnet `192.168.12.0/24`. The host's gateway is defined as `192.168.12.254`. This subnet must not conflict with any existing network interfaces on the host.
```
grafnu: maybe `parity-spotter-net` just for namespace consistency with the other `parity-` things?  Or maybe `spotter-parity-` everywhere?

edge/spotter/GEMINI.md
```diff
@@ -0,0 +1,184 @@
+# Spotter Testing & Verification Standards
```
grafnu, line 1: spotter is going to be the most well-tested part of the system after this file!

edge/spotter/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+Spotter runs as a single unified process using the UDMI Python Client Library (`clientlib`), managing three core managers under a single device identity:
+
+1. **`LocalnetManager` & Pluggable Family Providers**:
+   - **`BacnetFamilyProvider`**: Performs active BACnet Who-Is / I-Am discovery, object enumeration, and UDP port extraction ([b/549909773](https://b.corp.google.com/issues/549909773)).
+   - **`EtherFamilyProvider`**: Performs Layer-2 Ethernet / ARP and ping discovery.
+   - **`PassiveFamilyProvider`**: Listens for passive broadcast network traffic and extracts discovered device metadata.
+2. **`SpotterDiscoveryManager`**:
+   - Manages scheduled and on-demand discovery sweeps.
+   - Streams live remote packet capture traces (`events/stream`) safely buffered in RAM.
```
grafnu: what do you mean by "safely buffered in RAM"?

schema/events.json
```diff
@@ -14,6 +14,9 @@
     },
     "discovery": {
       "$ref": "file:events_discovery.json#"
+    },
+    "stream": {
```
grafnu: I'm just pondering if this should be `stream`, `streams`, or `streamset`  -- it should definately support more than one stream at a time. At some point (doesn't need to be part of _this_ PR) I think there would also be config/state messages associated with streams e.g. to indicate what has been received already, and at that point it would likely be a set of things like `blobset` or `pointset` which means `streamset` has parity.

schema/events_streams.json
```diff
@@ -0,0 +1,40 @@
...
+    "timestamp": {
+      "description": "RFC 3339 UTC timestamp the stream chunk event was generated",
+      "type": "string",
+      "format": "date-time"
+    },
+    "version": {
+      "description": "Version of the UDMI schema",
+      "type": "string"
+    },
+    "session_id": {
```
grafnu, line 19: With this structure, I think it might be `streams` since this is describing which one of the streams, which is different from pointset where the event itself has a set of points.

grafnu (commented):
Mostly stuff about testing and then, of course, a naming decision!

.github/workflows/testing.yml
khyatimahendru: Thank you, I have removed this step. I still needed to add the steps to setup the ca and reflector keys.

edge/spotter/GEMINI.md
khyatimahendru: I added spotter as an optional service in tmux bridgehead instead (similar to pubber).

edge/spotter/GEMINI.md
khyatimahendru: Done. I have renamed it to parity-spotter-net everywhere.

edge/spotter/README.md
khyatimahendru: I meant it won't be written to disk. It is buffered in the volatile memory for streaming and never written to disk. I have rephrased for clarity.

schema/events_streams.json
khyatimahendru: Done. I've renamed it to `streams`.

schema/events.json
khyatimahendru: Renamed to streams

edge/spotter/README.md
```diff
@@ -0,0 +1,195 @@
...
+   - **Zero-Disk Streaming**: Packets are buffered dynamically in volatile memory (RAM) and sequentially published as reliable base64 chunks (`StreamsEvents`) over the universal streaming MQTT event topic (`events/streams`).
+   - **Zero Secret Distribution**: Leverages the existing mTLS hardware key/certificate connection directly, avoiding external network credentials or outbound HTTP rules at the edge.
+3. **Ad-hoc PCAP Reassembly ([bin/reassemble_pcap](bin/reassemble_pcap))**:
+   Reassembles chunked `events/streams` messages (from JSON, JSONL, or `mosquitto_sub`) back into a valid `.pcap` binary capture file:
+   ```bash
+   # Reassemble stream events file into a pcap file:
+   ./edge/spotter/bin/reassemble_pcap stream_events.json capture.pcap
+
+   # Or stream directly from mosquitto subscriber:
+   mosquitto_sub -h $BROKER -t '/r/+/d/+/events/streams' | ./edge/spotter/bin/reassemble_pcap - live.pcap
```
grafnu, line 90: THis is the part that I think we were sorting out -- i.e. maybe not "events/streams" but something like "streams/XXX" -- just so that the concept of a stream could be easily split out at the message routing layer.  Rather than "subType == events" we would end up with "subType == streams" (or functionally equivalent).  

However, I don't think we should delay _this_ PR until that's sorter, rather I'd say let's keep this PR like this and have it be provisional, and we can adjust in an iteration!  Functionally it shouldn't be much different for the base case, just a moving a few fields around.

grafnu (approved):
I took a detailed look, short of running it all, and it looks good as far as I can tell. The only comment really is that ongoing discussion about the message handling, but I think we should keep it as you have it now and then iterate on that when some of the other bits get put into place (e.g. the reassembly bit).

edge/spotter/README.md
khyatimahendru: I am addressing this as well as other bits we discussed in another PR. I am merging this one.

